PR #2, opened 2018-09-06
Add parameter names to function prototypes

t-sommer:
This PR adds parameter names to the function prototypes in the FMI headers to improve readability and allow IDEs to provide code completion.

Changes:

- add parameter names in fmi2FunctionTypes.h (as given in the specification document)
- add revision note
- replace non-ASCII copyright character w/ "(C)"
- fix indentations to align function parameters

@pmai did some research on this:

> Function Type Declarators with parameter names have been legal since ANSI C89/ISO C90, i.e. since declarators have been well-defined (see section 3.5.4.3 Function declarators, especially also the examples provided, together with sections 3.5.1/3.5.6). So any compilers having problems with this would not even be C88/90-compliant, and should be disregarded for FMI 3.0. If someone wants to use FMI3.0 with such compilers they can easily fix the headers for their compiler-specific needs in any case, without impacting compatibility.
> 
> The benefits in readability and IDE support far outweigh the out of the box support of obscure non-compliant C compilers.

### DIFF
--- a/headers/fmi2FunctionTypes.h
+++ b/headers/fmi2FunctionTypes.h
@@ -7,6 +7,7 @@
    It declares data and function types for FMI 2.0
 
    Revisions:
+   - Sep.  6, 2018: Parameter names added to function prototypes
    - Apr.  9, 2014: all prefixes "fmi" renamed to "fmi2" (decision from April 8)
    - Apr.  3, 2014: Added #include <stddef.h> for size_t definition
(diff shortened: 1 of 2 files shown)


beutlich:
Please wait with new PRs and issues till Trac2GitHub succeeded. We already have a #2 as https://trac.fmi-standard.org/ticket/2.

andreas-junghanns:
Much better -- because consistent with the current standard text -- will need some improvements on the names in another commit.

PR #444, opened 2018-10-20
Fix typo in function name

beutlich:


pmai:
Hmmm, if we are going to take git flow seriously as a branching model, then PRs should never be merged into master directly. They should either end up in develop, or for 2.0.1 fixes they should be merged into a newly created 2.0.x development branch...

There should never be stuff on master that is beyond develop...

Arguably the typo fixes and reformatting on master now should have happened on develop, but that’s neither here nor there.

chrbertsch:
OK, agree that following Gitlfow that was the wrong target to merge to.
Following https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow I suggest to create a "release" branch for 2.0.1 (no feature to be added), and use "develop" for 3.0.
Do you agree?

beutlich:
Well, which git flow should be adopted, is matter of taste and was never discussed nor decided.

pmai:
Actually it was (discussed and decided) IMHO. While I personally think something more lightweight would have been sufficient, and for 2.0.1 pure git flow does not really work that well (the fixes for 2.0.1 will not be mergeable into develop, i.e. we need to manually cherry pick). But the general consensus was to try git flow. And hence we should stick to that at least somewhat IMHO.

Regardless we will have to separate 2.0.1 maintenance from 3.0.0 work since there will be large changes in 3.0.0 and merging between those is not likely to be useful or easy.

Given that the fix already landed in master the easiest approach is likely to branch a new release/2.0.1 branch from that, merge the current develop into that branch and continue 2.0.1 work on that branch (as if everything where maintenance/hotfixes).

beutlich:
Ah sorry, confusing if each MAP adopts a different strategy,

chrbertsch:
I created a release/2.0.1 branch

beutlich:
FMI 2.0.1 and 3.0, right?

PR #446, opened 2018-10-20
Description of gitflow branches

chrbertsch:
Open question: Shall we directly commit to `master` when fixing typos in the conversion from the release PDF version to the Asciidoc version for the standard or use a separate branch for that? 
I propose to directly commit to  master in that exceptional case.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -11,6 +11,17 @@ It's supported by https://fmi-standard.org/tools/[100+] tools and maintained as 
 - `headers` -- C header files
 - `schema` -- the XSD schema for the modelDescription.xml
 
+== Workflow and Branching
+
+We follow the gitflow workflow, see, e.g., https://datasift.github.io/gitflow/IntroducingGitFlow.html
+
+For the current work on the next planned releases of the FMI standard we use
+
+- https://github.com/modelica/fmi-standard[master] only for released versions, i.e., currently for FMI 2.0
+- https://github.com/modelica/fmi-standard/tree/release/2.0.1[release/2.0.1] for bugfixes, maintenance of the FMI 2.0 standard (no new features)
+- https://github.com/modelica/fmi-standard/tree/develop[develop] for the development of the FMI 3.0 standard, used for merging features
+- `feature/XXX` for the development of a new feature (when starting, branch for develop)
+
 == Building the specification document
 
 The FMI specification is written in http://asciidoc.org/[AsciiDoc], a plain-text mark-up format that can be converted to HTML, DocBook, PDF and other formats. To build the documentation locally


t-sommer:
As I understand GitFlow the maintenance of old releases happens on [support branches](https://gitversion.readthedocs.io/en/latest/git-branching-strategies/gitflow-examples/#support-branches). IMHO it's also a good practice to prefix versions in tags and branches with "v" e.g. `release/v2.0.1`. Since 2.0.1 will likely never be merged back into master it should go into `support/v2.0.1`.

t-sommer:
> Shall we directly commit to master when fixing typos [...]?

No. I think we agreed that any contributions to the respective branches should be made via pull requests (this is not clearly stated in the docs). If it's only a typo we don't have to discuss it but at least it gets reviewed by another person.

pmai:
Right, though they are not mentioned in most GitFlow docs, support branches are what make sense here. Though I guess the branch should be named support/2.0.x, since in theory multiple 2.0.x releases could be tagged from that. (While I previously also used various v2.x or version-2-x etc. conventions, I think the rule nowadays is to just use support/2.0.x since for those branches it is always clear that a version number is meant; see the examples you linked to).

t-sommer:
I edited my previous comment as you suggested: `support/v2.0.x`

t-sommer:
If we only have development, feature (FCP) and support branches we could use the following schema (I guess this is what you had in mind @pmai). Minor changes (typos etc.) could be handled via PRs.

```
master
  |
  +--->+ branch "v2.0.x"
  |    |
  |    + tag "v2.0.1"
  |    |
  |    + tag "v2.0.2"
  |
  |
  +--->+ branch "v3.x"
  |    |
  |    + tag "v3.0"
  |    |
  |    +--->+ branch "v3.0.x"
  |    |    |
  |    |    + tag "v3.0.1"
  |    |    |
  |    |    + tag "v3.0.1"
  |    | 
  |    + tag "v3.1"
  |    |
  .    .
  .    .
  .    .
```

Every vertical line (`|`) could be a merged FCP branch or PR:

```
master
  |
  +<- merge PR "fix-typo-in-fmi-spec"
  |
  +--->+ branch "fcp06-arrays"
  |    |
  +<---+ merge "fcp06-arrays"
  |      
```

t-sommer:
As for the "v" prefix: [Semver use it for their tags](https://github.com/semver/semver/tags). I would also suggest to rebase feature branches onto the latest master branch before merging them to make it easier for people to review changes in the standard.

t-sommer:
I've put together [a proposal](https://github.com/t-sommer/fmi-standard/blob/branching-schema/CONTRIBUTING.adoc#branching-and-versioning) for a branching and tagging schema that only uses feature and maintenance branches.

chrbertsch:
Just to be sure, you propose _not_ to follow Gitflow any more, right?

t-sommer:
Yes, this would be a custom schema. I've considered it again and I think it suits our purpose better.

pmai:
FWIW on the v Prefix semver seems to have reversed course: While semver 1.0.0 had the v prefix in its spec (and hence also uses it in their repo) that language was explicitly removed from semver 2.0.0...

pmai:
On the proposal, I think this looks quite good, I‘d just urge to use prefixes like feature/ for feature branches, and maintenance/ or releng/ etc. for the maintenance branches since this really helps with command-line autocomplete and keeps everything tidier.

We’ll probably have more opportunity to discuss tomorrow... ;)

t-sommer:
I've updated [the proposal](https://github.com/t-sommer/fmi-standard/blob/branching-schema/CONTRIBUTING.adoc#branching-and-versioning) based on your suggestions and added some examples of what it would look like.

t-sommer:
This has been clarified in d43c130bf340b8157ccfacac4af868a607da3ac6

PR #448, opened 2018-10-23
Remove www from domain

beutlich:


t-sommer:
This is fixed with 1a32ee389d13d14b84447f1c2074cf9c91f37b13. Thanks!

PR #465, opened 2018-10-29
Add parameter names to function definitions in fmi3FunctionTypes.h

t-sommer:


andreas-junghanns:
This is already much better -- because consistent with the current standard text -- but we will have to fix a number of minor issues to make the names more consistent.

PR #467, opened 2018-10-29
Regularize state name and function name syntax

pmai:
States are now referenced in bold face everywhere, and function names
are referenced without '(..)' added to some of the reference.


andreas-junghanns:
As if we tried to produce the most inconsistent standardization document possible - thank you for getting yet more kinks out!

Issue #426, opened 2018-10-17
fmi2CancelStep clarification


modelica-trac-importer:
**Reported by masoud on 4 Dec 2017 12:54 UTC**
When a fmi2DoStep() command is cancelled using fmi2CanCelStep(), according to the statechart diagram in FMI2.0/CoSimulation standard in page 103, and also  the table  in page 105, the state of FMU can be changed using fmi2SetFMUstate(). But in the description of the fmi2CancelStep() (page 101) the function fmi2SetFMUstate() has not been mentionned as a legal function to be called in the "step Cancelled" state to change the state. This has caused confusion.

----

Migrated-From: https://trac.fmi-standard.org/ticket/426

KarlWernersson:
@pmai Did you fix this for 2.0.1?
When i investigate the text has been updated to
Afterwards only calls to `fmi2Reset`, `fmi2FreeInstance`, or `fmi2SetFMUstate` are valid to exit the *step Canceled* state.

From
Afterwards it is only allowed to call fmi2Reset or
fmi2FreeInstance.

Or was this changed in the conversion from word to adoc?

KarlWernersson:
Fixed by pmai 29 oct 2018 16.19

PR #468, opened 2018-10-29
Fix #461 clarify valid calls after canceled step

### LINKED ISSUE #461: fmi3CancelStep clarification
The clarification according to #426 (see PR #460 ) should be ported to 3.0 as well.

pmai:
Port clarification from 2.0.1 (see #460 and #426).


andreas-junghanns:
I agree with the changes.

PR #469, opened 2018-10-29
Include text from headers in source listings

t-sommer:


pmai:
Looks good on first look, however there seems some jumbled comments in the fmi3TypesPlatform.h and I'd switch back to C90-style comments for the headers, since nothing in there needs C99 for now (might change if we switch to use C99 types in the future, but I'd do this changeover later, if needed).

Will commit fixes in the next couple of minutes...

andreas-junghanns:
end tag is missing for // tag::VariableTypes[]

andreas-junghanns:
// tag::Component[]
copy-paste errors - used also where they should be renamed

andreas-junghanns:
The fmi3Binary argument in fmi3SetBinaryTYPE still does not have a name. Did we not commit that fix?

andreas-junghanns:
fmi3CompletedIntegratorStepTYPE argument list needs a line break

pmai:
I'm in the process of fixing those, commit landing any minute now...

andreas-junghanns:
just making sure you know those I also found...

pmai:
This should hopefully address all missing instances, and switch the headers to back to C90 compatibility for now.

andreas-junghanns:
No, I noted all the odd things I saw.

PR #478, opened 2018-11-03
Fix broken link to "Lee E.A., Zheng H. (2007)"

t-sommer:


beutlich:
FMI 2.0.1 and 3.0, right?

Issue #435, opened 2018-10-17
Remove variable arguments from C-API


modelica-trac-importer:
**Reported by tsommer on 6 Jun 2018 08:09 UTC**
Currently it is not possible to call or implement functions with a variable number of arguments in languages
and libraries that bind to functions in shared libraries dynamically like [Java Native Access](https://github.com/java-native-access/jna),
[Python ctypes](https://docs.python.org/3/library/ctypes.html) and [.NET P/Invoke](https://docs.microsoft.com/en-us/dotnet/standard/native-interop).

In order to allow these platforms to fully support the FMI API, I suggest to remove all variable arguments lists from the C-API.

Currently variable arguments are only used in one callback function (fmi2FunctionTypes.h, line 124):

```
typedef void (*fmi2CallbackLogger)(fmi2ComponentEnvironment, fmi2String, fmi2Status, fmi2String, fmi2String, ...);
```

The same result can be achieved by formatting the message string using [sprintf()](http://www.cplusplus.com/reference/cstdio/sprintf/) in the FMU and passing it to the callback as a single argument.

----

Migrated-From: https://trac.fmi-standard.org/ticket/435

modelica-trac-importer:
**Comment by andreas&#x2e;junghanns on 6 Jun 2018 09:37 UTC**
This request is asking to make the FMU code more difficult to write (e.g. memory handling) that now to make the importers life easier (no need to handle variable number of arguments).

Who should we make life easier: Exporters or importers?

modelica-trac-importer:
**Comment by tsommer on 6 Jun 2018 13:17 UTC**
This is not about making life easier for the (many) importers to make it harder for the (fewer) exporters but to allow (some) importers to use arguments that are currently "unreachable".

modelica-trac-importer:
**Comment by Andreas Nicolai on 6 Jun 2018 16:45 UTC**
Replying to [comment:1 Andreas Junghanns (QTronic)]:
> This request is asking to make the FMU code more difficult to write (e.g. memory handling) that now to make the importers life easier (no need to handle variable number of arguments).
> 
> Who should we make life easier: Exporters or importers?

I can only imagine a scenario for the exporting FMU where some of the
strings to be passed back are fixed character constants and one wants to
avoid to allocate a chunk of memory in the exporting FMU to fit the
entire concatenated string. This could only pose a problem for very
limited hardware (i.e. embedded systems). But surely, one would not
expect very verbose (prosa form) error messages from such a system,
[...]

PR #480, opened 2018-11-05
Remove variadic arguments from logger callback (#435)

t-sommer:
As decided in the FMI Steering Committee Meeting on Nov 6, 2018


beutlich:
FMI 3.0 only, right?

PR #482, opened 2018-11-06
Add new numeric variable types to C API

t-sommer:
- add new fmi3[U]Int{8,16,32,64} and fmi3Float variable types
- rename fmi3Boolean to fmi3Bool
- replace fmi3Integer with fmi3Int32
- replace fmi3Real with fmi3Double
- rename getters, setter and status functions accordingly
- remove "Real" from fmi3SetRealInputDerivatives and fmi3GetRealOutputDerivatives

### DIFF
--- a/headers/fmi3FunctionTypes.h
+++ b/headers/fmi3FunctionTypes.h
@@ -126,12 +126,12 @@ typedef struct {
 
 /* tag::EventInfo[] */
 typedef struct {
-    fmi3Boolean newDiscreteStatesNeeded;
-    fmi3Boolean terminateSimulation;
-    fmi3Boolean nominalsOfContinuousStatesChanged;
-    fmi3Boolean valuesOfContinuousStatesChanged;
-    fmi3Boolean nextEventTimeDefined;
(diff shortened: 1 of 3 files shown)


t-sommer:
I've added [the reasons](https://github.com/modelica/fmi-design/blob/master/FMI_ChangeProposals/FCP_017_Numeric_Variable_Types/FCP_Numeric_Variable_Types.adoc#naming) for the renaming in the FCP. We can discuss this.

pmai:
Yes, but the rationale does not fit for fmi3Boolean: We are not switching to bool/_Bool, in fact nothing changes for fmi3Boolean at all (and in fact switching to bool/_Bool would be a bad idea, since different compilers do not necessarily agree on the size of _Bool (1 byte vs. 32/64bit word) making it a bad choice for any kind of ABI use). Hence the renaming is just causing uneccessary code breakage and would even be misleading.

And in fact fmi3Boolean is not a numeric type at all (it's just discrete, but there is no implied order, etc.), so I would not touch this at all.

t-sommer:
OK, I will update the FCP and make a new PR.

PR #485, opened 2018-11-06
Initial support for numeric types in schemas

pmai:
This change does not change the enumeration type to map to any of the
new types, but leaves it still implicitly mapped to fmi3Int32 (the successor
of the original fmi3Integer type).  Making the enumeration type map to
any of the new integer types would be useful addition, but would make
code quite a bit more complex, and also leave the schema definitions not
as expressive, since the valid ranges for min/max, etc. will depend on
an attribute.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,14 @@ env:
 
 before_install:
   - docker pull asciidoctor/docker-asciidoctor
+  - sudo apt-get install -y python-lxml
   - gem install html-proofer
 
 script:
   # compile the headers
   - gcc headers/fmi3Functions.h
+  # parse the schema
+  - python -c "from lxml import etree; etree.XMLSchema(file='schema/fmi3ModelDescription.xsd')"
   # generate the HTML documentation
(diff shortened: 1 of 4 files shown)


t-sommer:
Have you noticed that I changed `fmi3Float32` to `fmi3Float` and `fmi3Float64` to `fmi3Double`?
(See also section "Naming" in the FCP)

Also the word "Real" still appears multiple times in the XSDs.

beutlich:
Should it be called fmi3Single then?

pmai:
Yeah, sorry for the missed Float32/64 renaming, updated. The name Real still occurs (as does the name Integer) for stuff that is common across the new real types.

beutlich:
Just make sure it aligns with DCP. @martinkrammer

martinkrammer:
In DCP we call those Float32 and Float64.

t-sommer:
@martinkrammer, is the current version of the DCP spec available?

martinkrammer:
> @martinkrammer, is the current version of the DCP spec available?
Release Candidate 2 is available via ITEA3 website: https://itea3.org/project/acosar.html
Release Candidate 3 is almost ready and will be provided to MA members for review within the next few days.

t-sommer:
From the data types point of view the current proposal seems to match the DCP types 1-to-1 except for fmi3Boolean. I don't see a problem if we call things differently.

PR #489, opened 2018-11-08
fix #411 Expanded text on mandatory api functions

### LINKED ISSUE #411: Are all of the API functions mandatory?
**Reported by aviel on 15 Feb 2017 08:27 UTC**
The model description for Model Exchange and Co-simulation defines three flags _canGetAndSetFMUstate_, _canSerializeFMUstate_ and _providesDirectionalDerivative_. According to sections 2.1.8 (p. 25) and 2.1.9 (p. 26), it is stated that the functions:
* `fmi2GetFMUstate`
* `fmi2SetFMUstate`
* `fmi2FreeFMUstate`
* `fmi2SerializedFMUstateSize`
* `fmi2SerializeFMUstate`
* `fmi2DeSerializeFMUstate`
* `fmi2GetDirectionalDerivative`
are only supported by the FMU, if the (above) optional capability flags (...) are explicitly set to _true_.
Does it mean that these functions may not or shall not exist in the binary part of the FMU if these flags are set to _false_?
Are these functions mandatory, and always present in the FMU? Indeed, if the flags are set to _false_, these functions could return _fmi2Error_ if incorrectly called.
It is not clear in the specification, and at this time some FMU from the Cross Check Repository provide some or all of the above functions when the flags are false, whereas some other FMU do not define the functions.

----

Migrated-From: https://trac.fmi-standard.org/ticket/411

KarlWernersson:
Please check that i have done this correctly

fix #411

### DIFF
--- a/docs/2_3_fmu_distribution.adoc
+++ b/docs/2_3_fmu_distribution.adoc
@@ -1,10 +1,14 @@
 === FMU Distribution
 
-An FMU description consists of several files. 
-n FMU implementation may be distributed in source code and/or in binary format.
+An FMU description consists of several files.
+An FMU implementation may be distributed in source code and/or in binary format.
 All relevant files are stored in a ZIP file with a pre-defined structure.
 The implementation must either implement all the functions of FMI for Model Exchange
 or all the functions of FMI for Co-Simulation or both.
+Specifically it is required that all functions specified for Model Exchange and/or Co-Simulation are present,
+even if they are only needed for capabilities that the FMU does not support.
+The behavior of those functions are unspecified, so while calling environments can rely on the functions being present,
+they cannot rely on any particular behaviour for functions only needed for capabilities the FMU does not support.
 The extension of the ZIP file must be "**.fmu**",
 for example, "HybridVehicle.fmu".
 The compression method used for the ZIP file must be "deflate" _[(most free tools,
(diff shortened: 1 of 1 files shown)


docs/2_3_fmu_distribution.adoc
```diff
@@ -1,10 +1,14 @@
...
-An FMU description consists of several files. 
-n FMU implementation may be distributed in source code and/or in binary format.
+An FMU description consists of several files.
+An FMU implementation may be distributed in source code and/or in binary format.
 All relevant files are stored in a ZIP file with a pre-defined structure.
 The implementation must either implement all the functions of FMI for Model Exchange
 or all the functions of FMI for Co-Simulation or both.
+Specifically it is required that all functions specified for Model Exchange and/or Co-Simulation are present,
+even if they are only needed for capabilities that the FMU does not support.
+The behavior of those functions are unspecified, so while calling environments can rely on the functions being present,
```
beutlich: The behavior of those functions **is** ...

docs/2_3_fmu_distribution.adoc
```diff
@@ -1,10 +1,14 @@
...
+they cannot rely on any particular behaviour for functions only needed for capabilities the FMU does not support.
```
beutlich: Spelling of behaviour vs. behavior in the same sentence.

KarlWernersson:
wops, i maybe screwed something up?

Issue #423, opened 2018-10-17
Strange license extension in FMI ME 1.0.1


modelica-trac-importer:
**Reported by beutlich on 14 Sep 2017 09:32 UTC**
The header files of [FMI ME 1.0.1](https://svn.modelica.org/fmi/branches/public/specifications/v1.0/FMI_for_ModelExchange_v1.0.1.zip) contain a new (w.r.t. FMI ME 1.0) license extension:

```
   with the extension:

   You may distribute or publicly perform any modification only under the
   terms of this license.
```

To me it seems the intention was to forbid sublicensing, however

* This extension now changes the license from a standard BSD license to a custom BSD-style license.
* BSD (all three flavours) already forbid sublicensing (which is the main difference to MIT license actually).
* This extension was only added to FMI ME, but not FMI CE, thus it rather seems an incomplete extension.

Proposal: Remove that extension again and get the license correctly sorted out to use the 3-Clause BSD license (probably). See also #347.

----

Migrated-From: https://trac.fmi-standard.org/ticket/423

modelica-trac-importer:
**Comment by beutlich on 14 Sep 2017 09:39 UTC**
Just have seen, that this license extension is already there in FMI ME 1.0, thus it is not a newly introduced issue.

modelica-trac-importer:
**Comment by andreas&#46;junghanns on 14 Sep 2017 11:09 UTC**
CAn we close this ticket?

modelica-trac-importer:
**Comment by beutlich on 14 Sep 2017 11:18 UTC**
You mean closing in favor of #347? Probably.

modelica-trac-importer:
**Comment by otter on 14 Sep 2017 12:06 UTC**
The license for FMI 1.0 was defined in a contract between the MODELISAR partners (Project Cooperation Agreement):

 Article 7.1:
 a) During the term of this agreement, and except if the Parties agree otherwise during a PCC, such license shall be granted by Parties as follows: For the FMI specification document: CC-BY-SA (http://creativecommons.org/licenses/by-sa/3.0/legalcode, http://creativecommons.org/about/licenses) For the FMI source code files: BSD license (http://www.opensource.org/licenses/bsd-license.html) incorporating the following paragraph: “You may distribute or publicly perform any modification only under the terms of this license”.

Therefore, it is not possible to just change the license without a new legal agreement.

[...]

Issue #347, opened 2018-10-16
Clarify BSD version in standard documents


modelica-trac-importer:
**Reported by andreas&#46;junghanns on 9 Sep 2015 19:43 UTC**
From veronica.arama@&hellip;:

>  I downloaded FMI 2.0 and I have an issue related to the license of some files. You mentioned in the provided pdf that the license is BSD 2 clause but still in the files itself you’ve listed BSD 3- clause. Which is the correct license BSd 2 clause or BSD 3 clause?
>
> Can you please clarify the issue?

----

Migrated-From: https://trac.fmi-standard.org/ticket/347

modelica-trac-importer:
**Comment by cbertsch on 24 Nov 2016 07:28 UTC**
Duplicate of #351.
Clarification of BSD3 usage shall be part of FMI 2.0.1. 

BSD3 usage shall also be clarified by putting the correct link on https://www.fmi-standard.org/development (instead of link to BSD2)

modelica-trac-importer:
**Modified by andreas&#46;junghanns on 6 Feb 2017 10:41 UTC**

modelica-trac-importer:
**Modified by beutlich on 6 Sep 2017 06:59 UTC**

beutlich:
Seems 44e7a9fbd2cab711e39051b73c0186da91ceff6c addresses this issue, but I could not figure out why there was not a decent pull request by @t-sommer.

beutlich:
Note that README.adoc links to the [2-Clause BSD License](https://opensource.org/licenses/bsd-license.html).

t-sommer:
This is already addressed in #462

KarlWernersson:
@andreas-junghanns Can this one be closed?

beutlich:
Just to summarize: In master branch, it is the 3-Clause BSD License in the README.adoc and the 3-Clause BSD License with the license extension (as discussed in #423) in the 3 header files. From my view, this clarifies the previous mix of 2-Clause BSD with 3 clauses. We still have #423 to clarify the unusual license extension.

PR #492, opened 2018-11-14
Change license headers to “2-Clause BSD License”

t-sommer:
to comply with FMI Project Rules v2.0 (2016-06-21)

- change license headers of *.h and *.xsd to "2-Clause BSD License"
- remove license extension
- add license LICENSE.txt

See also: #347, #423

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,473 @@
+Copyright (C) 2008-2011 MODELISAR consortium,
+              2012-2018 Modelica Association Project "FMI"
+              All rights reserved.
+
+Source code or other data, such as C-header and XML-schema files, that
+accompany the specification documents are released under the BSD 2-Clause
+license:
+
+----------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


chrbertsch:
https://fmi-standard.org/assets/FMI_ProjectRules_v2.0_2016_06_21.pdf mentions BSD-2 instead of BSD-3. But that was before MA decided to use BSD-3 for the MSL.  For consistency within the MA we could also consider switching to BSD-3 for the FMI headers. 
But this is not very important; much more relevant is the removal of the extension sentence.

beutlich:
> But that was before MA decided to use BSD-3 for the MSL. For consistency within the MA we could also consider switching to BSD-3 for the FMI headers.

Confirmed. It was only decided to go for the 3-Clause BSD License within MAP-LIB. MAP-FMI is still free in their choice.

> But this is not very important; much more relevant is the removal of the extension sentence.

True and true.

Issue #442, opened 2018-10-17
Remove #type in loggerCallback


modelica-trac-importer:
**Reported by karl&#46;wernersson on 26 Jun 2018 14:22 UTC**
Remove #type in loggerCallback since valueReference will be unique over types

----

Migrated-From: https://trac.fmi-standard.org/ticket/442

pmai:
See issue #435 for whether to completely remove #-style references in loggerCallback.

PR #497, opened 2018-11-26
Remove "#type" style variable references in log messages (#442)

t-sommer:
See #442.


pmai:
I think the task part/issue only deals with removing the type prefix letter (i.e. i, r, etc.), not the whole facility. I think that removing the whole facility is a good idea, but I don't think there has been any decision on that part.

APillekeit:
Removing the type prefix letter is probably a good idea since it does not provide any additional information. We nevertheless think that tools could exploit a #type style log message with the valueReference information (e.g. #<valueReference># ...) for showing more meaningful log messages (even in the case of aliases, which may lead to a list of variable names).

TorstenBlochwitz:
At the design meeting on February 13th in Paderborn we polled about this with the result to remove the r, i, s character but to keep the #<ValueReference># functionality (8:3 (abstain: 3)). We recommend to the Steering Committee to follow this poll.

Issue #388, opened 2018-10-17
Clarifications regarding FMI version string: forward/backwards compatibility; version string.


modelica-trac-importer:
**Reported by cbertsch on 15 Oct 2016 19:25 UTC**
The meaning of the digits of FMI versinos (e.g. 2.0, 2.0.1) has been clarified at the FMI Meeting in Oberpfaffenhofen 9.2.2015 ( https://svn.fmi-standard.org/fmi/trunk/Meetings/2015-02-09-FMI-Oberpfaffenhofen/ ), i.e. after the release of the FMI 2.0 standard.

Thus some clarifications are needed for the FMI 2.0.1 standard:
* The meaning of major, minor and bugfix releases shall be explained in the 2.0.1 standard and should be used consistently (e.g. on page 31 of the 2.0 standard, 2.0.1 is called a "minor" release, but should be a bugfix releases.
* What shall an FMU exported to a FMI 2.0.1 have as version string in the xml? According to the 2.0 standard, I would have expected "2.0.1". However, in the course of creating first reference FMUs, we created an FMU with version string "2.0.1" in the modeldescription.xml, but none out of five tested importing tools (FMI 2.0 supporting, released in 2016 after the clarification of the version digit meaning)  accepted this FMU. So de-facto, such an FMU will be incompatible with FMI 2.0 importing tools. We should consider leaving the version string in FMI 2.0.1 to "2.0" and giving the "2.0.1" information only as an annotation.
* What shall be the return value of fmi2GetVersion(void) for a 2.0.1 FMU be? "2.0" or "2.0.1"? This could both happen for a "2.0.1" source code FMU beeing compiled with  the fmi2Functions.h Header file from either an FMI 2.0  or 2.0.1 supporting target. Must the return value of fmi2GetVersion be exactly the same as the version string of the "modeldescription.xml"? Then we should leaver both to "2.0".

----

Migrated-From: https://trac.fmi-standard.org/ticket/388

modelica-trac-importer:
**Comment by rerbacher on 10 Nov 2016 09:37 UTC**
[...]

PR #498, opened 2018-11-26
Remove future minor revisions from XML specification (#388)

t-sommer:


chrbertsch:
FMI 2.0.1 is still used in picture https://github.com/modelica/fmi-standard/blob/support/v2.0.x/docs/images/fmiModelDescription_schema_2.png

PR #511, opened 2018-12-16
Remove the GUID

t-sommer:


chrbertsch:
In non-normative text or in an implementers guide we could point to https://github.com/viproma/fmu-uuid for the generation of GUIDs.

pmai:
Is there any benefit gained from restricting this to a GUID/UUID? I think the field is actually a misnomer, since a hash of defining data is really what is useful here, which is exactly not what a UUID provides...

And the importer does not need to know the format, nor should it care, but just treat it as an opaque chunk of data to provide to the FMU.

If it is specified to be a UUID, importers will want to use this to check for interface compatibility or other stuff, and that’s just wrong.

t-sommer:
Allowing arbitrary characters can cause problems when the GUID is used in automated workflows where strings need to be mangled. I'm open for a different definition but special characters should not be allowed (e.g. `:`).

xogeny:
I think there could be value to defining the syntax of the field (*i.e.,* what characters are allowed) but I would be a lot more cautious about specifying the semantics.  For example, @pmai's point about whether this is a hash or a uuid (very different things) is valid.  We should be very careful in assuming one vs .the others.

As to the general point that @t-sommer is making about "automated workflows", I would expect that many automated workflows would provide a means of escaping any arbitrary string if there are restrictions.  For example, if you want to include the GUID in a URL, there are well defined methods for escaping the string which completely avoids the issue of having to place unnecessary upstream restrictions on the contents of the string.  In a sense, my concern is that what @t-sommer is proposing is what we would call (in English), "the tail wagging the dog".  In other words, those "automatic workflows" should be robust enough to handle arbitrary strings or I would argue the issue is with those workflows.

t-sommer:
@pmai, can we resolve this here or should we move this to an issue and discuss it (or close it altogether)?

beutlich:
> should we move this to an issue

Every pull request also is an issue. No need to duplicate.

pmai:
As discussed, the GUID attribute really does not currently make sense as a GUID/UUID, and the rationale for having the GUID/UUID value as an opaque checking cookie blindly passed by the importer to the FMU for checking of XML vs. Code congruence also makes little sense, as modifications to the XML will not invalidate that cookie (i.e. the FMU would be better of having access to the XML and checking hashes directly).

So rather than trying to fix this whole mess, just deleting the GUID attribute from the spec seems like a more promising route. To be discussed and decided during next technical meeting / SC meeting.

APillekeit:
I agree that the purpose of checking the XML against the Source Code is not well addressed by the GUID. Nevertheless, it might be that removing the GUID completely will hurt tools that may have used this information for different purposes – not documented in the standard.

My suggestion would be to 
-	Keep the GUID attribute in the XML schema
-	IF this is seen as needed: Remove the text in the standard documentation related to checking the XML against the Source Code. In that case also the fmi3Instantiate() function would have to be adapted accordingly.
-	Provide a link to a description of a valid GUID generation. E.g. https://www.itu.int/en/ITU-T/asn1/Pages/UUID/uuids.aspx
-	Use the GUID attribute as a descriptive information for that model container in the same way as “author” or “generationTool”

andreas-junghanns:
Summary of discussion:
 - what we have in 1.0 and 2.0 is severely lacking definition of what we expect the UUID to encode and when it changes: we could define this to make UUIDs in XMLs reliably useful
 - since it is broken already, there is no big penalty to leave it as messy as it is, we also have other places that are underspecified, e.g. modelIdentifier
 - we could only remove it from the C-API only, leave in XML as optional
 - we could use MD5 of the XML as hash to send to the DLL at instantiation to allow the FMU to refuse to run

The DCP standard has a slightly different approach to separately store the UUID (to check when communicating).

If we can define good use cases (see AndreasP's comment), then we need to fix the definition in the standard.

PR #532, opened 2019-02-12
Feature/instantiation token

KarlWernersson:


chrbertsch:
In the C-API: we replace "fmuGUID" by "instantiationToken".  Shouldn't it be "fmuInstantionToken" or what is the logic when to include "fmu" in the name?

PR #536, opened 2019-02-18
Re-name instantiationToken back to GUID

t-sommer:
This PR partially reverts the renaming of the attribute "guid" to "instantiationToken" that has been merged at the FMI Design Meeting in a cloak-and-dagger operation w/o review and against the intention of #511.

Renaming the attribute would break a lot of existing implementations that support different versions of FMI w/o providing any obvious benefit.


chrbertsch:
(the discussion at the design meeting is documented in the meeting minutes, see here: https://github.com/modelica/fmi-design/blob/master/Meetings/2019-02-12_FMI-Design-Meeting/minutes.adoc)

Could you please explain, why this change "would break a lot of existing implementations that support different versions of FMI" more than other changes in the xml or the C-API?

t-sommer:
Take for example the `config.h` file for the Test FMUs (you could find lots of similar examples):

```
#ifndef config_h
#define config_h

// define class name and unique id
#define MODEL_IDENTIFIER Dahlquist
#define MODEL_GUID "{8c4e810f-3df3-4a00-8276-176fa3c9f000}"

// more definitions...

#endif /* config_h */
```

This file is the same for all FMI versions (1.0, 2.0, 3.0) / kinds (ME, CS). Changing the name makes the naming inconsistent for FMI 3.0. However the functionality (that nobody has complained about in 10 years) is still exactly the same. The same is true for the parsers.

My proposal was to either improve the documentation or to remove the attribute altogether but _never_ to rename it.

chrbertsch:
This topic was discussed in detail at the last design meeting with a lot of "reviewers" sitting in the room.

I am not the right person to decide on the effort regarding this change for tool vendors, but many representatives were in favour of this decision.  As "instantiationToken" is more descriptive than GUID I would stay with this.

(In your C-Code example, couldn't you use the DEFINE for the MODEL_GUID also for an "instantionToken"?)

chrbertsch:
Design meeting at Regensburg: 
Poll: 
Stay with instanciationToken: 7
Change back: 1 
abstain: 6

(In the past errors were made with GUID)

Issue #528, opened 2019-01-29
array changes are missing in the xsd schema section of the specification


klausschuch:
The xsd schema section of the specification (current master branch) does not contain the changes proposed by the array working group in the word file (“fmi-design\FMI_3.0\FunctionalMockupInterface_3.0.docx”).

PR #543, opened 2019-03-07
Feature/array description (#528)

klausschuch:
see #528 
Converted schema section 2.2 of the array FCP doc file to adoc.

The changes are based on the last FMI3.0 word file (https://github.com/modelica/fmi-design/blob/master/FMI_3.0/FunctionalMockupInterface_3.0.docx)

some open problems:
- xml examples are missing
- ModelStructure/NumberOfEventIndicators is missing (only mentioned in api section 2.1)
...

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1174,7 +1174,7 @@ Variables with `causality = "parameter"` or `"input"`, as well as variables with
 - If `causality = "output"` or `"local"`, then the start value is either an `initial` or a `guess` value, depending on the setting of attribute `initial`.
 
 |`derivative`
-|If present, this variable is the derivative of variable with `Variable` index "derivative".
+|If present, this variable is the derivative of variable with `Variable` value reference  "derivative".
 _[For example,
 if there are 10 `Variable`pass:[s] and `derivative = 3` for `Variable` 8, then `Variable` 8 is the derivative of `Variable` 3 with respect to the independent variable (usually time).
 This information might be especially used if an input or an output is the derivative of another input or output, or to define the states.]_
@@ -1285,8 +1285,7 @@ This list consists of all variables with
 
 3. all continuous-time states and all state derivatives (defined with element `<Derivatives>` from `<ModelStructure>`) with `initial = "approx"` or `"calculated"` _[if a Co-Simulation FMU does not define the <Derivatives> element, (3) cannot be present]_.
 
-The resulting list is not allowed to have duplicates (for example, if a state is also an output, it is included only once in the list).
-The `Unknowns` in this list must be ordered according to their `Variable` index (for example, if for two variables A and B the `Variable` index of A is less than the index of B, then A must appear before B in `InitialUnknowns`). +
+The resulting list is not allowed to have duplicates (for example, if a state is also an output, it is included only once in the list). +
 Attribute `dependencies` defines the dependencies of the `Unknowns` from the `Knowns` in _Initialization Mode_ at the initial time.
 The functional dependency is defined as:
 
@@ -1337,8 +1336,7 @@ Element `Unknown` in `Outputs`, `Derivatives` and `InitialUnknowns` has the foll
 |Optional attribute defining the dependencies of the unknown latexmath:[\color{blue}{v_{\text{unknown}}}] (directly or indirectly via auxiliary variables) with respect to latexmath:[\color{blue}{\mathbf{v}_{\text{known}}}].
 If not present, it must be assumed that the `Unknown` depends on all `Knowns`.
 If present as empty list, the `Unknown` depends on none of the `Knowns`.
-Otherwise the `Unknown` depends on the `Knowns` defined by the given `Variable` value references.
-The value references are ordered according to magnitude, starting with the smallest index. +
+Otherwise the `Unknown` depends on the `Knowns` defined by the given `Variable` value references. +
 `Knowns` latexmath:[\color{blue}{\mathbf{v}_{\text{known}}}] in _Event_ and _Continuous-Time Mode_ (ModelExchange) and at _Communication Points_ (CoSimulation) for elements `Outputs`, `Derivatives`:
 
 - inputs (variables with `causality = "input"`)
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -635,11 +635,11 @@ The `variability` of the dimension size is in this case the variability of the r
 These two options are mutually exclusive, i.e. for each `Dimension` element either a `start` attribute or an `valueReference` attribute can be supplied, but not both.
 However different dimension sizes can be specified using different mechanisms and can have differing `variability` attributes.
 
-All initial dimension sizes (i.e. prior to any (re)configuration) must be positive integers (i.e. not zero), so that no dimension is initially vanished.
+All initial dimension sizes (i.e. prior to any configuration or reconfiguration) must be positive integers (i.e. not zero), so that no dimension is initially vanished.
 Changes to dimension sizes are constrained by the `min`/`max` attributes of the referenced structural parameters, which can be any non-negative integer, including zero.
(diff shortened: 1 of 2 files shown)


docs/2_2_common_schema.adoc
```diff
@@ -619,15 +619,31 @@ It provides the static information of all exposed variables and is defined as:
...
-The `ModelVariables` element consists of an ordered set of `ScalarVariable` elements (see figure above).
-The first element has `index = 1`,
-the second `index=2`, etc.
-This `ScalarVariable` `index` is used in element `ModelStructure` to uniquely and efficiently refer to `ScalarVariable` definitions.
-A `ScalarVariable` represents a variable of primitive type, like a real or integer variable.
-For simplicity,
-only scalar variables are supported in the schema file in this version and structured entities (like arrays or records) have to be mapped to scalars.
-The schema definition is present in a separate file `fmi3ScalarVariable.xsd`.
-The attributes of `ScalarVariable` are:
+The `ModelVariables` element consists of an ordered set of `Variable` elements (see figure above).
```
t-sommer: When citing form the XML, shouldn't we keep the angle brackets (e.g. `<Variable>`) and use the plain word "variable" when we refer to the abstract idea of a variable?

docs/2_2_common_schema.adoc
```diff
@@ -672,16 +688,29 @@ The algebraic relationship to the inputs is defined via the `dependencies` attri
 - `"local"`: Local variable that is calculated from other variables or is a continuous-time state (see <<ModelStructure>>).
 It is not allowed to use the variable value in another model or slave.
 
-- `"independent""`: The independent variable (usually `time`).
+- `"independent"`: The independent variable (usually `time`).
```
t-sommer: I suggest to only use `independent` (w/o double quotes) except you cite a complete XML tag.

docs/2_2_common_schema.adoc
```diff
@@ -864,24 +899,26 @@ calculated
 
 _[Note: (1) If `causality = "independent"`, it is neither allowed to define a value for `initial` nor a value for start.
 (2) If `causality = "input"`, it is not allowed to define a value for `initial` and a value for start must be defined.
-(3) If \(C) and `initial = "exact"`, then the variable is explicitly defined by its start value in Initialization Mode (so directly after calling `fmi3ExitInitializationMode`, the value of the variable is either the start value stored in element `<ScalarVariable><XXX start=YYY/>` or the value provided by `fmi3SetXXX`, if this function was called on this variable).]_
+(3) If \(C) and `initial = "exact"`, then the variable is explicitly defined by its start value in Initialization Mode (so directly after calling `fmi3ExitInitializationMode`, the value of the variable is either the start value stored in element `<Variable><XXX start=YYY/>` or the value provided by `fmi3SetXXX`, if this function was called on this variable).]_
```
t-sommer: This should be `<Variable start="x"/>` (with double quotes because it cites an imaginary XML tag)

docs/2_2_common_schema.adoc
```diff
@@ -934,7 +975,7 @@ _Discussion of the combinations that are [underline]#not allowed#_:
 |_The combinations `"constant" / "parameter"`, `"constant" / "calculatedParameter"` and `"constant" / "input"` do not make sense, since parameters and inputs are set from the environment, whereas a constant has always a value._
 
 ^|_[red]#(b)#_
-|_The combinations `"discrete / parameter"`, `"discrete / calculatedParameter"`, `"continuous / parameter"` and `"continuous" / "calculatedParameter"` do not make sense, since `causality = "parameter"` and `"calculatedParameter"` define variables that do not depend on time, whereas `"discrete"` and `"continuous"` define variables where the values can change during simulation._
+|_The combinations `"discrete / structuralParameter"`, `"discrete / parameter"`, `"discrete / calculatedParameter"`, `"continuous / structuralParameter", `"continuous / parameter"` and `"continuous" / "calculatedParameter"` do not make sense, since `causality = "structuralParameter"`, causality = "parameter"` and `"calculatedParameter"` define variables that do not depend on time, whereas `"discrete"` and `"continuous"` define variables where the values can change during simulation._
```
t-sommer: This should be reformulated in plain english (is or equals vs. =)

docs/2_2_common_schema.adoc
```diff
@@ -1032,14 +1081,14 @@ Even if this Dirac impulse would be modeled correctly by the modeling environmen
...
 
 _FMI for Model Exchange:_ +
-_Therefore, "tuning a parameter" during simulation does not mean to "change the parameter online" during simulation.
+_Therefore, "tuning a (structural) parameter" during simulation does not mean to "change the parameter online" during simulation.
 Instead, this is a short hand notation for:_
 
 . _Stop the simulation at an event instant (usually, a step event, in other words, after a successful integration step)._
 
-. _Change the values of the tunable parameters._
+. _Change the values of the tunable (structural) parameters. For tunable structural parameters, the "reconfiguration Mode" state must be entered before and left afterwards._
```
t-sommer: Are we using double quotes for states? If so this should go in the contribution guide.

docs/2_2_common_schema.adoc
```diff
@@ -1032,14 +1081,14 @@ Even if this Dirac impulse would be modeled correctly by the modeling environmen
...
 
-. _Compute all parameters that depend on the tunable parameters._
+. _Compute all parameters (and sizes of variables, states, derivatives, zero crossings, ...) that depend on the tunable (structural) parameters._
```
t-sommer: Maybe we should stick with "event detection" vs. "zero crossing"?

docs/2_2_common_schema.adoc
klausschuch: Just checked, for all other xml tags we don't use angle brackets in the schema description to refer to the tag, e.g.  we write `ModelVariables` instead of `<ModelVariables>`.
It might be a good idea to use angle brackets in the specification whenever an xml tag/element is referred.
For variables we have to consider that we should introduce a meta-tag `Variable`. For this meta-tag wie should not use the angle brackets but maybe curly brackets to not confuse it with an actual tag/element.

docs/2_2_common_schema.adoc
klausschuch: I just fixed the typo here (two closing " instead of one "). This is part of a list of all allowed attribute values, where each value is written in "value".
I suggest that this pull request just contains the missing changes of the doc file.
This is a more general style issue.

docs/2_2_common_schema.adoc
klausschuch: I missed the newly introduced inheritance for variables here.
Not sure what you mean with double quotes here. Can you change it like you think it should be done?
Would this be ok: `<{Variable} start="x" />`
(see discussion above about imaginary xml tag)

docs/2_2_common_schema.adoc
klausschuch: right, it should be "event indicators"

docs/2_2_common_schema.adoc
klausschuch: In other sections, we write the states without quotes but in bold letters. (I will fix it)

General topic:
We should try to avoid state names that consist of more than 1 word.
Additionally, we should either call it mode or state. Currently this is quite messy.

docs/2_2_common_schema.adoc
klausschuch: I did this part before our decision to use plain english.
However, switching to plain english seems to be a bigger topic. (`=` for attribute values is used at several places). I think we should not mix up both changes. This pu request should only cover the content that was missing to cover arrays. There should be a general issue to harmonize equal vs `=` in the specification.

docs/2_2_common_schema.adoc
t-sommer: Actually this should be `{VariableType}` as a placeholder for Float64 etc.
To me the curly braces look a bit confusion but that's a matter taste.
We should define the placeholders in a central place (including all possible values).

docs/2_2_common_schema.adoc
klausschuch: Actually, my last comment is unfortunately not true: State names are not consistently written in bold in other sections. (sometimes bold, sometimes not)

I suggest to write state names bold.

Issue #420, opened 2018-10-17
Who shall define the prefix for source code FMUs?


modelica-trac-importer:
**Reported by cbertsch on 5 Jun 2017 15:26 UTC**
I have a perhaps naive question regarding the prefixing of FMI-Functions in source code FMUs.

Citing from the FMI 2.0 Standard p.14:
```
The goal is that both textual and binary representations of FMUs are supported and that several FMUs
might be present at the same time in an executable (for example FMU A may use an FMU B). In order
for this to be possible, the names of the functions in different FMUs must be different or function pointers
must be used. To support the first variant macros are provided in “fmi2Functions.h” to build the actual
function names by using a function prefix that depends on how the FMU is shipped. Typically, FMU
functions are used as follows:
// FMU is shipped with C source code, or with static link library
#define FMI2_FUNCTION_PREFIX MyModel_
#include "fmi2Functions.h"
< usage of the FMU functions >
```

I am wondering, if the prefix shall be defined in the source code (as I have done it in the reference FMUs now) or by the importing tool in the compile command with something like
```
cl ... /DFMI2_FUNCTION_PREFIX=MyModel_ source_file.c ...
```
(Visual Studio specific: (https://msdn.microsoft.com/en-us/library/aa235412%28v=vs.60%29.aspx) or -D for gcc https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html)

The advantage of this solution would be, that multiple instances of the same source-code FMU could be compiled and that the same C-Code code could also be used to create a dynamic link library (e.g. for a different OS)

Or do I get something wrong?

----

Migrated-From: https://trac.fmi-standard.org/ticket/420

modelica-trac-importer:
**Comment by andreas&#x2e;junghanns on 5 Jun 2017 15:31 UTC**
I agree that both versions should be possible from the same source code FMU.
We have so far failed to define exactly how this can be accomplished and it takes looking at the sources and figuring it out.

This is something to be clarified/defined for the next standard version.

KarlWernersson:
The specification states that the function prefix must be the modelIdentifier in the xml

> FMU is shipped with C source code, or with static link library:
The constructed function name is “MyModel_fmi2GetReal”, in other words the function name is
prefixed with the model name and an “_”. As FMI2_FUNCTION_PREFIX the “modelIdentifier”
”attribute defined in <fmiModelDescription><ModelExchange>, or
[...]

PR #549, opened 2019-03-09
Clarify that a modified macro may be used

t-sommer:
to build shared library

(see #420)

### DIFF
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -65,9 +65,9 @@ In order for this to be possible,
 the names of the functions in different FMUs must be different, or function pointers must be used.
 To support the first variant macros are provided in `"fmi2Functions.h"` to build the actual
 function names by using a function prefix that depends on how the FMU is shipped.
-_[These macros can be modified in a target specific variant of `"fmi2Functions.h"` to compile the shared library from the source code e.g. to simulate an FMU that has been exported on a different platform.]_
+_[These macros can be defined differently in a target specific variant of `fmi2Functions.h` to adjust them to the requirements of the supported compilers and platforms of the importing tool.]_
 
-An FMU C-file must include at the beginning a `define` of `FMI2_FUNCTION_PREFIX` as the `modelIdentifier` attribute defined in `<fmiModelDescription><ModelExchange>` or `<fmiModelDescription><CoSimulation>`together with `"pass:[_]"` at the end (see sections 3.3.1 and 4.3.1).
+An FMU C-file must include at the beginning a `define` of `FMI2_FUNCTION_PREFIX` as the `modelIdentifier` attribute defined in `<fmiModelDescription><ModelExchange>` or `<fmiModelDescription><CoSimulation>`together with `pass:[_]` at the end (see sections 3.3.1 and 4.3.1).
 
 This `define` must be directly follwed with an `#include "fmi2Functions.h"` statement.
 
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -65,7 +65,12 @@ In order for this to be possible,
 the names of the functions in different FMUs must be different, or function pointers must be used.
 To support the first variant macros are provided in `"fmi2Functions.h"` to build the actual
 function names by using a function prefix that depends on how the FMU is shipped.
-_[These macros can be modified to compile the shared library from the source code e.g. to simulate an FMU that has been exported on a different platform.]_
+_[These macros can be modified in a target specific variant of `"fmi2Functions.h"` to compile the shared library from the source code e.g. to simulate an FMU that has been exported on a different platform.]_
+
+An FMU C-file must include at the beginning a `define` of `FMI2_FUNCTION_PREFIX` as the `modelIdentifier` attribute defined in `<fmiModelDescription><ModelExchange>` or `<fmiModelDescription><CoSimulation>`together with `"pass:[_]"` at the end (see sections 3.3.1 and 4.3.1).
+
+This `define` must be directly follwed with an `#include "fmi2Functions.h"` statement.
+
 Typically, FMU functions are used as follows:
 
 [source, C]
@@ -88,16 +93,14 @@ A function that is defined as `fmi2GetReal` is changed by the macros to a functi
 - If the FMU is shipped with C source code or with static link library: +
 The constructed function name is `MyModel_fmi2GetReal`.
 In other words the function name is prefixed with the model name and an `"pass:[_]"`.
-As `FMI2_FUNCTION_PREFIX` the `modelIdentifier` attribute defined in `<fmiModelDescription><ModelExchange>`,
-or `<fmiModelDescription><CoSimulation>` is used,
-together with `"pass:[_]"` at the end (see sections 3.3.1 and 4.3.1).
 A simulation environment can therefore construct the relevant function names by
 generating code for the actual function call.
 In case of a static link library, the name of the library is MyModel.lib on Windows
 and libMyModel.a on Linux; in other words the `modelIdentifier` attribute is used as library name.
 
 - If the FMU is shipped with DLL/SharedObject: +
-The constructed function name is `fmi2GetReal`, in other words, it is not changed.
+The constructed function name is `fmi2GetReal`, in other words, it is not changed. 
+_[This can be realized in the case of a source code FMU with a target-specific version of `fmi2Functions.h` that modifies the macros.]_
 A simulation environment will then dynamically load this library and will explicitly import the function symbols by providing the FMI function names as strings.
 The name of the library is `MyModel.dll` on Windows or `MyModel.so` on Linux;
 in other words the `modelIdentifier` attribute is used as library name.
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -92,7 +92,7 @@ A function that is defined as `fmi2GetReal` is changed by the macros to a functi
 
 - If the FMU is shipped with C source code or with static link library: +
 The constructed function name is `MyModel_fmi2GetReal`.
-In other words the function name is prefixed with the model name and an `"pass:[_]"`.
+In other words the function name is prefixed with the model identifier and an `"pass:[_]"`.
 A simulation environment can therefore construct the relevant function names by
 generating code for the actual function call.
 In case of a static link library, the name of the library is MyModel.lib on Windows
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -65,6 +65,7 @@ In order for this to be possible,
 the names of the functions in different FMUs must be different, or function pointers must be used.
 To support the first variant macros are provided in `"fmi2Functions.h"` to build the actual
 function names by using a function prefix that depends on how the FMU is shipped.
+_[These macros can be modified to compile the shared library from the source code e.g. to simulate an FMU that has been exported on a different platform.]_
 Typically, FMU functions are used as follows:
 
 [source, C]
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -74,6 +74,10 @@ Typically, FMU functions are used as follows:
 #define FMI2_FUNCTION_PREFIX MyModel_
 #include "fmi2Functions.h"
 < usage of the FMU functions >
+----
+
+[source, C]
+----
 // FMU is shipped with DLL/SharedObject
 #include "fmi2Functions.h"
 < usage of the FMU functions >
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -100,7 +100,7 @@ and libMyModel.a on Linux; in other words the `modelIdentifier` attribute is use
 
 - If the FMU is shipped with DLL/SharedObject: +
 The constructed function name is `fmi2GetReal`, in other words, it is not changed. 
-_[This can be realized in the case of a source code FMU with a target-specific version of `fmi2Functions.h` that modifies the macros.]_
+_[This can be realized in the case of a source code FMU with a target-specific version of `fmi2Functions.h` that does not use FMI2_FUNCTION_PREFIX to construct the function names.]_
 A simulation environment will then dynamically load this library and will explicitly import the function symbols by providing the FMI function names as strings.
 The name of the library is `MyModel.dll` on Windows or `MyModel.so` on Linux;
 in other words the `modelIdentifier` attribute is used as library name.


docs/2_1_common_api.adoc
```diff
@@ -65,6 +65,7 @@ In order for this to be possible,
 the names of the functions in different FMUs must be different, or function pointers must be used.
 To support the first variant macros are provided in `"fmi2Functions.h"` to build the actual
 function names by using a function prefix that depends on how the FMU is shipped.
+_[These macros can be modified to compile the shared library from the source code e.g. to simulate an FMU that has been exported on a different platform.]_
 Typically, FMU functions are used as follows:
```
chrbertsch: Can't we explicitly write, where these FMU functions are used? Is the C-Code defining the FMU?

docs/2_1_common_api.adoc
```diff
@@ -65,6 +65,7 @@ In order for this to be possible,
...
 
 [source, C]
```
chrbertsch: The following options are not exclusive. What if the FMU shipped with a DLL and the source code?

chrbertsch:
Do I understand the following right:

- If an FMU is shipped with source code, the prefix should be defined within the C-Code.
- But in order to built a dynamic library from the included C-code, one should not use the prefix.
- This is done by changing the fmi2functions.h file. (Shouldn't we provide an example for the modified file for this use case?)

pmai:
Sorry, I was not present at the last design meeting, so I don’t know what exactly was decided, and from the proposed changes I still don’t understand what the standard is supposed to say here, so I think the wording has to be made quite a bit clearer, if we want implementers to do the right thing...

Just a note: Currently nothing really requires the FMU source code to actually use the macros from fmi3Functions.h to define its exported symbols. If we now want unspecific changes to those macros to have a defined effect, we will have to specify this in much more detail.

I don’t see how this is going to be better or more robust than changing FMI 3 so that the supplied source code should not define a prefix, but must respect a defined FMI3_FUNCTION_PREFIX. In that way the importer can chose whatever prefix it wants, and DLL/SO compilation is as simple as possible. This would be much easier to specify unambiguously and would not rely on any unspecified changes to the headers.

But again, whatever was decided needs to be documented very explicitly, or people will get it wrong...

andreas-junghanns:
The decision was to leave everything as is. Just explain better, how to use the macros so the prefix-on/off is better understood. And example might help.

masoud-najafi:
Do we really need to change fmi2Functions.h? If we want to have different hearder files for static and dynamic linking (instead of adding /DFMI2_FUNCTION_PREFIX=MyModel_  compiler argument) , then modifying "fmi2TypesPlatform.h" would not be better option?
For example by adding this line to "fmi2TypesPlatform.h" for dynamic libaray linking?
#undef FMI2_FUNCTION_PREFIX

t-sommer:
My intention was to give implementers a hint that they can do it this way. AFAIK the documentation only talks about changing typedefs (but it's already perfectly legal to use whatever header file you want).

masoud-najafi:
So the question is which header file is allowed to get customized?

andreas-junghanns:
fmi2TypesPlatform.h is  - as the name says - meant to be platform dependent.

The macro FMI2_FUNCTION_PREFIX can be set as compile flag outside (e.g. /D) - no need to change any header file for that.

Please re-read fmi2Functions.h - this describes what we intended. If this makes it clear, then this explanation should be in the standard document, somewhere.

pmai:
I think the clarification (I'd argue it is actually a change, but that change makes sense) needs to state that FMU source code MUST _not_ define FMI3_FUNCTION_PREFIX internally, and MUST not otherwise hard-code prefixes, but MUST respect an external FMI3_FUNCTION_PREFIX definition and use the appropriate resulting symbol names (without requiring it to use fmi3Functions.h or the name mechanisms therein).

This would allow an implementation to compile a source code FMU to a static library using a prefix of its choice (if we losen the restriction that the provided prefix must be identical to the model identifier in the case that an FMU is being compiled by the importing implementation), and it can compile the same source FMU to a shared object/DLL FMU without a prefix by just not defining the FMI3_FUNCTION_PREFIX.

Even if this was the intent all along, I think the wording in FMI2.0 really leads to source FMU code hard-coding the prefix internally.

masoud-najafi:
The motivation was the presence of a few FMU with source code on the GitHub (e.g.,
https://github.com/modelica/fmi-design/blob/fmusdk3/Reference_FMUs/src/normal/fmu_boolRef/all_ME.c)
that define FMI2_FUNCTION_PREFIX inside the source code (all_ME.c) that prevents the compiler from generating dynamic libraries. So should we forbit this kind of declaration and clarify it in the spec or change the FMI header file to generate dynamic libraries?

chrbertsch:
I agree with Pierre. 
In fact, in the the draft "reference FMUs" that Masoud refers to, the prefix in the source code was not present in the original commit, but was later introduced because it was explained to me that the FMI 2.0 standard would require this.
So for FMI 2.0.1: can we clarify it that the prefix shall not be present in the C-Code? Or would this be the suggestion for FMI 3.0? What would we then write for FMI 2.0.1?

andreas-junghanns:
This is precisely why we produce bug-fix versions: clarifications of this kind can only help, why not add them?

This is clearly one of those cases when everyone writing the standard knew what was meant, but readers might not "get it".

IZacharias:
I disagree with the suggestion that the prefix in the source code must not be set. On the contrary, it was the intention in FMI2.0 that the prefix must be set in the FMU. This should not change for 2.0.1 and 3.0, as discussed at the Design Meeting in Regensburg.
The specification should therefore read:

> FMU source code MUST define FMI3_FUNCTION_PREFIX internally before including fmi3Functions.h. As FMI3_FUNCTION_PREFIX, the "modelIdentifier" attribute defined in \<fmiModelDescription>\<ModelExchange> or \<fmiModelDescription>\<CoSimulation> must be used, together with "_" at the end. 
FMU source code MUST not use FMI3_FUNCTION_PREFIX internally, i.e. the function names used in the FMU source code always begin with "fmi3", e.g. "fmi3GetFloat32". The actual names are defined by the definitions in fmi3Functions.h and are therefore given by the importer.

Thus the following requirements are fulfilled:

1. An FMU allows the importer to compile the FMU source code in such a way that the default prefix (=modelIdentifier+"_") is used for the function names. The importer does not have to set this prefix FMU-specific.
(This can be particularly useful if several FMUs are linked statically together. In this case the same header file can be used and no FMU-specific macro values have to be set by the importer.)

1. An FMU allows the importer to compile it in such a way that a function prefix defined by the importer is used.

1. An FMU allows the importer to compile it in such a way that the function names are not prefixed.


Example code of a fmi3Functions.h that can even cover all cases:
```
#IF USE_PREFIX_AS_DEFINED_IN_FMU
/* Importer wants to use FMI3_FUNCTION_PREFIX as defined in the FMU source code */
  #define fmi3Paste(a,b) a ## b
  #define fmi3PasteB(a,b) fmi3Paste(a,b)
  #define fmi3FullName(name) fmi3PasteB(FMI3_FUNCTION_PREFIX, name)
#elseif USE_IMPORTER_PREFIX
/* Importer wants to use its own prefix */
  #define IMPORTER_FUNCTION_PREFIX someNicePrefix_
  #define fmi3Paste(a,b) a ## b
  #define fmi3PasteB(a,b) fmi3Paste(a,b)
  #define fmi3FullName(name) fmi3PasteB(IMPORTER_FUNCTION_PREFIX, name)
#else
/* Importer wants to use no prefix. This must always be the case for DLL/shared object  */
  #define fmi3FullName(name) name
#endif
```

pmai:
Probably I'm missing something, but an implementation that is compiling source FMUs has to have a defined compiler, and is thus always able to provide a defined function prefix via command-line flag, so I think a source FMU compiling implementation is in control of the prefix to be used, if it wants to, IFF we specify that the FMI3_FUNCTION_PREFIX macro has to be heeded by the source FMUs source.

In order for the more complicated version noted above to work, we would first have to specify that an FMU has to include fmi3Functions.h at all (it currently doesn't have to, and for IMHO good reasons), must ensure that the fmi3Functions.h that is used is the one of the importer (i.e. must not shadow with own filenames, and hope that importer sets this up properly), and must use the convenience macros to define the entry point names (which might not be workable for the code involved, depending on how they do it/platform issues-> special declarations).

This means many more restrictions on the source code, more complexity, with the only added benefit that a source-code importing implementation (that already must know how to call a compiler) does not have to provide a simple command-line flag to define the FMI3_FUNCTION_PREFIX macro with a default prefix.

So I don't think this approach is going to be as useful as the other one (only caring about 3.0 here, because I think 2.0.1 can't be changed).

Maybe merits further discussion in today's webex, however I won't be able to attend from the beginning due to a clash with another webex.

IZacharias:
Of course, both proposed approaches would work for importers and exporters, but why do we want to change a procedure that has worked for years? I still do not see the benefit. I think it would be confusing for tools that support 2.0 and 3.0 if with 3.0 we "reversed" the logic of the handling of FMI_FUNCTION_PREFIX.

I don't understand why the established procedure should lead to more complex source code than when FMI_FUNCTION_PREFIX is set by the importer? For the currently used method, only a few lines need to be present in source code FMUs to define the FMI_FUNCTION_PREFIX and include fmi3Functions.h. And the interface functions must be provided with the names specified in the FMI specification. That's all.
If FMI_FUNCTION_PREFIX is defined externally, the source code FMU itself must ensure that the names of the functions are generated with the help of the predefined prefix.

Unfortunately, I cannot attend the meeting this afternoon.

chrbertsch:
@IZacharias : it seems there is no consensus on the "established procedure" even for FMI 2.0 ...
Pierre Mai plans to attend starting 16:00, so I  propose this topic then (in case you might be able to partially attend the meeting)

chrbertsch:
Regular design meeting:

Currently it is not defined, if the prefix is defined in the source code.

Pierre: If it is defined in the source code, there is no way to undefine it.
It is not clearly defined, if you have to include the external header files.
If we would continue with the propsoal to change the header files, we would have to much more clearly define it in the standard text.
Defining the prefix externally would be much more complicated.
Defining the prefix externally is the simpler solution, would support more use case

Torsten S.:  The most important thing is that it is clearly defined. Technically, defining the prefix externally is better

(Assuming current status: prefix defined internally, headers must be included, in the headers one could change it)

Opinion of the participants of this meeting: 
Define prefix externally: Christian B., Masoud, Torsten S., Pierre, Karl, Kaska
Define prefix inside FMUs (reconfigure by exchanging header file):
Abstain: Klaus

Commment we need to discuss this with @IZacharias and @TorstenBlochwitz  and other interested people. We should schedule it for the next webex

APillekeit:
To be honest, I'm a bit surprised about this discussion. Why don't we stick to the decision that was made at the FMI design meeting in Regensburg? 
There are pros and cons for both solutions. But I can not see any new argument in this discussion thread that was not already discussed in Regensburg. So please specify: What is the new point that hasn't been discussed in Regensburg already that is important enough to reevaluate the decision already made at the FMI design meeting in Regensburg?
I have the feeling that discussing already made descisions again and again is using up time and resources that could have been spend on other currently open topics.

t-sommer:
> Why don't we stick to the decision that was made at the FMI design meeting in Regensburg?

Is this decision documented anywhere? I can't find it in the [meeting minutes](https://github.com/modelica/fmi-design/tree/master/Meetings/2019-03-06-FMI-Design-Meeting-Regensburg).

sjoelund:
In my opinion the default headers and options should be possible to use to compile a shared object into a regular FMU because the function names are standardised.

For a static executable, the names don't really matter as the generated executable won't be imported by any tool so it could have any function names in there as long as they are consistent and the importer knows what they will be.

Thus it should be the importer that is responsible for setting FMI2_FUNCTION_PREFIX (the 2.0 specification does not state if it is the importer or exporter who should set FMI2_FUNCTION_PREFIX; only that it should be set sometime before including the header).

APillekeit:
@t-sommer: Thank you for pointing out that the FMI meeting minutes would profit from a greater level of detail. You are probably aware that this is a general problem with most FMI meeting minutes. If you have a good suggestion for solving this I'm sure this will be implemented.
Please refer to the following lines 280-282. You have been part of the discussion, I have been part of the discussion, others too. The outcome of the approximately 30 minutes of discussion is condensed in these three lines. But we can of course also repeat this discussion in this thread and in design web meetings and at the next face-2-face FMI Design Meeting and it might be also a good idea to involve the FMI Steering Committee in this matter.

@sjoelund: The specification includes the following text: 
Typically, FMU functions are used as follows:
// FMU is shipped with C source code, or with static link library
#define FMI2_FUNCTION_PREFIX MyModel_

This has been understood by all source code FMU exporting tools I know in a way that the define would have to be set in the source file. Recently a discussion has started that this is not clear enough in the specification - which is true! This has to be fixed! - And therefore, the result of the discussion in Regensburg was as described above: A more detailed description is required for FMI 2.0.1 and 3.0.

sjoelund:
For years, OpenModelica has not been setting it in the source code...

pmai:
We are discussing this as part of a Pull Request, so really, we need to be specific which changes to the standard should happen exactly to solve the problem:

If we only "clarify" that source code FMUs must define the prefix in their source file(s), then there is no reliable way of changing or eliminating the prefix from the outside, since currently no requirement is in the standard that fmi3Functions.h must be included and that the macros defined in their need to be used to define external entry points.

So we would need at least Irina's proposal (which goes beyond what I understand what was decided in Regensburg), which specifies that the header files need to be included and the macros defined in them need to be used to define the external entry points. Only in this way the importer has a, somewhat complex, way of overriding the prefix from outside.

This BTW makes source FMU prefix handling different from other FMUs, since requiring the use of fmi3Functions.h for non-source FMUs is neither useful nor really possible (it would preclude binary FMUs produced from different languages or through direct compilation).

While we can decide to follow this route (and someone then has to update the PR to actually do all of this), I'm trying to understand the rationale for this much more complex and invasive approach, since I sadly could not be present in Regensburg. Putting this rationale in the standard as non-normative text would probably also help implementers in doing the right thing.

t-sommer:
FWIW, in `fmi3Functions.h` (line 5) it's stated that the header file *must* be used to compile FMUs:

> This header file must be utilized when compiling a FMU.

chrbertsch:
> 
> 
> FWIW, in `fmi3Functions.h` (line 5) it's stated that the header file _must_ be used to compile FMUs:
> 
> > This header file must be utilized when compiling a FMU.

We need at least a clarification anyway, as I would not consider comments in header files as normative.

chrbertsch:
Comments from regular design webmeeting today (see https://github.com/modelica/fmi-design/tree/master/Meetings/2019-04-12-regular-design-meeting)

Andreas P: Both solutions work work, why change?
Torsten S.: It kind of works out of accitents
Christian: There is not even a common understanding for FMI 2.0
Andreas P.: A lot of tools have a common understanding
Torsten S.: It is nowhere stated in the documentation that one has to use the header files
Torsten S: for 2.0.1 we should clarify it
Anreas P.: Why change? One has either to change the header files or the compiler calls?
Torsten S: There was another issue raised by Pierre: What if two FMUs have the same prefix?
Irina: With the current approach: You could overwrite it?
Torsten S: One could set different prefix
Andreas: One could change this in the header files
Torsten B: As generating tool vendors: with the current solution, it is not allowed to ship the standard header files witin the FMU. This made it complicated e.g. for compiling dynamic libraries for other platforms like targets. (We had to ship the header files )
Antoine: An FMU should not need additional header files
Christian B.: for FMI 2.0 it is clearly defined to provide the header files
Andreas P.: should we get rid of compiler specific partf of the header files
Christian: We should define, that the FMU has to include header files of the specified names, the FMI standard provides default files, one can replace them. this is unrelated of where to define the prefix.
Masoud: How do exporters do it?
In source Code: Maplesim, Dymola, SimX, AMESim, Target Link
Not in source code: OpenModelica, Activate, FMU-SDK (?)
Christian: For FMI 2.0.1 we should recommend to define the prefix in the source code
Karl: There are strong opinions, let’s prepare for the next design meeting how to do this in FMI 3.

chrbertsch:
I proposed wording for decision a the design meeting in https://github.com/modelica/fmi-standard/pull/549. Please inspect and comment.

chrbertsch:
Decision to merge this in Webmeeting.

PR #550, opened 2019-03-10
Replace hard-coded section numbers with cross-references

t-sommer:


chrbertsch:
Looks good, I made some more changes.
References to figures shall be changed separately

PR #555, opened 2019-03-11
Fix cross references to tables and sections

t-sommer:


chrbertsch:
Looks good. 
(There are still many references to "above table" in the text, but including numbers + links would not increase readability much.)

PR #566, opened 2019-04-05
Apply common naming conventions to C API

t-sommer:
to improve consistency, readability and self documentation:

- functions: camel case, imperative mood
- parameters: camel case, verbose names, plural for arrays
- types: pascal case


pmai:
Seems to be missing the relevant renaming of fmi3FMUstate to fmi3FMUState in types -> build failure.

KarlWernersson:
Hi some comments

Re-name variables of fmi3GetDirectonalDerivatives()
This is not ok and wrong, you do not send in derivatives.
The current naming is constend with the description in the standard and I was unable to figure out a better name. Possibly if you want to be more verbose you could talk about deltaKnowns and deltaUnKnowns. But the current naming, is good, changing the valueRefrence fro VrUnknown to UnKnown is good, as long as we don’t replace the fmi3Valureference with an uInt type (se vr->valueReference further down)


component->instance
I agree that the name instance is better than component, however since component is well established and as long as nothing else changes I would prefer to keep it as is. Otherwise we might need some comment about the name change in the standard and that feels like unesseary. I wouldn’t want to read, “it was previous called x but now y, it is still the same but this name is better so we changed it”, if we can avoid it. 
if we do change it, the componentEnvironment in messageLogger should also change

vr->valueRefernce,
A reason to keep vr, is that we have the type fmi3ValueReferece which then is self-explanatory, so writing fmi3ValueReference valueReferences is long and unessesary.. However with the new types we could change the type to uint32 or posilby 64 and then the verbose name would be good to have.

Value[], size[], order[],  -> Values[], sizes[], orders[],
I prefer the singular, but maybe just because I am used to it, but I don’t have a strong opinion in this matter and is ok with changing it if that is the general consensus.

Fmi3stepFinished ->fmi3CallbackStepFinished
The plan is to scarp asynchronous doStep, as I Understand this function would then also be unnecessary and should then be removed, unless I have mistaken something.

“x_nominal” to “nominals”
you lose some info by removing the x_ , a more correct name could be stateNominals, or continuousStateNominals (the last word is valid Swedish/German grammar but could look bad to a native English speaker )

“FMUstate” to “FMUState”
“logger” to “logMessage”
parameters of fmi3GetVariableDependencies()
Remove “fmi3” prefix from parameter name

All OK and fine by me.

t-sommer:
> Re-name variables of fmi3GetDirectonalDerivatives()
This is not ok and wrong, you do not send in derivatives.
The current naming is constend with the description in the standard and I was unable to figure out a better name. Possibly if you want to be more verbose you could talk about deltaKnowns and deltaUnKnowns. But the current naming, is good, changing the valueRefrence from VrUnknown to UnKnown is good, as long as we don’t replace the fmi3Valureference with an uInt type (se vr->valueReference further down)

You're right. I've update the PR.

> “x_nominal” to “nominals”
you lose some info by removing the x_ , a more correct name could be stateNominals, or continuousStateNominals (the last word is valid Swedish/German grammar but could look bad to a native English speaker )

From the function name `fmi3GetNominalsOfContinuousStates()` it should be clear that `nominals` refers to the nominals of the continuous states.

chrbertsch:
We have not discussed this at the design meeting in Renningen. So we should continue the discussion here in the ticket.

If we go on with this PR: should it be merged soon, or only after the merger of requests like FCP007?

KarlWernersson:
Hi in gerneral everything is fine however. I did not get a response to some of my questions. The ones i am most curious about are:

> component->instance
> I agree that the name instance is better than component, however since component is well established and as long as nothing else changes I would prefer to keep it as is. Otherwise we might need some comment about the name change in the standard and that feels unnecessary. I wouldn’t want to read, “it was previous called x but now y, it is still the same but this name is better so we changed it”, if we can avoid it.
> if we do change it, the componentEnvironment in messageLogger should also change
> 

> 
> vr->valueRefernce,
> A reason to keep vr, is that we have the type fmi3ValueReferece which then is self-explanatory, so writing fmi3ValueReference valueReferences is long and unnecessary.. However with the _new_ types we could change the type to uint32 or posilby 64 and then the verbose name would be good to have.

I missed this discussion in renningen, Was it discued and decided then?

t-sommer:
> I did not get a response to some of my questions.

Sorry, @KarlWernersson, here you go...

> component -> instance

A _component_ is defined as "a part or element of a larger whole, especially a part of a machine or vehicle" (https://www.lexico.com/en/definition/component) whereas _instance_ is defined as "[...] a concrete occurrence of any object, existing usually during the runtime of a computer program [...]. The creation of an instance is called instantiation." (https://en.m.wikipedia.org/wiki/Instance_(computer_science)). The result of `fmi3Instantiate()` should consequenlty be an "instance", not a "component".  IMHO this is a necessary correction and not a renaming for the sake of renaming (see also the discussion GUID vs. instantiationToken).

> vr -> valueReference

We use verbose names for every parameter except for the value references, so for consistency, I suggest to name it `valueReference`. Note that this is just a parameter name. You can keep using `vr` in your code.

KarlWernersson:
@t-sommer  Thanks =)
I can agree with the  instance change.
I do not agree with the vr->ValueReference change, however its not that important to me. So if it is the general consensus that this change is necessary, I will not block it =)

chrbertsch:
Regular Design Webmeeting: 
no objections to merging. Decision: merge.

PR #567, opened 2019-04-14
Move XML examples to separate files

t-sommer:
fix syntax and validate in CI


t-sommer:
> (Which consequence does changing from encoding="UTF8"to encoding="UTF-8" have?)

"UTF-8" seems to be the standard, so I changed it for consistency. (see e.g. https://en.wikipedia.org/wiki/XHTML#XML_declaration)

Issue #581, opened 2019-06-26
Reference RFC 2119 for wording


chrbertsch:
FMI Design Meeting:
We decided to use the wording of RFC2119 as in DCP and SSP Standard, see
https://dcp-standard.org/assets/specification/DCP_Specification_v1.0.pdf#page=104

Add text and reference to
- contributors guide
-standard text document

PR #584, opened 2019-06-26
Add RFC 2119 key words to appendix

t-sommer:
see #581


docs/0___preamble.adoc
```diff
@@ -187,3 +187,5 @@ used to define the sparseness structure of partial derivative matrices).
 In order to access an array element the C convention is used.
 For example,
 the first element of input argument `x` for function `setContinuousStates` is `x[0]`.
+
+* The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in https://tools.ietf.org/html/rfc2119[RFC 2119].
```
pmai:
```suggestion
* The key words *MUST*, *MUST NOT*, *REQUIRED*, *SHALL*, *SHALL NOT*, *SHOULD*, *SHOULD NOT*, *RECOMMENDED*, *NOT RECOMMENDED*, *MAY*, and *OPTIONAL* in this document are to be interpreted as described in https://tools.ietf.org/html/rfc2119[RFC 2119].
```

Issue #561, opened 2019-03-15
FMI 3.0: TypeDefinitions and ModelVariables have a different definition style


MartinOtter:
In ModelVariables a new variable is defined as:

```
<ModelVariables>
   <Float64 name = "var1"/>
   <Float64 name = "var2"/>
</ModelVariables>
``` 

However, in TypeDefinitions a new type is defined as:

```
<TypeDefinitions>
   <SimpleType name = "Mass">
         <Float64 unit="kg"/>
  </SimpleType>
   <SimpleType name = "Velocity">
         <Float64 unit="m/s"/>
  </SimpleType>
</TypeDefinitions>
``` 

It would be better to use the same style as for ModelVariables and define a type more compactly as:

```
<TypeDefinitions>
   <Float64 name = "Mass" unit="kg"/>
   <Float64 name = "Velocity" unit="m/s"/>
</TypeDefinitions>
```

chrbertsch:
FMI Regular design meeting: 
Makes sense to do this consistently. 
This will also change the modeldescription.xml

Is there a problem that we define a mass for 64bit and 32bit?
It must be unique over the data type. (has to be described in the documentation.)

Use same inheritance for types as for variables as proposed.

Torsten S. will try this out in fmpy and test fmus

PR #591, opened 2019-07-03
Use inheritance for TypeDefinitions and Unknowns

t-sommer:
see #561

### DIFF
--- a/schema/fmi3ModelDescription.xsd
+++ b/schema/fmi3ModelDescription.xsd
@@ -107,11 +107,16 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 						</xs:sequence>
 					</xs:complexType>
 				</xs:element>
-				<xs:element name="TypeDefinitions" minOccurs="0">
+				<!-- <xs:element name="TypeDefinitions" minOccurs="0">
 					<xs:complexType>
 						<xs:sequence maxOccurs="unbounded">
 							<xs:element name="SimpleType" type="fmi3SimpleType"/>
 						</xs:sequence>
+					</xs:complexType>
+				</xs:element> -->
+				<xs:element name="TypeDefinitions" minOccurs="0">
+					<xs:complexType>
+						<xs:group ref="fmi3TypeDefintion" maxOccurs="unbounded"/>
 					</xs:complexType>
 				</xs:element>
 				<xs:element name="LogCategories" minOccurs="0">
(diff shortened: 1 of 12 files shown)


andreas-junghanns:
As you promised: the XML shrinks signigicantly AND readabilty is better. How can we not like this!

The example already had an issue with the type here Modelica.SIunits.AngularVelocity (Float64 or Int32). You changes kept the issue intact, maybe we can fix this here right now.

t-sommer:
> The example already had an issue with the type here Modelica.SIunits.AngularVelocity (Float64 or Int32).

@andreas-junghanns, can you point me to the issue or suggest a change?

otronarp:
The #561 part of this looks great and makes a lot of sense. However, to also remove the containers for `Outputs`, `Derivatives` and `InitialUnknowns` is in my opining taking optimization one step too far. If they are an ordered lists of outputs it makes more sense to actually keep them separated as a list in their own container.

andreas-junghanns:
Otto: What is the value of this additional layer? When we discussed this, we could not see one and when I look at the examples, I like the simplicity and the removal of the strange "unknown" element.

otronarp:
@andreas-junghanns it enforces that the lists are kept together and makes it more clear that there are actually four separate ordered lists and where these lists starts and ends, which I find nice. With the proposal you can for example have the list of outputs interleaved with the derivatives.

We could still ge rid of the Unknown element by using inheritance and using the singular elements (`Output`, `Derivative`, ...) in the lists.

But I don't have a strong opinion on this, I just don't see the benefit of removing those container elements.

t-sommer:
> With the proposal you can for example have the list of outputs interleaved with the derivatives.

This is not true. The validation of this (illegal) example

```
<ModelStructure>
    <Output valueReference="805306368"/>
    <InitialUnknown valueReference="805306368"/>
    <Output valueReference="805306369"/>
    <InitialUnknown valueReference="805306369"/>
  </ModelStructure>
```

yields

```
lxml.etree.XMLSyntaxError: Element 'Output': This element is not expected. Expected is one of ( InitialUnknown, NumberOfEventIndicators
```

otronarp:
I'm sorry I didn't realized that the order was enforced by xsd, then I guess that's not an issue then.

pmai:
The original intent in making this element mandatory in 2.0 was, as far as I can tell, to make clear that the model structure MUST be supplied (i.e. it is not optional to supply it), it just happens that for very special circumstances (i.e. FMU without any outputs, and without any other necessary model structure information), it could be empty.

Which is different from the ModelStructure being optional (i.e. you could leave it out, even if you have outputs). While this is of course required already by the standard text, the XSD making the element required makes this slightly more obvious/in your face.

chrbertsch:
Webmeeting (Pierre, Antoine, Karl, Torsten S., Christian B.): recommendation to merge this.

schema/fmi3ModelDescription.xsd
```diff
@@ -141,45 +146,13 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 						<xs:group ref="fmi3Variable" maxOccurs="unbounded"/>
 					</xs:complexType>
 				</xs:element>
-				<xs:element name="ModelStructure">
+				<xs:element name="ModelStructure" minOccurs="0">
 					<xs:complexType>
 						<xs:sequence>
-							<xs:element name="Outputs" type="fmi3VariableDependency" minOccurs="0">
```
friedrichatgc: Please make such cleanups in a consistent way or skip such cosmetic changes!
Before this change everything was consistent and by this cosmetic change the document is now inconsistent.
The removed container elements Output**s**, Derivative**s** and InitialUnknown**s** are still referenced at many many places in the document.

PR #600, opened 2019-07-24
Convert pseudo code examples to C

t-sommer:
and add validation job to CI. This change aligns the code examples
with the API and provides:

- well defined syntax
- validation in CI
- convenient copy & paste


chrbertsch:
@t-sommer : you already explained to me the rationale and benefit of this proposal on the phone, but could you please add a few lines for other  reviewers? Thanks!

t-sommer:
I've updated the first comment.

APillekeit:
I think it is a good idea to make examples checkable against the specified API. That avoids unnecessary errors especially since examples may be very confusing if they do not match to the specification. The general workflow seems to be suitable.

PR #604, opened 2019-07-26
Use "discrete" as default variability for non-Float variables

t-sommer:
to be consistent with the previous rule that only Float32/64 (previously Real)
can be "continuous"


t-sommer:
Should we also merge this into `support/v2.0.x`?

APillekeit:
In my view the change is OK for FMI 3.0. For `support/v2.0.x` we should stick to a clear description of the requirement of including the variability attribute in the XML for all non-real types.

PR #606, opened 2019-07-30
Add hybrid co-simulation functions and types

APillekeit:
to headers and schema. This commit adds the technical infrastructure of
FCP007: Hybrid Co-Simulation
FCP010: Intermediate Outputs
FCP015: Intermediate Variable Access

### DIFF
--- a/headers/fmi3FunctionTypes.h
+++ b/headers/fmi3FunctionTypes.h
@@ -61,7 +61,6 @@ typedef enum {
     fmi3Discard,
     fmi3Error,
     fmi3Fatal,
-    fmi3Pending
 } fmi3Status;
 /* end::Status[] */
 
@@ -71,6 +70,23 @@ typedef enum {
     fmi3CoSimulation
 } fmi3InterfaceType;
 /* end::InterfaceType[] */
+
+/* tag::CoSimulationMode[] */
+typedef enum {
+    fmi3ModeCoSimulation,
+    fmi3ModeHybridCoSimulation,
+    fmi3ModeScheduledExecutionSimulation
+} fmi3CoSimulationMode;
+/* end::CoSimulationMode[] */
+
+/* tag::CoSimulationConfiguration[] */
+typedef struct{
+    fmi3Boolean intermediateVariableGet;
+    fmi3Boolean intermediateInternalVariableGet;
+    fmi3Boolean intermediateVariableSet;
+    fmi3CoSimulationMode coSimulationMode;
+} fmi3CoSimulationConfiguration;
+/* end::CoSimulationConfiguration[] */
 
 /* tag::DependencyKind[] */
 typedef enum {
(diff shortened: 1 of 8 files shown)


t-sommer:
Since an FMU can only run either as Model Exchange or in one of the Co-Simulation modes, wouldn't it make sense to simply extend the `fmi3InterfaceType`?

```C
typedef enum {
    fmi3ModelExchange,
    fmi3CoSimulation,
    fmi3HybridCoSimulation,
    fmi3ScheduledExecutionSimulation
} fmi3InterfaceType;
```

APillekeit:
@t-sommer : This is a valid question. There have been many discussions about this topic at the FMI design meetings in Roanne, Paderborn, Regensburg and Renningen. The so far last discussion I remember can be found here: https://github.com/APillekeit/fmi-standard/issues/2 
The resulting decision was to stay with 2 FMU interface types since the additional interface types/modes are in essence differently capable co-simulation modes. So defining them based on capability flags seemed to be more fitting. Especially since other co-simulation features like intermediate variable access rely on the same basic API and it seemed to be suitable to configure the co-simulation features all together with the same currently used approach that can be extended for future co-simulation features (even though the struct would have to be changed for that in the future with the known consequences for version numbering).

headers/fmi3FunctionTypes.h
```diff
@@ -72,6 +71,23 @@ typedef enum {
...
+typedef enum {
+    fmi3ModeCoSimulation,
+    fmi3ModeHybridCoSimulation,
+    fmi3ModeScheduledExecutionSimulation
+} fmi3CoSimulationMode;
+/* end::CoSimulationMode[] */
+
+/* tag::CoSimulationConfiguration[] */
+typedef struct{
+    fmi3Boolean intermediateVariableGet;
```
t-sommer: Boolean variables should start with or contain "is", "has" or "can". So an if-clause reads like an english sentence:

```C
if (config.intermediateVariableHasBeenSet) {
  // do something
}
```

headers/fmi3FunctionTypes.h
```diff
@@ -94,15 +122,24 @@ typedef void* (*fmi3CallbackAllocateMemory) (fmi3InstanceEnvironment instanceEnv
...
                                              void* obj);
-typedef void  (*fmi3CallbackStepFinished)   (fmi3InstanceEnvironment instanceEnvironment,
-                                             fmi3Status status);
+
+/* tag::CallbackIntermediateUpdate[] */
+typedef fmi3Status (*fmi3CallbackIntermediateUpdate) (fmi3InstanceEnvironment instanceEnvironment,
+                                                      fmi3IntermediateUpdateInfo* intermediateUpdateInfo);
+/* end::CallbackIntermediateUpdate[] */
+/* tag::PreemptionLock[] */
+typedef void       (*fmi3CallbackStartPreemptionLock)   ();
```
t-sommer: A lock cannot be started or stopped and the function name should be imperative mood. To me this mechanism looks like a mutex. Could we maybe call it `lockPreemptionMutex` and `unlockPreemptionMutex`?

headers/fmi3FunctionTypes.h
```diff
@@ -453,20 +528,19 @@ typedef fmi3Status fmi3GetOutputDerivativesTYPE(fmi3Instance instance,
 typedef fmi3Status fmi3DoStepTYPE(fmi3Instance instance,
                                   fmi3Float64 currentCommunicationPoint,
                                   fmi3Float64 communicationStepSize,
-                                  fmi3Boolean noSetFMUStatePriorToCurrentPoint);
+                                  fmi3Boolean noSetFMUStatePriorToCurrentPoint,
+                                  fmi3Boolean* earlyReturn);
```
t-sommer: The value of `earlyReturn` can be derived from `fmi3IntermediateUpdateInfo.intermediateUpdateTime` provided by the callback function, so it should be omitted here.

headers/fmi3FunctionTypes.h
andreas-junghanns: I agree with Torsten here: Let´s find better names, and, let´s not rely on the documentation explaining what the flag is supposed to do.

headers/fmi3FunctionTypes.h
andreas-junghanns: We cannot rely on the == to work on two float times. This might be the reason for the flag.

PR #611, opened 2019-08-29
Move BuildConfiguration to fmiModelDescription

t-sommer:
to avoid redundant BuildConfigurations and allow recreation of a
shared or static library from source that supports both Co-Simulation
AND Model Exchange. The build configuration is associated with the
respective interface type element via the "modelIdentifier" attribute.

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -214,6 +214,9 @@ The `BuildConfiguration` provides the necessary information to compile and link 
 |Attribute
 |Description
 
+|modelIdentifier
+|The `modelIdentifier` of the `ModelExchange` or `CoSimulation` elements this build configuration is associated with.
+
 |platform
 |Platform tuple of the platform the build configuration is intended for (e.g. `x86_64-linux`)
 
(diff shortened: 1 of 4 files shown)


chrbertsch:
Are there no cases where one needs different BuildConfigurations for ME and CS FMUs (e.g., different source files, e.g. w.r.t to the solver needed for the CS FMU?

t-sommer:
> Are there no cases where one needs different BuildConfigurations for ME and CS FMUs (e.g., different source files, e.g. w.r.t to the solver needed for the CS FMU?

I'm not aware of any (source code) FMU that supports both ME & CS and uses different file sets. However this is still possible with the proposed solution by simply using a different `modelIdentifier`.

t-sommer:
> Given the small size of the build config, why not allow them to be specific? 

A Simscape model exported with FMI Kit's grtfmi.tlc has 250+ source files, so the size can be quite significant (at least for some environments).

> Do force people to use different modelIdentifier sounds a bit like a hack.

If the binary supports both ME and CS (and thus has the same modelIdentifier) it must have been created with the same settings, so why wouldn't this work as an FMU?

PR #614, opened 2019-09-06
Feature/terminals and icons

CSchulzeTLK:
Adding terminal and icon description to modelDescription.XML


chrbertsch:
In 2.2.7: For 'matchingRule' and 'terminalKind': in order to enable layerd standards and avoid name clashes: wouldn't it be good follow the same reverse domain notation as for the extra directory, see https://github.com/modelica/fmi-standard/pull/586/commits/6e91627f8c7940170631ef9a4d677548e264e3ad ?
E.g., org.modelica.Mechanics.Translational.Interfaces.Flange_a instead of Modelica.Mechanics.Translational.Interfaces.Flange_a

In 2.2.7 I miss 
* a statement, that terminals may contain variables of different causality (i.e., the terminal can be bi-directional)

How does the user get information which terminals can be connected in a meaningnful way?
Take, e.g., a Modelica model with a translational mechanical flange a.  That could be exported as FMUs with different causalities.,  with either the force as an input or output at each of the flanges. If the graphical representation of the terminal would looks like the original Modelica port: how can a user distinguish submodels (imported FMI 3.0 FMUs with terminal) with different causalities,  and real acausal (Modelica) submodels? 
Shall we give recommendations that graphical representations of terminals with certain causalties follow  a certain guideline (e.g., including a special kind of arrow, indicating if the flow variable is an inoput or output)?


We should add an example, e.g. for a a Modelica model: provide the xml and a picture of how the imported FMU shall look like.

All these changes can be made after merging, so I propose to merge the PR.

chrbertsch:
Regular design meeting: 
@CSchulzeTLK :Recommendation after doing the minor changes listed above and then continuing the refinement on the master branch.

PR #619, opened 2019-09-18
Pass reason for entering Event Mode to FMU

t-sommer:
and use fmi3EnterEventMode to enter Event Mode from Initialization Mode.

see #589


t-sommer:
@TorstenBlochwitz, can you please give your opinion on this (as discussed at the F2F in Berlin), before I continue to adapt the docs and examples?

PR #626, opened 2019-09-24
Add documentation for clocks to common concepts

APillekeit:


CThuleHansen:
I suggest numbering the bulletins in 2.1

CThuleHansen:
`An FMI function may change settings of the process/thread in which it is called` - when an fmu is loaded it becomes part of the process of the loading tool, is that not correct?
If so, then it is contradiction with `FMI functions must not change global settings which affect other processes/threads.`

CThuleHansen:
2.1 `An FMI function may change settings of the process/thread in which it is called` - when an fmu is loaded it becomes part of the process of the loading tool, is that not correct?
If so, then it is contradiction with `FMI functions must not change global settings which affect other processes/threads.`

CThuleHansen:
2.1.1: `Example for a definition in this header file:` -> `Example of a definition in this header file:`

CThuleHansen:
2.1.1 prefix "fmi3" -> prefix `fmi3`

CThuleHansen:
2.1.1:
`definition. [Example for a definition in this header file:
typedef fmi3Status fmi3SetTimeTYPE(fmi3Instance, fmi3Real);]`

Move [...] to new line

CThuleHansen:
2.1.1: `This header file includes fmi3PlatformTypes.h and fmi3FunctionTypes.h`
Which header file? It is not indented as the rest of the description for `fmi3Functions.h`

CThuleHansen:
2.1.1: [Example for a definition in this header file [2]:
`FMI3_Export fmi3SetTimeTYPE fmi3SetTime;`]

Formatting error

CThuleHansen:
2.1.1: `in other words the modelIdentifier attribute is used as library name.` -> In case of .a it is prefixed by lib. This sentence is imprecise. 
I am not sure it is necessary.

CThuleHansen:
`FMI 2.0 improves this situation since model names are no longer used as prefix in case of DLL/Shared Objects:`
What about FMI 3.0?

CThuleHansen:
2.1.1: Maybe adapt a formal notation for all these naming details? Similar to what is used in 2.2.10

CThuleHansen:
2.1.2: `This allows to restart a simulation from a previous FMU state (see [get-set-fmu-state]).`

Well, yes and no. It can also be used to move an FMU into the future, if a state has been previously serialised. So in a sense it is a previous FMU state, but it is from a different execution...

CThuleHansen:
2.1.2: `This is a handle to a (base type) variable value of the model. Handle and base type (such as fmi3Real) uniquely identify the value of a variable. Variables of the same base type that have the same handle, always have identical values, but other parts of the variable definition might be different [for example, min/max attributes].`

I am unsure of the word value in this context. But I have the feeling, that it might have been discussed before.

CThuleHansen:
2.1.2: `This is a handle to a (base type) variable value of the model. Handle and base type (such as fmi3Real) uniquely identify the value of a variable. Variables of the same base type that have the same handle, always have identical values, but other parts of the variable definition might be different [for example, min/max attributes].`

This is the first notion and definition of base type, and it seems weird to include this definition in a "side note"

CThuleHansen:
2.1.2: `If a function in the following sections is called with a wrong` mention the section numbers instead of following sections.

CThuleHansen:
2.1.2: `These are the basic data types used in the interfaces of the C functions`
basic type or base types?

CThuleHansen:
2.1.2: `If the target environment has different definitions and the FMU is distributed in binary format, it must be newly compiled and linked with this target header file.`

This has already been mentioned. Is it necessary to duplicate the information? Number the previous entry and reference it instead.

APillekeit-PR:
> I suggest numbering the bulletins in 2.1

Not sure what you meant. Could you please provide an example?

APillekeit-PR:
> 2.1 `An FMI function may change settings of the process/thread in which it is called` - when an fmu is loaded it becomes part of the process of the loading tool, is that not correct?
> If so, then it is contradiction with `FMI functions must not change global settings which affect other processes/threads.`
> 
> I am not an expert in this, so might be an incorrect correct ;)

That very much depends on the implementation of the loading tool. But I agree, this is an explanation that can be improved. I would suggest to keep that in mind for the global rework of the specification.

APillekeit-PR:
> Globally: `Example for` -> `Example of`

done. I will update the pull request later (currently not able to log in with my account).

APillekeit-PR:
> 2.1.1 prefix "fmi3" -> prefix `fmi3`

done.

APillekeit-PR:
> 2.1.1:
> `definition. [Example for a definition in this header file: typedef fmi3Status fmi3SetTimeTYPE(fmi3Instance, fmi3Real);]`
> 
> Move [Example ...] to new line for formatting consistency

done

APillekeit-PR:
> his header file includes

done.

APillekeit-PR:
> 2.1.1: [Example for a definition in this header file [2]:
> `FMI3_Export fmi3SetTimeTYPE fmi3SetTime;`]
> 
> Formatting error

done.

APillekeit-PR:
> 2.1.1: `in other words the modelIdentifier attribute is used as library name.` under `If the FMU is shipped with C source code or with static link library:` ->
> 
> In case of .a it is prefixed by lib, and this leaves this sentence imprecise.
> Perhaps remove the sentence?

I changed it to: in other words the `modelIdentifier` attribute is used to create the library name.
OK?

APillekeit-PR:
> 2.1.1 `FMI 2.0 improves this situation since model names are no longer used as prefix in case of DLL/Shared Objects:`
> 
> Perhaps change to improved or remove all together. I think it is a leftover from version 2

I changed the whole remark to: 
_[An FMU can be optionally shipped so that it basically contains only the communication to another tool (needsExecutionTool = true, see <<CoSimulation>>).
This is particularily common for co-simulation tasks.
In this tool coupling case one DLL/Shared Object can be used for all models due to no function prefixing.]_

OK?

APillekeit-PR:
> 2.1.1: Maybe adapt a formal notation for all these naming details? Similar to what is used in 2.2.10

This might be a good idea. What is your suggestion for this notation?

APillekeit-PR:
> 2.1.2: `This allows to restart a simulation from a previous FMU state (see [get-set-fmu-state]).`
> 
> Well, yes and no. It can also be used to move an FMU into the future, if a state has been previously serialised. So in a sense it is a previous FMU state, but it is from a different execution...

I understand the comment, but I'm not sure how to fix this in the best way. 
Maybe: 
This allows to restart a simulation from a saved FMU state

OK?

APillekeit-PR:
> 2.1.2: `This is a handle to a (base type) variable value of the model. Handle and base type (such as fmi3Real) uniquely identify the value of a variable. Variables of the same base type that have the same handle, always have identical values, but other parts of the variable definition might be different [for example, min/max attributes].`
> 
> I am unsure of the word value in this context. But I have the feeling, that it might have been discussed before. And I also think that it is more than the value of a "base type"
> 
> I.e. a valuereference in contxt of `fmi3GetNumberOfVariableDependenciesTYPE` is a handle to the variable for which the number of dependencies should be returned, and therefore NOT a handle to a base type variable value.
> 
> It is also mentioned that `An fmi3ValueReference references one such scalar.`.
> So in this case it is a reference to a scalar.
> 
> For these reasons I think the definition of fmi3valuereference is imprecise

I suggest discussing this at an fmi-design meeting.

APillekeit-PR:
> 2.1.2: `This is a handle to a (base type) variable value of the model. Handle and base type (such as fmi3Real) uniquely identify the value of a variable. Variables of the same base type that have the same handle, always have identical values, but other parts of the variable definition might be different [for example, min/max attributes].`
> 
> This is the first notion and definition of base type, and it seems weird to include this definition in a "side note"

I suggest discussing this at an fmi-design meeting.

APillekeit-PR:
> 2.1.2: `If the target environment has different definitions and the FMU is distributed in binary format, it must be newly compiled and linked with this target header file.`
> 
> This has already been mentioned, so I think it would be better to number the previous entry and reference it instead.

I think it is a good idea to keep it there just to increase the awareness that changing something here will break the binary compatibility.

APillekeit-PR:
> 2.1.2: `These are the basic data types used in the interfaces of the C functions`
> basic type or base types?

the documents uses "base" type in other places. Not sure which would be to best choice: @t-sommer which one fits better?

CThuleHansen:
I will respond tomorrow :)

CThuleHansen:
> > 2.1.1: `in other words the modelIdentifier attribute is used as library name.` under `If the FMU is shipped with C source code or with static link library:` ->
> > In case of .a it is prefixed by lib, and this leaves this sentence imprecise.
> > Perhaps remove the sentence?
> 
> I changed it to: in other words the `modelIdentifier` attribute is used to create the library name.
> OK?

Yes!

CThuleHansen:
> > 2.1.1 `FMI 2.0 improves this situation since model names are no longer used as prefix in case of DLL/Shared Objects:`
> > Perhaps change to improved or remove all together. I think it is a leftover from version 2
> 
> I changed the whole remark to:
> _[An FMU can be optionally shipped so that it basically contains only the communication to another tool (needsExecutionTool = true, see <>). This is particularily common for co-simulation tasks. In this tool coupling case one DLL/Shared Object can be used for all models due to no function prefixing.]_
> 
> OK?

Yes!

CThuleHansen:
> > 2.1.2: `This allows to restart a simulation from a previous FMU state (see [get-set-fmu-state]).`
> > Well, yes and no. It can also be used to move an FMU into the future, if a state has been previously serialised. So in a sense it is a previous FMU state, but it is from a different execution...
> 
> I understand the comment, but I'm not sure how to fix this in the best way.
> Maybe:
> This allows to restart a simulation from a saved FMU state
> 
> OK?

Yes. For this case it works. But I just found another issue in the same paragraph: 
````
This is a pointer to a data structure in the FMU
that saves the internal FMU state of the actual or a previously saved time instant.
````
In cases where the state has been deserialised, then it is actually not a pointer to a data structure in the FMU. It is a pointer to the state of an FMU. 
So I think it is more accurate to state the following:
````
This is a pointer to the internal state of an FMU at a specific time instant.
````
and then avoid the rest entirely, since the pointer is just a type, not a function, and therefore it does not really "do" anything. I.e. it does not save. It is the function that performs the saving.

CThuleHansen:
> > 2.1.2: `This is a handle to a (base type) variable value of the model. Handle and base type (such as fmi3Real) uniquely identify the value of a variable. Variables of the same base type that have the same handle, always have identical values, but other parts of the variable definition might be different [for example, min/max attributes].`
> > I am unsure of the word value in this context. But I have the feeling, that it might have been discussed before. And I also think that it is more than the value of a "base type"
> > I.e. a valuereference in contxt of `fmi3GetNumberOfVariableDependenciesTYPE` is a handle to the variable for which the number of dependencies should be returned, and therefore NOT a handle to a base type variable value.
> > It is also mentioned that `An fmi3ValueReference references one such scalar.`.
> > So in this case it is a reference to a scalar.
> > For these reasons I think the definition of fmi3valuereference is imprecise
> 
> I suggest discussing this at an fmi-design meeting.



> > 2.1.2: `If the target environment has different definitions and the FMU is distributed in binary format, it must be newly compiled and linked with this target header file.`
> > This has already been mentioned, so I think it would be better to number the previous entry and reference it instead.
> 
> I think it is a good idea to keep it there just to increase the awareness that changing something here will break the binary compatibility.

Fine for me. Is it possible to create the text in one place and then including it both places? Just to avoid errors if edits are made.

CThuleHansen:
> > 2.1.1: Maybe adapt a formal notation for all these naming details? Similar to what is used in 2.2.10
> 
> This might be a good idea. What is your suggestion for this notation?

Is the postfix `_` to be part of the value of the macro definition FMI3_FUNCTION_PREFIX? Does it matter whether it is part of the macro definition or not?

I do not know. A textual rewrite could be:
- If the FMU is shipped with C source code or with static link library:
In the following paragraph _modelIdentifer_ refers to the value of the `modelIdentifier` attribute within the element `<fmiModelDescription><ModelExchange>` or
`<fmiModelDescription><CoSimulation>`, see Section 3.3.1 and Section 4.3.1.
The constructed function name must be `FMI3_FUNCTION_PREFIX_fmi3GetReal`, where `FMI3_FUNCTION_PREFIX` is a macro definition with the value _modelIdentifier__, note the appended _.
In case of a static link library the name of the library is _modelIdentifier_.dll on windows and lib<i>modelIdentifier</i>.a on Linux.
A simulation environment can therefore construct the relevant function names by generating code for the actual function call.

If you like this way of writing it, I will be happy to go over the last part of 2.1.1 and present a possible rewriting.

chrbertsch:
> the documents uses "base" type in other places. Not sure which would be to best choice: @t-sommer which one fits better?

Regular design webmeeting: 
Torsten S.: I would prefer base type.  We can fix this later.

PR #631, opened 2019-09-25
Add FMI Specification Word document

t-sommer:
from modelica/fmi-design repository


chrbertsch:
Where will we put changes to the xsd file (only documentation, no functionality)?
https://github.com/modelica/fmi-design/commit/07d9ae75b2ea1b042861f9b6899bd5c5159429d0

t-sommer:
I would not touch the XSDs, since we've decided (if I remember correctly) that the documentation in the XSDs is not normative.

chrbertsch:
> I would not touch the XSDs, since we've decided (if I remember correctly) that the documentation in the XSDs is not normative.

What is the problem of  fixing a wrong docuemtation string?

If we leave the xsd files untouched, shouldn't we write in the 2.0.1 specification docuement, that comments in the xsd files are non-normative and possibly wrong? (with an errata list, possibly only containing the version string issue)

t-sommer:
The documentation in the XSDs is inconsistent anyways (and that's why we removed it in 3.0). Therefore we could

1. leave it as it is (and maybe add a note to the specification document that the XSD docs are not normative or
2. fix the documentation in the XSDs (unrealistic) or
3. remove it altogether

I would prefer 1.

chrbertsch:
> The documentation in the XSDs is inconsistent anyways (and that's why we removed it in 3.0). Therefore we could
> 
>     1. leave it as it is (and maybe add a note to the specification document that the XSD docs are not normative or
>     2. fix the documentation in the XSDs (unrealistic) or
>     3. remove it altogether
> 
> I would prefer 1.

Thinking it through I agree to go for 1. **It would be good to get more opinions.**

Where should we add the note? Perhaps on p.5 under "Conventions used in this Document"? With a subsection headline "Accompying files"? With a text like
> "This specification document is shipped together with XML schema files that are mandatory to be used and C header files that are partially mandatory to be used (see below). Any comments and documentation text in these file shall be considered as non-normative text that may be contradicting the specification text which  has to be followed.  
> 
> Errata in FMI 2.0 .xsd files: 
> - use fmiVersion "2.0" also for FMUs compliant with this FMI version 2.0.1, see https://github.com/modelica/fmi-standard/issues/388

**Please comment on and improve this text.**

Do we know other problems of FMI 2.0 header files and .xsd files?

t-sommer:
> Do we know other problems of FMI 2.0 header files and .xsd files?

Maybe we should also mention that we've relaxed the license in the headers of the XSDs and header files to "2-clause BSD" (w/o extensions).

chrbertsch:
> Maybe we should also mention that we've relaxed the license in the headers of the XSDs and header files to "2-clause BSD" (w/o extensions).

So anyway we will ship changed .xsd files compared to FMI 2.0. Thus I would also change the documentation string regardint the FMI version string modelica/fmi-design@07d9ae7

PR #637, opened 2019-09-29
Adapt xsd files and header files

chrbertsch:
- correct documentation string regarding FMI verssion, see #388 
- Change copyright year to -2019
- Add FMI version in comment


chrbertsch:
@t-sommer : would it be possible that you test the modified header and xsd files for existing / newly generated 2.0 FMUs (e.g., with fmpy) so that we see that nothing gets broken?

PR #649, opened 2019-10-02
Update changelog

t-sommer:
[skip ci]


chrbertsch:
We should change to markdown so that the links to the tickets work. but this will be anywas the case  on the releases page https://github.com/modelica/fmi-standard/releases , right?

chrbertsch:
How shall we deliver the changelog?
Currently in RC6 it looks like this: 
image
benefit: we could later improve the description of changes, perhaps even by  known errata of 2.0.1. +
but: changes of 2.0.1 compared to 2.0 are not listed itself in the document.

But we could also include the content of https://github.com/modelica/fmi-standard/blob/support/v2.0.x/CHANGELOG.adoc in the PDF.

Or shall we include the a changelog.txt in the "complete package"

What do you think?

beutlich:
Still small typo: `w.r.t` -> `w.r.t.`

PR #651, opened 2019-10-02
Remove parameter eventTime from fmi3EnterEventMode()

t-sommer:
since the current time is passed to the FMU in fmi3SetTime() for Model Exchange
and fmi3DoStep() for Co-Simulation


CSchulzeTLK:
I don't see the benefit to know that a state event had happened, but I cannot reproduce the event indicator values. So I would have to know which event indicators had a zero crossing to make use of that information.

Detecting input events can be handled by "logging" the calls to setXXX(). This information is already available.

Providing only a boolean for time events is sufficient, since there is only one next time event known.

masoud-najafi:
@CSchulzeTLK : Indicating the EventIndicator crossing values was proposed in https://github.com/modelica/fmi-standard/issues/588
But did not get much success. 
Concerning "inputEvent", I do not know the way the simulator can distinguish a smooth input (using setXXX) and a jump in input (using setXXX). The only guy who know that there is a discontinuity in AN input is the simulator.

t-sommer:
> Indicating the EventIndicator crossing values was proposed in #588
But did not get much success.

This decision was based on the assumption that it is a minor improvement. However, as @CSchulzeTLK explained to me at the FMI Design Meeting in Lund, it can make a big difference, if a model has a large number of event indicators.

CSchulzeTLK:
No, I was wrong. Input events cannot be detected properly by logging setXXX() calls.

t-sommer:
> No, I was wrong. Input events cannot be detected properly by logging setXXX() calls.

So the proposed API would solve the problem (for input events)?

t-sommer:
@CSchulzeTLK, would a7f21049ba1f9ae2e3e348bd4a7473a16829f9ae work for you?

CSchulzeTLK:
> So the proposed API would solve the problem (for input events)?

I think this should be enough, yes.

CSchulzeTLK:
> @CSchulzeTLK, would [a7f2104](https://github.com/modelica/fmi-standard/commit/a7f21049ba1f9ae2e3e348bd4a7473a16829f9ae) work for you?

I think it is also relevant in which direction the change happened. Some FMUs might apply a hysteresis to the event indicator to avoid chattering (see p. 46 in FMI 1.0 ME spec.). If the master claims that an event had happened, but the FMU cannot reproduce the event indicators precisely, then the FMU shouldn't purely rely on the information from the master without knowing the direction.
I think it would help to define that the value is not just != 0 but rather that "+1 means zero crossing from - to +, and -1 means zero crossing from + to -". See CVodeGetRootInfo in the CVode documentation (p. 62 https://computing.llnl.gov/sites/default/files/public/cv_guide.pdf).

However, if the information is optional, then the FMU cannot rely on this. I think one approach to solve this issue is to have two event indicators per actual if condition, one with +hysteresis and one with -hysteresis. That would have to be decided during export.

Maybe more tool vendor input from the exporter side might be useful. Maybe you know someone from Lund? Or Dresden?

By the way, the section describing the hysteresis has been moved for FMI 2.0 to a separate document "FunctionalMockupInterface-ImplementationHints.docx" which is referenced in the FMI 2.0 spec, but not available to the public? This doesn't seem to be updated since 1.0. Is that reference removed from 2.0.1?

Issue #678, opened 2019-11-23
Modeldescription - main source of truth?


CThuleHansen:
Hi,

What should be considered the “most truthful” definition of the structure of the model description for fmi 3 at this point - The test fmus, the actual written standard or the xsd schemas? Or is it both xsd and written part (such that these should be in sync) or some other combination?

Cheers

nickbattle:
I agree this is important to clarify. Currently there are differences between the different sources. For example, comparing the CATIA test FMUs, HTML, docs/examples and XSDs:

The Test FMUs have SourceFile within ModelExchange/CoSimulation; the HTML and XSD describes SourceFileSets within BuildConfiguration.

The Test FMUs have ModelVariables with element names like "Int16"; the HTML agrees, but the XSD seems to define "fmi3Int16" etc, though the TypeDefinitions in the XSD are called "Int16" etc.

One example in the HTML has ModelStructure entries that have "Output" as an element name, the other has "Outputs" and "Unknown" subentries, like 2.0. The XSD has "Output"; the Test FMUs have the 2.0 style.

Many of the XMLSpy diagrams in the HMTL have yet to be changed, so for example the ModelVariables diagram still shows ScalarVariable elements, even though the XSDs are updated.

One of the example files (co_simulation_early_return.xml, under docs/examples) has "Variable" tags for ModelVariables with "Int16" subelements.

All very confusing!

pmai:
FWIW my 2cents on this: The Test FMUs will never be normative, they are going to be examples, and will be fixed if they disagree with the normative reference. In the end I would expect that either only the specification is considered normative, or a combination of spec and schema.

For now I would consider the schema as the master of all truth, given that this evolves quickly, whereas spec and examples are likely to lag, due to the effort to maintain them.

One side-note: I do not see that the schema has a discrepancy in the element names for variable types: Only the internal complexTypes are called fmi3Int16, etc., the XML element names are Int16, matching the TestFMUs.

BTW: PRs for updating the test FMUs and/or spec to match the schema are likely welcome... ;).

chrbertsch:
Currently you look at an unreleased alpha status of FMI 3.0

The Test FMUs are by no means normative.  It is great that Dassault Systems started this effort for prototyping FMI 3.0 and providing them to the community, but they might  have to be updated. 
In a future version of the Test FMUs it is planned to modify and review them to be "reference FMUs" fully compliant with the coming FMI 3.0 standard in order to test importing tools.
[...]

PR #679, opened 2019-11-26
ModelDescription - Main Source of Truth

CThuleHansen:
Clarifying whether the XSD files or the standard is to be considered normative.

Added additional information on fmiVersion value for prototyping efforts.

Added information to readme on XSD/standard truth priority.

Related to issue: https://github.com/modelica/fmi-standard/issues/678

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -20,7 +20,6 @@ https://fmi-standard.org/downloads/[Releases] and the https://fmi-standard.org/d
 == Branches
 
 - `master` -- development of the next release
-Note: During development of the next FMI version the XSD files for the `modeldescription.xml` file are to be considered normative compared to the standard text in case of inconsistencies except for comments, where the standard text is to be considered normative. Please open issues and/or PRs in this case of discovering inconsistencies.
 - `support/v2.0.x` -- maintenance of the 2.0 release
 - `feature/*` -- development of features for the next release
 
--- a/README.adoc
+++ b/README.adoc
@@ -20,7 +20,7 @@ https://fmi-standard.org/downloads/[Releases] and the https://fmi-standard.org/d
 == Branches
 
 - `master` -- development of the next release
-Note: During development of the next FMI version the XSD files for the `modeldescription.xml` file trusted more than the specification in case of inconsistencies. Please open issues and/or PRs in this case.
+Note: During development of the next FMI version the XSD files for the `modeldescription.xml` file are to be considered normative compared to the standard text in case of inconsistencies except for comments, where the standard text is to be considered normative. Please open issues and/or PRs in this case of discovering inconsistencies.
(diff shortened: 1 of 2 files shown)


CThuleHansen:
I have chosen not to write about reference and test FMUs, since these are not part of the current standard. And I think that it should be kept that way.

docs/2_2_common_schema.adoc
```diff
@@ -20,7 +20,7 @@ With the enterprise edition of XMLSpy it is possible to automatically generate C
 C# and Java code that reads an XML file of fmiModelDescription.xsd.
 An efficient open source XML parser is SAX (http://sax.sourceforge.net/, http://en.wikipedia.org/wiki/Simple_API_for_XML).
 All data from the XML file is only defined via "attributes" and not via "elements".
-Therefore, only an "attribute" handler needs to be defined for a SAX parser.].
+Therefore, only an "attribute" handler needs to be defined for a SAX parser.] except for comments. If comments within the above mentioned schema files are inconsistent with the standard text, then the standard text is to be considered normative.
```
t-sommer: There should be no more comments in the XSDs. If so, they should be removed.

README.adoc
```diff
@@ -20,6 +20,7 @@ https://fmi-standard.org/downloads/[Releases] and the https://fmi-standard.org/d
 == Branches
 
 - `master` -- development of the next release
+Note: During development of the next FMI version the XSD files for the `modeldescription.xml` file are to be considered normative compared to the standard text in case of inconsistencies except for comments, where the standard text is to be considered normative. Please open issues and/or PRs in this case of discovering inconsistencies.
```
t-sommer: IMHO the text, XSDs and header files should be normative. Since we only include text from the headers in the AsciiDoc files, I don't see how they can be inconsistent.

CThuleHansen: The XSD files can differ from the standard - can it not?

CThuleHansen: Are all the pictures of the schema elements in the standard autogenerated from the XSD? Otherwise I see the standard potentially deviating during development.
But I do not mind removing this and just keeping the text on fmiVersion

docs/2_2_common_schema.adoc
CThuleHansen: I found that there are no comments in the XSDs, so I will remove this. Will update soon

README.adoc
t-sommer: > Otherwise I see the standard potentially deviating during development.

If there are any inconsistencies (e.g. outdated screenshots) it's an issue that needs to be resolved.

chrbertsch:
Regular Design meeting: 
OK

@t-sommer : could you please merge this?

t-sommer:
Superseded by #695

Issue #512, opened 2018-12-17
Add array size information to getters / setters 


t-sommer:
Currently the getters and setters for variables only provide information about total size (`nValues`) of the memory passed to the function e.g.:

```
fmi3Status fmi3GetFloat64(fmi3Component c,
                          const fmi3ValueReference vr[], size_t nvr,
                          fmi3Float64 value[], size_t nValues);
```

However this makes it impossible to check the sizes of the individual arrays and provide meaningful error messages if they don't match. This is especially important if array sizes change during runtime.

My suggestion is to pass the sizes of the arrays as a parameter `sizes[]` (of length `nvr`):

```
fmi3Status fmi3GetFloat64(fmi3Component c,
                          const fmi3ValueReference vr[], size_t nvr,
                          fmi3Float64 value[], size_t sizes[]);
```

The total size of the memory can be calculated from the `sizes[]` by summing up it's values.

masoud-najafi:
As Andreas said, this has been already discussed. For any FMI API, the size of any vector should be given in the API arguments too.

chrbertsch:
Discussion on Meeting 1.11.2019
Karl: Was decided very early in the arrays group for FMI 2.1
Andreas J: the argument of Torsten was never discussed. If there is a mismatch, there is no way of providing diagnostic information. But goes contra the efficiency 
Andreas P: overhead only for debugging (comment by Irina)
Andreas J: could be an argument with possible value NULL
Karl: Do not see the big overhead
Pierre: Do not know if people will use it the way it is intended. Do not know if people call for several variables simultaneously
Andreas J: If setting fails, one can use logging
Pierre: only significant benefit: multiple mismatches, that result in no error. Realisticly speaking this is only a constructed case. Most users will set arrays in separate calls (more efficient)
Andreas: is the combined call useless?
Pierre: Necessary for multiple scalars ....
Torsten S: also important for dynamic callers like Java (where function calls are expensive)
Andreas J: We need a node on this.
Pierre: I can do this

Christian: Regarding the ticket: if one needs debug information, one has to call separately, right?
Andreas: ...
Pierre: we do not have an error code when the sizes to not match
Andreas: In one case the FMU could give a good logging message, in the other case not
Pierre: in neither case we do require a specific error message in the case the sizes do not match
Andreas: The simulation will stop anyway; in one case one gets information what is not fitting.
Pierre: this would be an implementation fault of the either the importer or exporter; they could debug with different ways.
[...]

PR #681, opened 2019-12-02
Update 4_2_co-simulation_api.adoc

jph-tavella:
What is the purpose of the argument nValues???


t-sommer:
This has been discussed in #512.

PR #684, opened 2019-12-05
Fix inconsistent spelling of "Model Exchange" and "Co-Simulation"

t-sommer:


andreas-junghanns:
Sorry, but I think this change set is over-ambitious. I agree with about 20% of the changes, but the rest is not right. I think we need to discuss this to find a common understanding when to spell things how. And: I found a ModelExchange that I think should be Model Exchange.

Proposal lower/upper case:
 - "a co-simulation environment": this must be lower case because it is not a referring to a specific co-simulation ("named")
 - "a co-simulation slave": tricky. Do we mean FMI Co-Simulation? Then capital. If we mean a general co-sim slave, then lower case.
 - "for co-simulation tasks": clearly lower case: this is not the specific protocol here. 

Proposal with/without space:
 - "Model Exchange FMU": this refers to the part of the standard, not the XML element
 - "ModelExchange": only if we really mean the element in the XML

t-sommer:
@andreas-junghanns, can you turn your comment into a PR and update the [formatting rules](https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#formatting-rules)?

andreas-junghanns:
As I am not sure, I was hoping for a discussion first to make sure we all agree before we go through the work of changing things and formulating formatting rules...

PR #687, opened 2019-12-06
Add xrefs and fix formatting

andreas-junghanns:
- remove excess bold, underline and italic formatting 
- run asciidoctor with --verbose switch
- extend timeout for HTML check to 60 min


andreas-junghanns:
I hope this catches most of the strange formatting and does not break too many other things. Remaining issues should be smaller and we can deal with them as they come up, hopefully.

chrbertsch:
Do we have a rule when to insert xrefs for function calls?
E.g., for functions like fmi3ExitInitializationMode we have xrefs, while for fmi3SetFloat64 we do not (yet?) have xrefs.

chrbertsch:
When are function calls to be written with "()"?
E.g., shall it be fmi3GetDirectionalDerivative or GetDirectionalDerivative()? Shall the brackets be part of the xref label?

t-sommer:
@chrbertsch, you can run asciidoctor with the `-v` switch to show broken xrefs

Issue #691, opened 2019-12-08
still strange, excessive formatting in the text


andreas-junghanns:
e.g. [underline]

andreas-junghanns:
or: *clocked model partitions*

chrbertsch:
Regular Design meeting: 

Torsten S: 
Proposal to remove all underline, bold (that is not a state name) and italic text (that is not non-normative).

Keep only double-quotes where one actually quotes.

Units shall be in set in back-ticks: `Nm`

Torsten S: After this will be merged, I will put this in the formatting rules.

chrbertsch:
Regular design Webmeeting:

There are still e.g. colors in the tables that have semantic meaning.
We should find alternatives.

andreas-junghanns:
> 
> 
> Regular design Webmeeting:
> 
> There are still e.g. colors in the tables that have semantic meaning.
> We should find alternatives.

What is wrong with the colors? I suggest we either leave it as is, or open a new issue for the colors (so we can close this).

andreas-junghanns:
I suggest (again) to drop the color issue from here and close this ticket, reopen if you disagree. If the colors seem important, let´s open a new issue to discuss alternatives to colors.

PR #694, opened 2019-12-13
Add "Font styles" section to contribution guide

t-sommer:
see #691


chrbertsch:
When we add xrefs for the states, will we add the "boldness" already there so that this will be enforced at a central point (i.e. via a definition  [[StepMode,*Step Mode*]] )?

PR #698, opened 2020-01-02
Fix formatting

t-sommer:
- apply one-sentence-per-line rule
- add underscores to every italic line
- remove excess line breaks
- fix double quotes vs. backticks
- add xrefs to literature

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -333,6 +333,16 @@ will be rendered as
 
 > A variable with `variability &#8800; constant`...
 
+=== Italic paragraphs
+
+Every line of an italic paragraph (e.g. in non-normative text) should be surrounded by underscores, so the text is highlighted correctly in code editors.
+Example:
+
+----
+_[This is the first line._
+_This is the second line.]_
+----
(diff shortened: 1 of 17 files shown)


t-sommer:
> I see footnote 10 is broken, parts of the text are missing including the closing ']':
> The notation ∙t is from [BCP10

In Chrome it renders fine:

Screen Shot 2020-01-08 at 13 08 29

andreas-junghanns:
Sorry, this is what it looks like here, in Chrome:
9. This definition is identical to the syntax of an identifier in Modelica version 3.2.
10. The notation ∙t is from [BCP10
11. This definition is slightly different from the standard definition of state events: "zj(t)⋅zj(ti−1)≤0". This often used definition has the severe drawback that zj(ti−1)≠0 is required in order to be well-defined and this condition cannot be guaranteed.

Note: even formatting for the numbers is different, the last number is not blue/bold. This might be connected to my issue. I am available for a webex to show how I build, maybe this is a tooling issue?

andreas-junghanns:
footnote now ok

Issue #716, opened 2020-01-14
Large tables waste a lot of space


andreas-junghanns:
Some of the tables are badly formatted and waste a lot of space.

By turning the left cells of a line into a heading the cells are listed in the table of content.

andreas-junghanns:
I have trouble merging master into my changes... I am not sure why I get so many conflicts. Help would be appreciated.

t-sommer:
> I have trouble merging master into my changes...

Can you share a link to the branch or open a PR?

andreas-junghanns:
PR #716 is addressing this

andreas-junghanns:
We (Torsten and I) looked at the result and did not like it. At the very least, it is not clearly better than the table solution because tables allow easier location in the text, while headings seem to get lost in the pages and pages of text.

PR #717, opened 2020-01-15
#716 tables to headings

andreas-junghanns:
I made a mess here because I took too long for all those changes and master moved away from me...


andreas-junghanns:
Torsten and I (Andreas) agree turning tables into headings does not improve readability

PR #720, opened 2020-01-16
Remove formatting from Variable and turn it into the general variable

andreas-junghanns:


andreas-junghanns:
I now realize this change set was build not on master... closing, trying again later...

PR #722, opened 2020-01-23
Adjoint derivatives

chrbertsch:
Proposal for an adjoint derivative API function according to #664


docs/2_1_common_api.adoc
```diff
@@ -1047,6 +1049,64 @@ _After each such call, the elements of the resulting directional derivative vect
...
+If the capability attribute `providesAdjointDerivative = true`, <<fmi3GetAdjointDerivative>> computes the adjoint deriatives of an FMU.
+
+_[Adjoint derivatives are beneficial in several contexts_
+
+* _in Artifical Intelligence (AI) frameworks the adjoint derivatives are called "VJP"s meaning vector-gradient-products. There adjoint derivatives are used in the backpropagation process to do gradient-based optimization of parameters using reverse mode automatic differentiation (AD), see, e.g., <<BPRS15>>._
+* _in pararameter estimation (see <<BKF17>>)_
+
+_Typically, reverse mode automatic differentiatian (AD) is much more efficient for these use cases than forward mode AD, as explained in the cited references._
+
+_If one would like to construct the full Jacibian matrix, one can use either <<fmi3GetDirectionalDerivative>> (to column-wise construct it) or <<fmi3GetAdjointDerivative>> (to row-wise construct it, possibly improved with coloring methods as mentioned above)._
```
HansOlsson: Jacibian -> Jacobian

HansOlsson: _Add_: However, for computing the full Jacobian matrix the column-wise construct is generally more efficient.
(I know that we shouldn't discuss implementation issues etc, but if we say that there are two ways of computing something and we know that one will be more efficient we should say that and not lead users down the wrong path.)

docs/2_1_common_api.adoc
```diff
@@ -1047,6 +1049,64 @@ _After each such call, the elements of the resulting directional derivative vect
...
+_However in the applications motivating the adjoint derivatives, one does not need the full Jacobian matrix latexmath:[\mathbf{J}], but vector  latexmath:[\mathbf{J}] mulitplied from the left to the Jacobian, i.e.  latexmath:[\mathbf{v^T J}] ]_
```
HansOlsson: mulitplied -> multiplied

docs/2_1_common_api.adoc
ChrisRackauckas: That depends on the sparsity pattern of the Jacobian.

docs/2_1_common_api.adoc
```diff
@@ -1108,6 +1110,63 @@ To get the directional derivative of latexmath:[{Y}] with respect to latexmath:[
...
+
+* _in pararameter estimation (see <<BKF17>>)_
+
+_Typically, reverse mode automatic differentiatian (AD) is more efficient for these use cases than forward mode AD, as explained in the cited references._
+
+_If one would like to construct the full Jacobian matrix, one can use either <<fmi3GetDirectionalDerivative>> (to column-wise construct it) or <<fmi3GetAdjointDerivative>> (to row-wise construct it, possibly improved with coloring methods as mentioned above)._
+_However in the applications motivating the adjoint derivatives, one does not need the full Jacobian matrix latexmath:[\mathbf{J}], but vector  latexmath:[\mathbf{J}] multiplied from the left to the Jacobian, i.e. latexmath:[\mathbf{v^T J}]._
+_For computing the full Jacobian matrix the column-wise construct is generally more efficient.]_
+
+Using the same notation as for the directional derivatives for the different modes in which the FMU can be, the adjoint derivative is
```
t-sommer: > for the different modes in which the FMU can be

What does that mean? States?

docs/2_1_common_api.adoc
chrbertsch: @HansOlsson:
> _Add_: However, for computing the full Jacobian matrix the column-wise construct is generally more efficient.

Is there an explanation or reference for this?

This seems also very much to depend on the dimenstions:
https://arxiv.org/pdf/1811.05031.pdf provides examples where either one or the other has compuational advantages depending, e.g. on the dimensions n and m of the m*n Jacobian , e.g. 
grafik

ChrisRackauckas: For R^n -> R^m, forward mode is O(n) and reverse mode is O(m) with a larger constant. So if n is approximately m, forward mode will be better. But if m<<n, then reverse mode is better. When you do matrix coloring of a sparse matrix, it's the same idea except the number of colors for the column vs the row space. So for the standard case of R^n -> R^n Jacobians like in a stiff ODE solver forward mode is almost always better, in general it's much more complex.

docs/2_1_common_api.adoc
chrbertsch: > > for the different modes in which the FMU can be
> 
> What does that mean? States?

Yes. It refers to state, that we call modes like "initialization mode".
It is analogous to the text that we we have for the directional derivatives.
(Citing from https://github.com/modelica/fmi-standard/blob/master/docs/2_1_common_api.adoc:  "An FMU has different **modes** and in every **mode** an FMU might be described by different equations and different unknowns. The precise definitions are given in the mathematical descriptions of Model Exchange ([math-model-exchange]) and Co-Simulation ([math-co-simulation]). In every **mode**, the general form of the FMU equations are:"

Shall we change all of these ocurrances to "states"?

t-sommer: > Shall we change all of these ocurrances to "states"?

In this PR I would use "states" instead of "modes" to avoid ambiguities.

docs/2_1_common_api.adoc
```diff
@@ -1108,6 +1110,69 @@ To get the directional derivative of latexmath:[{Y}] with respect to latexmath:[
...
+
+* _in artifical intelligence (AI) frameworks the adjoint derivatives are called "vector gradient products" (VJPs)._
+_There adjoint derivatives are used in the backpropagation process to perform gradient-based optimization of parameters using reverse mode automatic differentiation (AD), see, e.g., <<BPRS15>>._
+
+* _in pararameter estimation (see <<BKF17>>)_
+
+_Typically, reverse mode automatic differentiatian (AD) is more efficient for these use cases than forward mode AD, as explained in the cited references._
+
+_If one would like to construct the full Jacobian matrix, one can use either <<fmi3GetDirectionalDerivative>> (to column-wise construct it) or <<fmi3GetAdjointDerivative>> (to row-wise construct it, possibly improved with coloring methods as mentioned above)._
+_However in the applications motivating the adjoint derivatives, one does not need the full Jacobian matrix latexmath:[\mathbf{J}], but vector  latexmath:[\mathbf{J}] multiplied from the left to the Jacobian, i.e. latexmath:[\mathbf{v^T J}]._
```
t-sommer: Shouldn't this be

```
... , but vector  latexmath:[\mathbf{v}] multiplied from the left to the Jacobian
```

?

docs/2_1_common_api.adoc
```diff
@@ -1108,6 +1110,69 @@ To get the directional derivative of latexmath:[{Y}] with respect to latexmath:[
...
+\Delta \mathbf{v}_{known} = \left( \frac{\partial \mathbf{h}}{\partial \mathbf{v}_{known}} \right)^T \Delta \mathbf{v}_{unknown}
+++++
+
+[[fmi3GetAdjointDerivative,`fmi3GetAdjointDerivative()`]]
+[source, C]
+----
+include::../headers/fmi3FunctionTypes.h[tags=GetAdjointDerivative]
+----
+
+- Argument `vrUnknown` contains the `valueReference`pass:[s] of the unknown variables.
```
t-sommer: The parameter names in the documentation do not match the definition in the header file.

chrbertsch: Fixed by @t-sommer 's commit https://github.com/modelica/fmi-standard/pull/722/commits/7fa71ad6ecac45e9a16719a7b285e1bddc4fd12c

friedrichatgc:
I think it's quite confusing that in the c-API for adjoint derivatives a function parameter named "**deltaUnknowns**" is, from the view of the c-API, a **known** input value and a function parameter named "**deltaKnowns**" is, from the view of the c-API, a **unknown** output value.

I propose to switch to the names "seed" and "sens" (sensitivity) for adjoint **and** forward directional derivatives. The seed vector is the vector multiplied with the jacobian and the sens vector is the result of this multiplication. Both names are commonly used in the context of automatic differentiation.

t-sommer:
> I propose to switch to the names "seed" and "sens" (sensitivity)

Thumbs up, but let's use `sensitivity` (instead of `sens`).

chrbertsch:
Why are some of the varibles and formulae  formatted in blue? Can we have all of them in black?

chrbertsch:
The variable \mathbf{u} for unknonws is problematic as we use \mathbf{u} as input variable e.g. in the mathematical model or ME FMUs. 
Currently this conflicts with the example for the adjoint derivatives where we use \mathbf{u} in that sense.

t-sommer:
> Why are some of the varibles and formulae formatted in blue?

Maybe a copy and paste error?

> Can we have all of them in black?

Yes, I think so.

chrbertsch:
@t-sommer : I could not resolve the conflict between the usage of u for unknowns and u for inputs (in the examples and, e.g., in the mathematical description of ME.)
Thus I changes back u and v to  v_unknown and v_known.
This is consisten with the usage of v for a general variable in the mathematical description of ME and CS.

The formula for the Jacobian looks a little more ugly now, but I could not find a better solution.

(I changed the colour of formulae from blue to black, not to green as stated in the commmit message above)

So I propose to merge this now. Especially the API is good, and slight improvements of the text and examples can still be done afterwards.

Issue #712, opened 2020-01-14
Clarify rules regarding get/set for numeric variables of different widths


nickbattle:
FMI3 has a range of get/set APIs for numerics, where the name of the API indicates the width of the data value concerned (Float32 or Float64, Int32, Int16 etc.)

Is the intention that it is **only** legal to use fmi3Get/Set<type> on value references whose variables are *precisely* the same type, or are (value preserving) implicit casts permitted?  For example, can I call fmi3GetInt32 on a UInt16 variable, since a Unit16 value is guaranteed to fit into an Int32? Can the array of vrefs passed to an API refer to a range of types, as long as they all fit into the type of the data array?

Can the rule(s) be clarified in section 2.1.7?

(Note, I am working together with Claudio Gomes and Casper Thule from Aarhus University, who will join the design meeting in the end of February).

chrbertsch:
Regular FMI Desing Meeting:

Torsten S: Should we really forbid it?
Andreas J: In FMI 2.0, the namespace for value reference was unique
Christian: Should we not forbid it, so that the casting is done explicitly from the outside?
Andreas J: One could do the casting inside the FMU within the FMU, kwnowing e.g. about limits
Torsten S: I would forbid it.

Torsten S: Suggestion: you must call the matching function for the precisely correct type, otherwise the return value is not defined.
Andreas: J: I would forbid it, otherwise some will implement it and other not.

Decision: 
We forbid calling such function. [The FMU should raise an error in this case.]

nickbattle:
Thank you for the clarification. Is this similar with FMI2 and GetReal/GetInteger for numeric values? (Can we legally fmi2GetReal of an Integer variable, for example?)

pmai:
In FMI 2 this is impossible since VRs are unique per base type, so that fmi2GetReal of VR 2 would get the real variable with VR 2 whereas fmi2GetInteger of VR 2 would get the integer variable with VR 2 which would be a completely different variable.

PR #726, opened 2020-01-24
Clarify access rules for variables,

t-sommer:
improve wording and fix formatting

see #712


t-sommer:
@andreas-junghanns, I intentionally not included the recommendation regarding the error handling, as this should not be done on a pre-function basis.

PR #730, opened 2020-02-03
rephrase-communication-interval

APillekeit:
rephrase "communication interval" into "two consecutive communication points" because the term "communication interval" is not defined.


chrbertsch:
There is another occurrence of "communication interval" in 2_1_...

APillekeit:
> There is another occurrence of "communication interval" in 2_1_...

please check the pull request if the change in 2.1 is OK for you. Communication step size is at least defined in the glossary.

Issue #729, opened 2020-02-03
Aliases and start values for variability constant


APillekeit:
The specification states in 2_2_common in line 1774ff:
```
. At most one variable of the same alias set of variables with <<variability>> = <<constant>> can have a <<start>> attribute _[since <<start>> variables are independent initial values.]_ 

. A variable with <<variability>> = <<constant>> can only be aliased to another variable with <<variability>> = <<constant>>.
It is then required that the <<start>> values of all aliased (constant) variables are identical.
```

If the term "alias set" has a special meaning, a definition would be helpful.
If it is just a set of aliased variables it seems that the two points descibe two different solutions for the same problem. You either allow only a start value for one variable of the set or you enforce that all variables of the set have the same start values - in case of variablity equals constant.

nickbattle:
I agree that the text is as you quote in the FMI3 spec. But the equivalent in FMI2 has a "not equal" for the first bullet. This is also how I've interpreted the rule in the VDM model. Does that help? (ie. there is a typo in FMI3 regarding the "=" sign).

2. At most one variable of the same alias set of variables with variability ≠ "constant" can have a
start attribute.

t-sommer:
@APillekeit, does this problem still exist with the definition of "alias set"?

APillekeit:
> @APillekeit, does this problem still exist with the definition of "alias set"?

Not for me, the correction of equal to unequal sign made it clear for me. Now an alias set is just a set of aliased variables as defined by language. So this is fine for me.

PR #734, opened 2020-02-05
All variables in an alias set must have the same Dimensions

### LINKED ISSUE #731: Aliases and Dimensions for array variables
Presumably all variables of in the same alias set (ie. with the same valueReference) would have to have the same Dimensions settings, if the value is an array? If so, this should be added to section 2.2.10 where the restrictions on aliases are described.

nickbattle:
Resolves #731

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1778,6 +1778,8 @@ It is then required that the <<start>> values of all aliased (constant) variable
 
 . All variables of the same alias set must all have either no `<Unit>` element defined, or all of them must have the same `<Unit name>` and the same `<Unit><BaseUnit>` definitions.
 
+. All variables of the same alias set must have the same `<Dimension>` elements, or be scalar variables.
+
 The aliasing of variables only means that the `value` of the variables is always identical.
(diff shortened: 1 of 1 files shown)


chrbertsch:
FMI regular design web meeting:

Torsten S: Can we simplify the wording? Is "alias set" defined?
Andreas P: I think "alias set" is used even in FMI 2.0
Torsten S: "Alias set" appears two times in FMI 3.0
Andreas P.: But it is not defined.
Torsten S: one should write: "all variables sharing the same value reference"
Torsten S: Lets define "alias set".
Klaus: currently the definition of "alias" is too sloppy ...  @klausschuch : I will create a ticket for this.
Andreas: Could you then also fix #729 ?
Klaus: Perhaps we should change the definition of aliases: one variable has the information of dimensions, the other alias variables only have other name. Otherwise one has to check all the attributes of the dimensions of the different variables ...
Andreas P: We have a similar topic with start values of const variables, on  p.52 of FMI 2.0
Christian: Should an alias variable then have an attribute "alias"?

t-sommer:
superseded by #748

PR #736, opened 2020-02-09
Clarify the integer type of Dimensions' start attributes

### LINKED ISSUE #733: Dimensions and valueReferenced sizes
The document says (section 2.2.10) that a Dimension size can be specified by a valueReference to a constant or structuralParameter variable.

It says that the variable must be of "integer type". Does this mean any signed or unsigned "Int" type from 64-bits to 8-bits? It later says that they must be positive integers, though perhaps that is a value constraint rather than a type constraint (ie. Int32 could be fine, but perhaps requiring a min value of >=0)?

It also says that reconfiguration changes to the Dimension via the structuralParameter are constrained by that variable's min/max. Does this mean the valueReference may not have an alias-set (since these are permitted to have different min/max values for members of the set)? Or are changes constrained by the "overall" limits of the alias-set, similar to setting values of aliased variables, where I think the most restrictive min/max of the set applies?

nickbattle:
The integer type of start and valueReference attributes of Dimensions are expected to be the same, xs:unsignedInt, ie. fmi3Uint32.

Fixes #733

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1139,11 +1139,11 @@ Variable elements can uniformly represent variables of primitive (atomic) types,
 Variable elements representing array variables must contain a `Dimensions` element specifying the array dimensions.
 The `Dimensions` element contains a sequence of `Dimension` elements, each specifying the size of one dimension of the array:
 
-- If the `<<start>> attribute of the `Dimension` element is present, it defines a constant size for this dimension, namely the integer value of the <<start>> attribute.
+- If the <<start>> attribute of the `Dimension` element is present, it defines a constant `fmi3Uint32` size for this dimension.
 The <<variability>> of the dimension size is <<constant>> in this case.
 
 - If the <<valueReference>> attribute of the `Dimension` element is present, it defines the size of this dimension to be the value of the variable with the value reference given by the <<valueReference>> attribute.
-The referenced variable must be a variable of integer type, and must either be a constant (i.e. with <<variability>> = <<constant>>) or a <<structuralParameter,`structural parameter`>>(i.e. with <<causality>> = <<structuralParameter>>).
+The referenced variable must be a variable of `fmi3Uint32` type, and must either be a constant (i.e. with <<variability>> = <<constant>>) or a <<structuralParameter,`structural parameter`>>(i.e. with <<causality>> = <<structuralParameter>>).
 The <<variability>> of the dimension size is in this case the <<variability>> of the referenced variable.
 
 These two options are mutually exclusive, i.e. for each `Dimension` element either a <<start>> attribute or an <<valueReference>> attribute can be supplied, but not both.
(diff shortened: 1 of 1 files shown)


andreas-junghanns:
What if some platform does not support such large types?

pmai:
I think the problem is that the two ways of specifying array dimensions, i.e. through values in the XML and through reference to an FMU variable have to agree on types. Since we do not have "per-platform" types for variables anymore (and never had for the XML values), I think using something like size_t is difficult to achieve (we'd need to introduce a new variable type, and think of how to handle the XML schema part).

The current proposed change just harmonizes the spec to what is in the schema. Whether we want array dimension sizes larger than uint32 is I think a seperate discussion topic.

docs/2_2_common_schema.adoc
```diff
@@ -1139,11 +1139,11 @@ Variable elements can uniformly represent variables of primitive (atomic) types,
...
 The `Dimensions` element contains a sequence of `Dimension` elements, each specifying the size of one dimension of the array:
 
-- If the `<<start>> attribute of the `Dimension` element is present, it defines a constant size for this dimension, namely the integer value of the <<start>> attribute.
+- If the <<start>> attribute of the `Dimension` element is present, it defines a constant `fmi3Uint32` size for this dimension.
 The <<variability>> of the dimension size is <<constant>> in this case.
 
 - If the <<valueReference>> attribute of the `Dimension` element is present, it defines the size of this dimension to be the value of the variable with the value reference given by the <<valueReference>> attribute.
-The referenced variable must be a variable of integer type, and must either be a constant (i.e. with <<variability>> = <<constant>>) or a <<structuralParameter,`structural parameter`>>(i.e. with <<causality>> = <<structuralParameter>>).
+The referenced variable must be a variable of `fmi3Uint32` type, and must either be a constant (i.e. with <<variability>> = <<constant>>) or a <<structuralParameter,`structural parameter`>>(i.e. with <<causality>> = <<structuralParameter>>).
```
pmai:
```suggestion
The referenced variable must be a variable of `UInt32` type, and must either be a constant (i.e. with <<variability>> = <<constant>>) or a <<structuralParameter,`structural parameter`>>(i.e. with <<causality>> = <<structuralParameter>>).
```
Better to use the XML schema element name, rather than the API type definition, I think.

docs/2_2_common_schema.adoc
```diff
@@ -1139,11 +1139,11 @@ Variable elements can uniformly represent variables of primitive (atomic) types,
 Variable elements representing array variables must contain a `Dimensions` element specifying the array dimensions.
 The `Dimensions` element contains a sequence of `Dimension` elements, each specifying the size of one dimension of the array:
 
-- If the `<<start>> attribute of the `Dimension` element is present, it defines a constant size for this dimension, namely the integer value of the <<start>> attribute.
+- If the <<start>> attribute of the `Dimension` element is present, it defines a constant `fmi3Uint32` size for this dimension.
```
pmai:
```suggestion
- If the <<start>> attribute of the `Dimension` element is present, it defines a constant unsigned 32-bit integer size for this dimension.
```

Again since we are talking about non-API stuff, better to not reference the API typedefs.

t-sommer:
@nickbattle, can you please enable "Allow commits by maintainers" or merge @pmai's suggestions yourself?

nickbattle:
> Since we use `size_t` for array sizes in the API (which maps to `unsigned long long` on 64-bit machines), shouldn't we use `fmi3UInt64` instead?

Section 2.2 has a table that defines the C mapping of xs:unsignedInt as uint32_t. So the XSD "start" attribute would have to change as well if we moved to 64-bit array sizes?

PR #739, opened 2020-02-10
Derive all state machines from a template

t-sommer:
that contains the common states and transitions


andreas-junghanns:
You merged the fatal state with the terminal state. We need to remove the fatal state from the text and describe what it means to be in the end state (just like fatal).

PR #743, opened 2020-02-10
Added table for Scheduled Execution Simulation

### LINKED ISSUE #711: Scheduled Execution state/API table missing
Section 4.2.5 defines the state machine for Scheduled Execution Simulation, which includes some new states, but there is no API table (unlike the other state machine sections) that gives clear set of rules about which APIs are allowed in which state. This would be useful and consistent.

There are "allowed calls" and "forbidden calls" listed in different states in the text, but these do not mention every API.

nickbattle:
This is a mostly empty table with the right state headings.
We need to fill it in.

Fixes #711


CThuleHansen:
The vertical text seems to be overflowing. 
image

CThuleHansen:
I have created a separate issue that addresses this, so I vote for merging this pull request. 
https://github.com/modelica/fmi-standard/issues/803

PR #746, opened 2020-02-11
valuereference is unique in FMI 3.0

friedrichatgc:
In fmi 3 the valueReference is unique across ALL variables not only across variables of the same base type.


t-sommer:
Please open PRs from your fork (not the modelica repo).

PR #752, opened 2020-02-17
Arrays are no longer flattend to variables.

friedrichatgc:
FMI 3.0 has arrays as "native" variables. Structures are still flattend
to single variables.


andreas-junghanns:
I think we do not force use of array variables, one could still flatten an array into single, atomic variables, right? If so, should we not clarify this?

friedrichatgc:
Sure, we can add that nothing must be flattened. But I think we should not state "arrays are flattened" as it is now.

PR #756, opened 2020-02-17
Adapted variables names in document to header files

friedrichatgc:
For directional derivatives the variable names in the document are
different compared to the names in the header files.


friedrichatgc:
Obsolute with PR #722.

PR #759, opened 2020-02-17
No need to restrict the clock priority to 99.

friedrichatgc:


friedrichatgc:
I removed the comment in line 705 now. I don't recognized that their is already such a TODO comment.

APillekeit:
The reason to limit to 99 was to avoid a rearranging the priority numbering for most platforms. It is very unlikly that a model needs more than 99 priority levels. If so, it is unlikely that this model will be  simulatable on most real-time simulatiors. So at least a remark should be added to the standardization. I will do that.

PR #761, opened 2020-02-17
Fixed static linked library function call.

friedrichatgc:
In a static linked library the FMI function name is prefixed with the
modelIdentifier.

### DIFF
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -89,7 +89,7 @@ Typically, FMU functions are used as follows:
 // FMU is shipped with C source code, or with static link library
 #define FMI3_FUNCTION_PREFIX MyModel_
 #include "fmi3Functions.h"
-< usage of the FMU functions e.g. fmi3SetTime >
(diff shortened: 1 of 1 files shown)


t-sommer:
This is handled by the `fullName()` macro:

```C
/* Macros to construct the real function name (prepend function name by FMI3_FUNCTION_PREFIX) */
#if defined(FMI3_FUNCTION_PREFIX)
  #define fmi3Paste(a,b)     a ## b
  #define fmi3PasteB(a,b)    fmi3Paste(a,b)
  #define fmi3FullName(name) fmi3PasteB(FMI3_FUNCTION_PREFIX, name)
#else
  #define fmi3FullName(name) name
#endif
```

friedrichatgc:
I think most production codes will import more than one FMU which will be done by having more than one
```
#define FMI3_FUNCTION_PREFIX MyModel_
#include "fmi3Functions.h"
```
in the source file, one for each FMU. The calls to FMU functions must then be done using the full name including the modelIdentifier.
We should not give here an example which works only for importing one FMU, as done by testing tools like FMPy but does not work for production codes.

PR #766, opened 2020-02-18
Some clarification of array variables regarding structured variable naming

friedrichatgc:
The section about structured variable naming describes the flattening of
arrays using the [...] syntax of structure variable naming.
Clarify that this flattening is possible but no longer required for leaf
node variables of the hierarchy when the new array are used.


t-sommer:
Should we still allow the `array[n]` style array convention, if we have propper arrays in FMI 3?

andreas-junghanns:
> Should we still allow the `array[n]` style array convention, if we have propper arrays in FMI 3?

Yes, we should. I remember the wild Modelica example with array elements of all kinds of different types. If we encounter an array with different element properties, we can currently not do this with the array type (and should not).

PR #769, opened 2020-02-18
Fixed constness of parameters in fmi3SetIntervalDecimal and fmi3SetIntervalFraction

friedrichatgc:
Missing const for input parameters


andreas-junghanns:
Can we look through all the other header files to identify other such cases or make sure we have them all ok?

andreas-junghanns:
Oh, you broke the examples!

friedrichatgc:
I detected this by looking through all the functions in the header files, but you can double check this.
I will check the examples, I think these must be adapted too.

PR #728, opened 2020-01-26
Add Hybrid Co-Simulation and Scheduled Execution Simulation to interface types

t-sommer:

PR #770, opened 2020-02-18
add abbreviations for FMI interface types 

chrbertsch:
according to discussion in #728

### DIFF
--- a/docs/0___preamble.adoc
+++ b/docs/0___preamble.adoc
@@ -93,7 +93,7 @@ This document defines the Functional Mock-up Interface (FMI),
 version 3.0 to (a) exchange dynamic models between tools and (b) define tool
 coupling for dynamic system simulation environments.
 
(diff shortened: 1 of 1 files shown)


andreas-junghanns:
I agree with the two letter abbreviations. But I know Torsten thinks that 3-letters might be better.
Torsten: How strong is your opinion here?

t-sommer:
I'm for "HCS" and "SES". I find it strange to have the "Simulation" in "CS" but not in "HC" and "SE".

t-sommer:
@andreas-junghanns and @chrbertsch, if this is okay with you, I'll update the PR.

chrbertsch:
@t-sommer : fine for me,  I even like your proposal better, e.g. to show the relation between CS and HCS. 
And the the two new types get introduced in FMI 3, so they deserve 3 letters :-)

PR #773, opened 2020-02-19
Define the path of FMU and terminal icons.

friedrichatgc:
A clear definition of the path of the attributes iconSource_PNG and
iconSource_SVG were missing.

### DIFF
--- a/docs/2_3_fmu_distribution.adoc
+++ b/docs/2_3_fmu_distribution.adoc
@@ -21,15 +21,17 @@ This ZIP file has the following structure:
 ----
 // Structure of ZIP file of an FMU
 modelDescription.xml          // description of FMU (required file)
-diagram.png                   // image file of a descriptive diagram view of the model (optional)
-icon.png                      // image file of icon without terminals (optional)
 documentation                 // directory containing the documentation (optional)
    index.html                 // entry point of the documentation
+   diagram.png                // image file of a descriptive diagram view of the model (optional)
+   diagram.svg                // if existing the diagram.png is required (optional)
    <other documentation files>
    licenses // Optional directory for licenses
       license.{txt,html}      // Entry point for license information
       <license files>         // For example BSD licenses
 icon                          // FMU and terminal icons (optional)
+   icon.png                   // image file of icon without terminals (optional)
+   icon.svg                   // if existing the icon.png is required (optional)
    // all terminal and fmu icons referenced in the graphical representation
 sources                       // directory containing the C sources (optional)
    // all needed C sources and C header files to compile and link the FMU
--- a/docs/2_3_fmu_distribution.adoc
+++ b/docs/2_3_fmu_distribution.adoc
@@ -21,13 +21,16 @@ This ZIP file has the following structure:
 ----
 // Structure of ZIP file of an FMU
 modelDescription.xml          // description of FMU (required file)
-model.png                     // image file of FMU icon (optional)
+diagram.png                   // image file of a descriptive diagram view of the model (optional)
+icon.png                      // image file of icon without terminals (optional)
 documentation                 // directory containing the documentation (optional)
    index.html                 // entry point of the documentation
    <other documentation files>
    licenses // Optional directory for licenses
       license.{txt,html}      // Entry point for license information
       <license files>         // For example BSD licenses
+icon                          // FMU and terminal icons (optional)
+   // all terminal and fmu icons referenced in the graphical representation
 sources                       // directory containing the C sources (optional)
    // all needed C sources and C header files to compile and link the FMU
    // with exception of: fmi3PlatformTypes.h, fmi3FunctionTypes.h and fmi3Functions.h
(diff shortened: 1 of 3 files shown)


pmai:
Since the importing tool needs access to the icons (and not the FMU itself), I think the paths should be specified as entries in the ZIP archive, and thus should be relative to the root of the archive, not restricted to resources and relative to that ...

This would also make it clearer how to parse the paths, since you can either treat them as relative URIs, or as direct ZIP archive entries, both alternatives giving clear processing expectations (including path separators, valid names, etc.).

Also is more compatible with the current Icon mechanism...

friedrichatgc:
You are right, making it relative to resources and talking about fmuResourceLocation and the fmi3Instantiate call is nonsense since only the importing tool needs the path.

You propose to define the path as a relative URI reference (which is defined by the absence of a scheme and authority and does not start with "/")?
E.g. icondir/fmu.png or ./icondir/fmu.png would be valid.
And we define that the base of this relative URI is the root of the zip archive?

What do you mean with "the current Icon mechanism"?

pmai:
With the current (2.0) Icon mechanism I meant the model.PNG file that you can place in the root of the FMU archive to have an icon for the whole FMU... I.e. that this also resides outside the resources path.

The relative URI proposal would make it similar to the way we handle references in SSP, which makes handling by tools easier to specify and there are lots of libraries that already know how to do URI resolution. 

If we'd go this route I would propose to speficy that the path is a relative URI that is resolved as per RFC 3986 Section 5 against the baseURI of the modelDescription.xml itself. This is morally equivalent to looking up the path (after ./.. removal as per section 5.2.4 of RFC 3986) from the root of the ZIP archive.

The wording we used in SSD was along the following lines:

> Attribute indicating the source of the parameters as a URI (cf. RFC 3986). For purposes of the resolution of relative URIs the base URI is the URI of the SSD.

For FMI I'd suggest:

> Both attributes define the source of the image file as a relative URI according to RFC 3986. The base URI that this relative URI is resolved against is the URI of the modelDescription.xml file. Implementations are only REQUIRED to support relative URIs, excluding relative URIs that move beyond the baseURI (i.e. go "up" a level via ..). Implementations are NOT REQUIRED to support any absolute URIs and any specific URI schemes.

with the following as non-normative text to point out one obvious implementation aspect:

> Note that this specification is functionally equivalent to looking up image sources from the root of the FMU ZIP Archive after dot removal from the path as per section 5.2.4 of RFC 3986.

andreas-junghanns:
Let me throw in a monkey wrench: why do we have an attribute name `iconSource_PNG`?
It has an underscore - as the only attribute in the entire xsd?
Why name it `PNG` - can´t you tell by the file name extension?

pmai:
Hehe, more fun for you tomorrow (I’m watching from family holiday sidelines;)...

A nicer approach would be to have a set of elements that can specify pairs of relative URI and MIME types, thus allowing any number of formats to be supplied, and also more future proof... Spec would then highly recommend to have SVG and PNG among the entries...

otronarp:
@andreas-junghanns there reason for the suffix in the attribute is that there are two attributes one for the SVG and one for the PNG and the PNG should be mandatory.
```
<xs:attribute name="iconSource_PNG" type="xs:string" use="required"/>
<xs:attribute name="iconSource_SVG" type="xs:string" use="optional"/>
```

chrbertsch:
> It has an underscore - as the only attribute in the entire xsd?

I would just remove the underscore and let the rest of the proposal stay as it is.

friedrichatgc:
Decision: all images for ports and icons should go into a new subdirectory in the zip file of the FMU. Then the path must be defined to be relative to this new subdirectory.

pmai:
Is there a reason why making stuff relative to new subdirectories is a useful approach? This just makes the rules for path resolution more complex... Just leave everything relative to the root (the resources folder already was a bad idea and lead to lots of implementation deviations vis-a-vis paths, fmiResourceLocation etc.); and if you follow the URI suggestion no need to define what “path” is supposed to mean (i.e. what are allowed characters, what is the path separator, are . and .. allowed, .... there are so many things to define, why bother?).

docs/2_3_fmu_distribution.adoc
```diff
@@ -21,13 +21,18 @@ This ZIP file has the following structure:
 ----
 // Structure of ZIP file of an FMU
 modelDescription.xml          // description of FMU (required file)
-model.png                     // image file of FMU icon (optional)
 documentation                 // directory containing the documentation (optional)
    index.html                 // entry point of the documentation
+   diagram.png                // image file of a descriptive diagram view of the model (optional)
+   diagram.svg                // if existing the diagram.png is required (optional)
```
chrbertsch: if  existing, ....

there should be normative text on this

docs/2_3_fmu_distribution.adoc
```diff
@@ -21,13 +21,18 @@ This ZIP file has the following structure:
...
    <other documentation files>
    licenses // Optional directory for licenses
       license.{txt,html}      // Entry point for license information
       <license files>         // For example BSD licenses
+icon                          // FMU and terminal icons (optional)
```
chrbertsch: icons

chrbertsch:
Design Meeting: 

Consensus to split between a digram and icon picture.
Poll:  Do we require a PNG if the SVG is provided?
Yes:3
No:5
Abstain:4

--> remove hard requirement

Poll
Shall we move the diagram and icon to the respective subfolders?
Yes: 8
No: 1
Abstain: 4
--> yes

PR #774, opened 2020-02-19
Only top level terminal's can have a graphical representation.

friedrichatgc:
If something else is required/wanted then a definition is required how
the name mapping from the linear list of graphical terminal
representations to the hierarchical list of terminals is defined.


chrbertsch:
Regular Design meeting

This was discussed in a short meeting of Markus, Antoine, Christian S.

Andreas J: I will merge it

PR #778, opened 2020-02-21
Fixed the description of causality indpendent.

friedrichatgc:
Removed "All variables are a function of this <<independent>> variable."
since this is not true. E.g. a constant variable is not a function of
the independent.
Moreover this will implicitly mean that in ModelStructure this
independent variable is always a dependent variable of each other variable.

Removed the text that an implicilty defined independent variable has the
name = `time`. Since the name of all variables must be unique this would
mean that we need to disallow name = `time` of all other variables when
no independent variable is defined explicitly.

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1215,10 +1215,9 @@ It is not allowed to use the variable value in another model or slave.
 
 [[independent,`independent`]]
 - `independent`: The independent variable (usually `time`).
-All variables are a function of this <<independent>> variable.
 <<variability>> must be <<continuous>>.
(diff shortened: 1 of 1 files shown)


t-sommer:
> this would mean that we need to disallow name = time of all other variables

We should mention, however, that it is strongly discouraged to use the name `time` for a variable. The name is commonly used for the x-axis in log files (e.g. for the reference results in the cross-check).

friedrichatgc:
A new issue is opened for a discussion on using time instead of independent everywhere (enforcing the independent to be always time) or to rename fmi3SetTime to fmi3SetIndependent.

PR #779, opened 2020-02-21
Do not define two entries for "input" and "output" in the causalitiy …

friedrichatgc:
…table

Each causaility should only be listed ones in the table.
Merged the two entries for "input" and "output".

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1202,12 +1202,16 @@ Allowed values of this enumeration:
 <<initial>> must be <<approx>>, <<calculated>> or not present (meaning <<calculated>>).
 
 [[input,`input`]]
+[[inputClock,`input clock`]]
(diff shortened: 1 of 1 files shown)


andreas-junghanns:
I am not sure this is right: One are normal inputs and outputs, the others are input clocks and output clocks. You can see that this is carried all the way through and the references to the clocks from the text are carefully checked.
I think this proposal goes too far.

friedrichatgc:
I don't think that I have actually changed something semantically. I just combined the two "input" causalities and the two "output" causalities in the text since I think each value of causality should only appear ones but not twice with a different description. The other possible option would be to rename the second "input" causality to "inputClock" (same for output). I think it was like this before and it changed than to a new variability "clock" which is used now to distinguish between normal inputs and clock inputs. But let's discuss this in Berlin.

t-sommer:
You cannot define more than one anchor per paragraph.

Issue #735, opened 2020-02-07
Automatic conversion from FMI 2.0 word document to FMI 3.0 adoc document seems to have a problem with unequal signs beeing changed into equal signs


APillekeit:
I don't know how big this problem is. 
But issue #729 is the second occurrence of this problem after an issue we found in the state machine description of FMI 3.0. It might be worth checking all unequal signs in FMI 2.0 and compare their occurrence in FMI 3.0.

chrbertsch:
Please compare with FMI 2.0.1. :-)
If I remember well, there was at least one bug with an exchanged equal to unequal sign in FMI 2.0 that was fixed in 2.0.1

PR #780, opened 2020-02-21
fixing two not-equal signs compared to FMI 2.0.1

chrbertsch:
I compared all &#8800; signs from FMI 2.0.1 with the current 3.0 master and found two occurances that were wrong.

fixing #735

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1776,7 +1776,7 @@ This requirement leads naturally to the following restrictions:
 _For example, if variables `a` and `b` have <<causality>> = <<parameter>>, then the value references of `a` and `b` must be different._
 _However, if variable a has <<causality>> = <<parameter>> and `b` has <<causality>> = <<calculatedParameter>> and `b := a`, then `a` and `b` can have the same value reference.]_
 
-. At most one variable of the same alias set of variables with <<variability>> = <<constant>> can have a <<start>> attribute _[since <<start>> variables are independent initial values.]_
+. At most one variable of the same alias set of variables with <<variability>> &#8800; <<constant>> can have a <<start>> attribute _[since <<start>> variables are independent initial values.]_
 
 . A variable with <<variability>> = <<constant>> can only be aliased to another variable with <<variability>> = <<constant>>.
 It is then required that the <<start>> values of all aliased (constant) variables are identical.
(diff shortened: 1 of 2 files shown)


andreas-junghanns:
One thought though: we have &#8800; used in places, and also \neq and even \ne -- should we not use one of these versions only?

chrbertsch:
> One thought though: we have ≠ used in places, and also \neq and even \ne -- should we not use one of these versions only?

Within Larger LaTeX formulae \neq or \ne make sense. (They are equivalent, right?)
For ≠ within text we use eiter `latexmath:[\neq]` or  `&#8800;` (which is unreadable in the text editor).  To which should we unify?

andreas-junghanns:
More readable are the \neq, right?
I suggest:
 - &#8800;  --> latexmath:[\neq]
 - latexmath:[\ne] --> latexmath:[\neq]
 - &#8805; --> latexmath:[\geq]
 - &#8804; --> latexmath:[\leq]

I created a PR for that.

andreas-junghanns:
The only thing I did not fix (because I am not sure there is no difference: the \ne --> \neq -- any idea?

chrbertsch:
> 
> 
> The only thing I did not fix (because I am not sure there is no difference: the \ne --> \neq -- any idea?

They seem to be totally, see e.g.
https://books.google.de/books?id=jSSSyT2SAkgC&pg=PA548&lpg=PA548&dq=Latex+%5Cne&source=bl&ots=2vIK8qDZqy&sig=ACfU3U3OlUXTq5b7VBrCmNuAodleJu1vzQ&hl=de&sa=X&ved=2ahUKEwiptZazpufnAhWLepoKHX1TAloQ6AEwA3oECAgQAQ#v=onepage&q=Latex%20%5Cne&f=false

There sometimes seems to  be a problem with \neq in .dvi, which we will not use:
https://tex.stackexchange.com/questions/16456/latex-vs-pdflatex-on-neq

\not= seems to the basic command.

andreas-junghanns:
Created PR to turn the last two \ne into \neq.

chrbertsch:
Shall we leave the &#8800; or shall we replace it with  latexmath:[\neq] ?

andreas-junghanns:
> 
> 
> Shall we leave the ≠ or shall we replace it with latexmath:[\neq] ?

I messed this up: I have a commit 572337f in my clone andreas.junghanns but the PR somehow got lost...

andreas-junghanns:
Maybe PR #791 is the one containing the removal of the ASCII characters?

chrbertsch:
> Maybe PR #791 is the one containing the removal of the ASCII characters?

ok, thanks.

Issue #732, opened 2020-02-04
logMessage #vref# behaviour with aliases


nickbattle:
Section 2.1.5 describes the expansion of `#<valueReference>#` values into the name of the referenced variable in log messages. But if the value reference has an alias-set of variables, which variable's name is substituted? Is one is chosen arbitrarily or is it maybe an error to attempt this?

(I think this is also a question with FMI2 - the fmi-sdk seems to pick the first SV it finds with the right vref)

pmai:
As you say, the importing implementation has no way of knowing which of multiple variable names to choose here, as discussed in #435, esp. my comment https://github.com/modelica/fmi-standard/issues/435#issuecomment-433656106. So it will pick an arbitrary name.

It was decided that the ability for FMU exporters to not have to place knowledge of variable names into exported code was deemed more important than removing this complex mechanism with its draw backs. As such I think the rationale for this decision and the consequences should probably be pointed out in non-normative text, A pull request would probably be appreciated.

chrbertsch:
Regular design webmeeting:
A discussion has started if "alias variables" should refer to a variable that is no not an alias. This could then also resolve this issue. @klausschuch plans to create a proposals.

PR #781, opened 2020-02-22
Add explanation for logMessage in case of valueReference pointing to …

andreas-junghanns:
…alias set of variables (#732)


nickbattle:
It looks like this has already been merged anyway, but this change does clarify what happens in the case of an alias-set to an extent, so I approve.

Saying that the environment can choose "any or all" of the variable names to substitute leaves open the question of the format for multiple names (comma separated, space separated or some other format?), and it is not clear whether "any" means any one name or any subset of them. The order is presumably undefined too. But perhaps none of this is critical.

A very minor point, but I think it would be more natural to start the sentence with "If `#<valueReference>#` refers to an alias set of..."?

andreas-junghanns:
We should leave printing details to the simulation environment. I tried to keep it as open as possible to hand this freedom specifically to the importer that can then format the message however it sees fit (maybe even asking the user with options or whatever). It is clear to me, that it is not the FMI standard that can specify how to solve the conflict which name of the alias set to use.

PR #785, opened 2020-02-22
Fix tiny issue in non-normative text

andreas-junghanns:


andreas-junghanns:
Torsten: this is a simple change, please merge.

Issue #672, opened 2019-11-10
Fix hard coded xrefs


t-sommer:
e.g. `(see 2.2.3)` in [docs/2_1_common_api.adoc](https://github.com/modelica/fmi-standard/blob/master/docs/2_1_common_api.adoc) (line 611).

PR #788, opened 2020-02-22
Fix hard coded refs (#672)

andreas-junghanns:


chrbertsch:
In 2_1_common_api.adoc we have "see remark 3".  Shall we add an anchors to this remark and link to it?

andreas-junghanns:
> 
> 
> In 2_1_common_api.adoc we have "see remark 3". Shall we add an anchors to this remark and link to it?

I could not find this reference. Was that fixed in the meantime?

chrbertsch:
> I could not find this reference. Was that fixed in the meantime?

Ah, it seems to be in 3.1.1.
grafik

andreas-junghanns:
The remarks are all manually labeled in the text, so referencing the remark with a fixed reference seems safe, right?

Issue #669, opened 2019-11-10
Update type names in documentation, schema and examples


t-sommer:
The old FMI 2.0 type names (e.g. "Real" and "Integer") still appear in the documentation, XSDs and code examples. They should be update to match the new definitions in `fmi3PlatformTypes.h`.

I suggest to use `fmi3Float*` and `fmi3Int*` instead of `fmi3Real` and `fmi3Integer` to reflect the added 32-bit and signed / unsigned integer types.

chrbertsch:
We shall not close this ticket when merging PR 799 as there are still open points, e.g.
-    table entries: there is a lot more that needs fixing in the tables and it is not enough to set {VariableType} here as entries are different for floating point and integer types 

![grafik](https://user-images.githubusercontent.com/16363372/75172698-f3951500-572d-11ea-8d86-664b2bf0177c.png)

-  pseudo code in SES: This is for AndreasP to fix and should not go into this PR

andreas-junghanns:
These are different issues with a different focus, I think.
The tables need a general makeover.
The pseudo-code needs more than a quick fix and other issues need to be fixed there too.
Therefore: In the general text, all is done.
Any other issue goes to other tickets, if not already open (table fixes should already be listed).
Therefore, I strongly suggest to close this so that we do not get lost in endless ticket descriptions (rather open a new ticket with a reference to here but a full description of the remaining issues there)

chrbertsch:
Closing as #799 was merged

PR #799, opened 2020-02-23
Update type names in documentation, schema and examples (#669)

andreas-junghanns:


andreas-junghanns:
I tried to remove Real wherever needed (will send Integer PR right after).

Missing also fixes in the function-call tables.

chrbertsch:
there is several times the occurance of "A vector of real  ..." in 3.1.1 (line 166ff. )  and 4.1.1.
Shall this be changes also? To "float"

3.1.2 This mode is used to compute the values of all (real) continuous-time variables 

Tables in 3.2.3  and 4.2.1
grafik
grafik

Code Example in 3.2.4
grafik

4.1.1
Step Mode
This mode is used to compute the values of all (real) continuous-time and discrete-time variables at communication points by 

Code Example for fmi3CoSimulation in 4.2.1
grafik

Same for 4.2.4

And pseudocode for SES; several times, e.g.: 
grafik

andreas-junghanns:
I fixed most of Christian´s suggestions. Missing:
 - table entries: there is a lot more that needs fixing in the tables and it is not enough to set {VariableType} here as entries are different for floating point and integer types and is outside the scope of this PR
 - pseudo code in SES: This is for AndreasP to fix and should not go into this PR

andreas-junghanns:
This is a big change set - can the reviewer please check this soon to avoid merge conflicts, please?

chrbertsch:
Do we have to define the newly introduced term "numeric type"?

andreas-junghanns:
> Do we have to define the newly introduced term "numeric type"?

Why define "numeric type"? Any type with a number is numeric. Standard comp sci.

chrbertsch:
> Why define "numeric type"? Any type with a number is numeric. Standard comp sci.
Just randomly taking a random google for "numeric type": 
http://www.informit.com/articles/article.aspx?p=31344&seqNum=3
--> does not mention Boolean (is this a "numeric" or "logical" type?)

However this is not very important.

andreas-junghanns:
This is a big fish, I hope that Torsten can check this tonight, else we merge it tomorrow morning.

chrbertsch:
Closing as #799 was merged

PR #801, opened 2020-02-23
Clarify stepsize and discourage usage of CS for real sampled data systems but recommend HCS or SES instead

### LINKED ISSUE #434: Clarification of stepSize for CoSim FMUs
**Reported by maplesoft on 1 Jun 2018 17:21 UTC**
The meaning of stepSize, if provided in the DefaultExperiment field, needs to be clarified.
On pg. 43 of the spec it says:
"for CoSimulation the _stepSize_ defines the **preferred** communicationStepSize"
but on pg. 96 it says:
"the **smallest** sample period as accessible by the outside of the FMU is defined by attribute _stepSize_ in element DefaultExperiment"

So should the importing tool interpret it as a minimum or just a preferred step size?  This is important especially when canHandleVariableCommunicationStepSize = false.

----

Migrated-From: https://trac.fmi-standard.org/ticket/434

chrbertsch:
attempt to fix #434

### DIFF
--- a/docs/4_1_co-simulation_math.adoc
+++ b/docs/4_1_co-simulation_math.adoc
@@ -84,7 +84,7 @@ The slave simulators are seen by the master simulator as purely sampled-data sys
 Such a sampled-data system can be:
 
 - A real sampled-data system (so a sampled discrete controller; the <<input,`inputs`>> and <<output,`outputs`>> can be of type Real, Integer, Boolean, String, or enumeration.
-Variables of this type are defined with <<variability>> = <<discrete>>; the smallest sample period as accessible by the outside of the FMU is defined by attribute `stepSize` in element `DefaultExperiment`).
+Variables of this type are defined with <<variability>> = <<discrete>>.
 
 - A hybrid ODE that is integrated between communication points (known as "sampled access to time continuous systems") where internal events may occur and be handled, but events are not visible from the outside of the FMU.
 It is assumed here that all <<input,`inputs`>> and all <<output,`outputs`>> of this hybrid ODE are Real signals (defined with <<variability>> = <<continuous>>),
--- a/docs/4_1_co-simulation_math.adoc
+++ b/docs/4_1_co-simulation_math.adoc
@@ -81,15 +81,12 @@ This section contains a formal mathematical model of a Co-Simulation FMU.
 The following fundamental assumptions are made:
 
 The slave simulators are seen by the master simulator as purely sampled-data systems.
-Such a sampled-data system can be:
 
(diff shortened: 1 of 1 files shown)


pmai:
Is the wording not a tad too strong? I.e. do we have reasons to suggest that the use of CS is problematic for all sampled data systems?

chrbertsch:
> Is the wording not a tad too strong? I.e. do we have reasons to suggest that the use of CS is problematic for all sampled data systems?

Should we have something like "... FMI for CS can be used for simple sampled data systems that can be modeled with certain assumptions like
- no events
- one timestep delay
- no defintion of various "tasks" ..."
for which the new features of HCS (events, clocks) or SES shall be used.

andreas-junghanns:
Even if such an exclusion list would be helpful, it is misleading. "No events" is not true, one can have events, just the communication is more difficult and maybe besides the standard. And a time step is not defined, really, as one can reduce the time step and remove the delay, if so needed, under certain assumptions.

I think it is better to explain what the new simulation types are good for. That would be simpler to explain, I think. Let´s make this a brief discussion tomorrow.

docs/4_1_co-simulation_math.adoc
```diff
@@ -81,15 +81,12 @@ This section contains a formal mathematical model of a Co-Simulation FMU.
...
-
-- A hybrid ODE that is integrated between communication points (known as "sampled access to time continuous systems") where internal events may occur and be handled, but events are not visible from the outside of the FMU.
+Such a sampled-data system consists typically of a hybrid ODE that is integrated between communication points (known as "sampled access to time continuous systems") where internal events may occur and be handled, but events are not visible from the outside of the FMU.
 It is assumed here that all <<input,`inputs`>> and all <<output,`outputs`>> of this hybrid ODE are Real signals (defined with <<variability>> = <<continuous>>),
 
-- A combination of the systems above.
+_[ In FMI 2.0, Co-Simulation FMUs were also used for real sampled-data systems (so a sampled discrete controller; the <<input,`inputs`>> and <<output,`outputs`>> could be of type Real, Integer, Boolean, String, or enumeration with <<variability>> = <<discrete>>._
+_However, in FMI 3.0, Hybrid Co-Simulation (HCS) and Scheduled Execution Simulation (SES) are intended for this use-case.]_
```
APillekeit: Frankly, I don't understand the point of this discussion. Would it be helpful if we change inline 89 "are intended" into "are more suitable".

APillekeit: We explain the use cases of the different interface types at another place. 
(see pull request https://github.com/modelica/fmi-standard/pull/728)
I would not repeat that here. The issue #434 is solved in my view with the updates Christian provided.

Issue #804, opened 2020-02-25
Define "numeric type" 


chrbertsch:
at its first occurance according to the discussion in https://github.com/modelica/fmi-standard/pull/799, especially including Boolean.

I will do this after the merge of https://github.com/modelica/fmi-standard/pull/799

chrbertsch:
fixed with PR #816

PR #816, opened 2020-02-26
define nuemeric data type

chrbertsch:
according to https://github.com/modelica/fmi-standard/issues/804


andreas-junghanns:
I removed more sentences that are not true anymore for 3.0, please check

Issue #741, opened 2020-02-10
fmi3GetNumberOfEventIndicators/fmi3GetNumberOfContinuousStates


friedrichatgc:
The number of event indicators and the number states can change dependent on the "structural parameters".

For the number of event indicators this dependency must be defined in modelDescription.xml using the "numberOfEventIndicators" attribute in "ModelStructure". An C-API function is provided to get the number at runtime. However, this dependency definition is fully unclear to me and seem not to be described in the standard document.

For the number of continuous states no such dependency is provided in modelDescription.xml. Only a C-API function is provided to get the number. This look inconsistent to me.

Please clarify and unify the number of event and states handling in the document.

t-sommer:
Alternatively we could use the `numberOfEventIndicators` as a fixed (maximum) number of event indicators and provide an additional array of type `fmi3Boolean` that indicates which event indicator is currently active.

friedrichatgc:
What is the number of states/eventindicators if state/eventindicators variables are of type array?
I think each element in the array count as one state/eventindicator but I can't find a definition of this in the document.

t-sommer:
@friedrichatgc, would it be sufficient to define that the event indicators / states vectors must be "compressed" (i.e. all active elements at the beginning) and attribute "numberOf..." defines the max. length of these vectors?

friedrichatgc:
I don't think that we can define a maximal number of states/event indicators since this is only possible if all Dimension's of array variables being states/event indicators have a max attribute.

I think we need to remove numberOfEventIndicators from the xml file as it was already done for numberOfContinousStates. The importing tool must than calculate this value based all other information in the xml file or even at runtime if structural parameters change.

t-sommer:
> The importing tool must than calculate this value based all other information in the xml file

How would that work?

chrbertsch:
Design Meeting: 
Klaus: for the getNumberOfContinues states : you can derive this from the modelstructure and their dimensions
Torsten: one can count the number of states 
Klaus: I do not think we have to change anything
Markus: So this is only needed for event indicators because they do not have a corresponding variable. 
Klaus: There is a request from the QSS people also to have a variable for the event indicators (#587)
Klaus: Klaus #587 would also be nice for debugging. Would be new section in the modelstructure. This would be a new causality.

Decision: check for non-normative text. @friedrichatgc : Will do it

friedrichatgc:
[...]

PR #829, opened 2020-02-27
Clarify the number of states/event indicators (#741)

friedrichatgc:
fixing #741 
Getting the number of states and event indicators cannot be done
consistently since every state is also present as a variable but event
indicators are not.
Now a clear defintion is given how the static number of states/events
can be get without using C-API function calls and also how to get the
dependent structural parameter the number of states/events depends on.

### DIFF
--- a/docs/1___overview.adoc
+++ b/docs/1___overview.adoc
@@ -228,8 +228,6 @@ Properties that might be additionally needed for trajectory optimization, for ex
 
 - No explicit definition of the variable hierarchy in the XML file.
 
-- The number of states and number of event indicators are fixed for an FMU and cannot be changed.
-
 === Acknowledgements
 
 Until Dec. 2011, this work was carried out within the ITEA2 MODELISAR project (project number: ITEA2-07006, https://itea3.org/project/modelisar.html).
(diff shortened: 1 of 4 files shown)


masoud-najafi:
But we decided to have event indicators as local variables and provide the list of these variables in  a new EventIndicators child in the ModelStructure to provide the list of event indicator variables. For example if we have three event indicator variables with references 31, 32, 34, then we may have this structure.
<ModelStructure>
  <EventIndicators>
    <Unknown reference="31" dependencies "1 3 4" />
    <Unknown reference="32" dependencies ""/>
    <Unknown reference="32"/>
  </EventIndicators>
</ModelStructure>

This element should be optional, if there is no event indicators, similar to <Derivatives> and <Outputs> and <InitialUnknowns>.
In this way, there is no need to have the attribute numberOfEventIndicators.

chrbertsch:
> But we decided to have event indicators as local variables

@masoud-najafi : as stated in #541:
> A clarification is added with PR #829. Issue #587 is still open and will be addressed with another PR.

So this PR is not intended to replace #587 but to fix another problem.

friedrichatgc:
Yes, this PR has nothing to do with issue #587. Its a bugfix/clarification PR for issue #741, as noted in the title.
The feature issue #587 needs still to be done (and will change again some of the changes introduced with this PR)

friedrichatgc:
What is with this PR now? Any objections? Since the needed changes for #587 will be partially in conflict with this changes which fixes #741 I will start with #587 after this PR is merged.

andreas-junghanns:
@friedrichatgc : will you merge?

friedrichatgc:
I can merge, but @masoud-najafi is still a pending reviewer.

masoud-najafi:
Sorry for delay. I had an issue with my mailbox...
@friedrichatgc : In the doc we have;   _[Therefore their is no `numberOfStates` attribute as it is for `numberOfEventIndicators`.]_
But I do not understand why do we need numberOfEventIndicators attribute in the XML. If event indicators depend on structural parameters, the number of eventIndicators changes and they should be recomputed dynamically by the importer. May be I am missing something here.

Issue #424, opened 2018-10-17
Missing dependencies on tunable parameters


modelica-trac-importer:
**Reported by manuel&#46;graeber on 10 Oct 2017 15:00 UTC**
For optimization applications I need the structural information, if outputs of a model exchange FMU directly depend on tunable parameters.

Use Case: FMU outputs in cost functions or constraints of an optimization problem. Derivative based algorithms need to know, if there are direct dependencies on optimization variables (tunable parameters).

Reading the FMI Spec, there seems to be something missing. On page 58 tunable parameters are listed, just as I would expect it:

(𝐲𝑐, 𝐲𝑑) ≔ 𝐟𝑜𝑢𝑡𝑝𝑢𝑡(𝐱𝑐, 𝒖𝑐, 𝒖𝑑, 𝑡, 𝒑𝑡𝑢𝑛𝑒)

But on page 61 in the description of the attribute "dependencies", tunable parameters are not listed anymore.

Is there a reason for it?

----

Migrated-From: https://trac.fmi-standard.org/ticket/424

modelica-trac-importer:
**Modified by cbertsch on 1 Jun 2018 15:17 UTC**

modelica-trac-importer:
**Modified by cbertsch on 18 Jun 2018 10:21 UTC**

chrbertsch:
Regular design meeting:

We should add this to output dependencies.
"Tunable parameters are essentially discrete inputs".

chrbertsch:
Regular Design Meeting: 

Adding partial derivatives w.r.t. to parameters would introduce a larger change for implementing partial derivatives by tools. 
We can move to this to FMI 3.1

A clarification is needed that by intent that parameters and tunable parameters are not part of the knowns in step mode. 
@friedrichatgc : I could create a PR for this.

clagms:
To add a bit more detail to this issue,
Suppose I have the following model.

-------------------

Parameters: $p_0$

Tunable Parameters: $p_1$

Calculated Parameters: $p_2$

Inputs: $u_3$

Outputs: $y_4$

All vars are reals (Float64): $p_0, p_1, p_2, u_3, y_4 \in \mathbb{R}$ 

Assignments active on initialization mode:

$p_2 := p_0 + p_1$

$y_4 := (p_1 + p_2)*u_3$ 

Assignments active on continuous time mode (when doStep executes):

$y_4 := (p_1 + p_2)*u_3$ 

[...]

PR #830, opened 2020-02-28
add dependencies on tunable parameters

chrbertsch:
fixing #424

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -2041,6 +2041,8 @@ Otherwise the <<Unknown>> depends on the `Knowns` defined by the given value ref
 
 - continuous-time states
 
+- tunable parameters
+
 - <<independent>> variable (usually time; <<causality>> = <<independent>>).
 If an <<independent>> variable is not explicitly defined under variables, it is assumed that the <<Unknown>> depends explicitly on the <<independent>> variable.
 


friedrichatgc:
I'm not sure if tunable parameters should be added here:
Why adding tunable parameters but not (normal) parameters?
The dependencies are used to get (directional) derivatives of the unknowns wrt the knows. For FMI 3.0 it was decided that (as for FMI 2.0) this does NOT include derivatives wrt parameters.
The topic of having derivatives wrt parameters was added to the DAE working group which will not go into 3.0.
So I think issue #424 should be moved to > FMI 3.0.

chrbertsch:
> Why adding tunable parameters but not (normal) parameters?

This could be discussed. 

> The dependencies are used to get (directional) derivatives of the unknowns wrt the knows. For FMI 3.0 it was decided that (as for FMI 2.0) this does NOT include derivatives wrt parameters.

Ah, I was not aware of this. What have been the reasons? When was this discussed and to we have notes on that discussion?

> The topic of having derivatives wrt parameters was added to the DAE working group which will not go into 3.0.

What is the Link of this topic to DAEs?

friedrichatgc:
The only reason for adding tunable parameters but not "normal" parameters I see is that for tunable parameters derivatives may be "emulated" by the environment using finite differences. For "normal" parameters this is not possible. But why requiring a parameter to be tunable to get (analytical) derivatives wrt the parameter. Hence, I propose to add all parameters.

> When was this discussed and to we have notes on that discussion?

See https://github.com/modelica/fmi-design/blob/master/WorkingGroups/02_DAE_PartialDerivatives/Meetings/2016-06-28-FMI-Meeting-Renningen/DerWrtPar.pdf for a proposal. This proposal basically states that adding parameters to the list of knowns is enought (as done with this PR). But some concerns regarding models with a very large number of parameters are given in the last slides. (but I don't remember details)

> What is the Link of this topic to DAEs?

None, I was also surprised about the decision to move this to the DAE working group.

In general I think we may need a capability flag whether the FMU can calculate derivatives wrt parameters or not since many tools can provide derivatives wrt inputs/states but not wrt parameters as needed for the more special case of optimization.
Moreover some text is needed for the use case of derivatives wrt parameters.

andreas-junghanns:
I am unconvinced. What is the use case for derivatives on parameters?
In ECUs, parameters are flashed. One change.
Tuning is a manual process, sometimes semi-automatic.
What restrictions do you want to model with derivatives?

chrbertsch:
> What is the use case for derivatives on parameters?

see slide 4 of https://github.com/modelica/fmi-design/blob/master/WorkingGroups/02_DAE_PartialDerivatives/Meetings/2016-06-28-FMI-Meeting-Renningen/DerWrtPar.pdf

andreas-junghanns:
> 
> 
> > What is the use case for derivatives on parameters?
> 
> see slide 4 of https://github.com/modelica/fmi-design/blob/master/WorkingGroups/02_DAE_PartialDerivatives/Meetings/2016-06-28-FMI-Meeting-Renningen/DerWrtPar.pdf

Ok, makes sense - this is quite useful to have.

chrbertsch:
Regular Design meeting: 

Markus: this introduces a large change, one has to check a lot of text
Claudio: One has to clearly defined what the derivatives w.r.t parameters mean
Markus: instant sensitivitiy in the sense of a Jacobian
Claudio: do tunable parameters change the outputs instantaneously?
Klaus: in ME they can have an imediate effect, in CS not
Markus: in CS you cannot emulate the partial derivatives numerically with deltas
Klaus: Do we not have dependencies on parameters in initial unknowns?
Markus: at most places talking about partial derivatives, we implicitely mean states and inputs. this is also how tools use it.
Is it allowed to get the partial derivatives in initialization mode?
Klaus: We might have a problem also for initial unknowns.
Markus: this is a smaller problem
Andreas:  Can we move this topic in 3.1?
Markus: I think we can, see slides of DAE proposal

Klaus: in 2.0 it is allowed to list parameters in initial unknowns. At least in ME one can call partial derivatives. 
Claudio: it is not easy to allow partial derivatives in step mode.
Claudio: in FMI 3.0, variables that can be changed in initial model, can be listed.

Decision: This is a too large change for FMI 3.0.

Issue #186, opened 2018-10-16
Clarification of Argument "loggingOn" of fmiInstantiate needed


modelica-trac-importer:
**Reported by torstenblochwitz on 19 Mar 2014 15:38 UTC**
It is not clear for me, how argument "loggingOn" interacts with the fmiSetDebugLogging function and the fmiModelDescription.LogCategories of modelDescription.

----

Migrated-From: https://trac.fmi-standard.org/ticket/186

modelica-trac-importer:
**Comment by torstenblochwitz on 27 Mar 2014 10:35 UTC**
Added non-normative text:
A FMU could set all logCategories according to this argument. Which logCategory the FMU sets is unspecified.]

modelica-trac-importer:
**Comment by rfranke on 6 Aug 2014 11:57 UTC**
We are experiencing problems with the logging behaviour of FMUs exported by different tools or by different versions of the same tool.

 - In one case the behavior changed with a new version of an exporting tool. An earlier version of the tool did only report warnings and errors with fmiInstantiate(..., loggingOn=true). A new version of the same tool reports all messages for the same fmiInstantiate call. The vast amount of Set/GetReal messages broke our importing application. Even if the logging function did do nothing, already the calls slowed down the whole application significantly.

 - In another case the behavior changes with a different tool: loggingOn=false means for this tool that no logs appear at all, not even in the case of critical errors.

The statement in section 2.1.5 "Which LogCategories the FMU sets is unspecified." is not satisfactory. Couldn't the standard define the expected behavior for loggingOn=fmi2True and loggingOn=fmi2False?

Alternatively, the function `fmi2SetDebugLogging` could be used. Its specification says:

"If loggingOn=fmi2True and nCategories > 0, then only debug messages according to
the categories argument shall be printed via the logger function."

This sounds good. But how to tell an FMU that it should do no logging at all?

modelica-trac-importer:
**Comment by rfranke on 6 Aug 2014 19:24 UTC**
Here is a solution proposal that clearly defines debug logging and that enables turning it off completely. The following paragraph in section 2.1.5:

_If loggingOn=fmi2True, debug logging is enabled. If loggingOn=fmi2False, debug logging is disabled. [The FMU enable/disables LogCategories which are useful for debugging according to this argument. Which LogCategories the FMU sets is unspecified.]_

is rephrased to:

[...]

PR #831, opened 2020-02-28
Clarify logging according (#186)

chrbertsch:
fixing #186


pmai:
Based on my reading of the discussion in #186, I think the definition of `fmi3SetDebugLogging` should also be adjusted to clarify that its loggingOn argument works in the same way, i.e. loggingOn = fmi3False disables *all* logging.

While we are at it, we might want to rename the function to `fmi3SetLogging`.

Issue #740, opened 2020-02-10
Use UInt64 for attribute "start" in element "Dimension" 


t-sommer:
to allow large arrays on 64-bit machines where `size_t` is `unsigned long long` (see also #736).

andreas-junghanns:
Go!

pmai:
This should then also change all other definitions of array dimension variables, to provide a consistent set of definitions, right? I.e. all array dimensions will then be 64bit uint, on both 32 and 64 bit platforms.

PR #832, opened 2020-02-29
Use UInt64 for dimension sizes

t-sommer:
see #740

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1154,6 +1154,10 @@ All initial dimension sizes (i.e. prior to any configuration or reconfiguration)
 Changes to dimension sizes are constrained by the `min`/`max` attributes of the referenced <<structuralParameter,`structural parameters`>>, which can be any non-negative integer, including zero.
 Specifying a minimum size of zero on a <<structuralParameter,`structural parameter`>>allows any related dimension sizes to be changed to zero in *Configuration Mode* or *Reconfiguration Mode*, thus causing the respective array size to go to zero, which leaves the respective array variable without any active elements.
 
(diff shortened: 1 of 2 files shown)


andreas-junghanns:
We use `size_t` everywhere else in the document when we describe sizes. Why do we force 64-bit here?
While it is likely ok to have a unsignedLong in the XSD for the size, to ask the valueReference to point to a 64-bit unsigned int requires all platforms to have a 64-bit int available. Is that ok? Does a way exist to have a "size_t" in the XML with flexible bit size?

pmai:
XML does not provide this (unsurprisingly, since the format is supposed to be machine independent), and neither to we provide a variable type that is platform-dependent (anymore, if ever ;). We could of course add verbiage that restricts start values to 32 bit on 32 bit platforms and 64bit on 64bit platforms, etc.

The bigger problem we have is that we'd also need a platform-specific variable type, SizeT, or whatever, to deal with the variable-derived dimension sizes, which then can't be easily used with integer variables, etc.

In this sense we are kind of stuck between a rock and a hard place, since the API does restrict most stuff to size_t, so on 32bit platforms you will not be able to access array elements larger than 2^32-1, even if the XML description and dimension variables allow you to provide sizes larger than that.

I think the only halfway sane solution to this is to let the dimensions be up to 64bit, and use 64bit values and variables for this (changing and reading dimension sizes is surely not performance critical), but point out that on non-64bit platforms you will be unlikely to use such large dimensions anyway...

So probably the PR needs to put this information somewhere as well...

pmai:
I have added some proposed wording to this effect, please review @t-sommer and @andreas-junghanns if this clarifies the situation enough, or amend as suitable.

Issue #814, opened 2020-02-25
Inconsistency in fmuResourceLocation and resources folder?


masoud-najafi:
The argument fmuResourceLocation in fmi3Instantiate is an URI, so it cannot be optional and the importer should provide a valid URI to the location of the resources directory of the unzipped FMU.
But, the "resources" directory is optional. So to what the URI should point to if the "resources" directory does not exist? Pointing to a non-existing folder may be dangerous.

chrbertsch:
I propose to make the resource folder mandatory.

chrbertsch:
I propose to make the resource folder mandatory.

andreas-junghanns:
PR #833 merged.

chrbertsch:
So there seems to be broad support to revert #833  and
- make the resource folder optional
- clearly describe to return a  NULL pointer, if the environment cannot provide the resource location 
-- either because there is no resource folder in the FMU
-- or because due to other reasons it cannot provide an URI to that location (e.g., because there is no filesystem)

(The importer can check in advance, if there is a resource location and provide a feedback to the user if it cannot handle it, and the FMU can raise an error if it gets returned a NULL but depends on information in the resources folder)

Any objections to following this path and creating a PR?

andreas-junghanns:
No objections - will you create the PR?

andreas-junghanns:
Should we continue the discussion in the open ticket #1020?

PR #833, opened 2020-02-29
making resource directory mandatory

chrbertsch:
fixing #814

### DIFF
--- a/docs/2_3_fmu_distribution.adoc
+++ b/docs/2_3_fmu_distribution.adoc
@@ -47,7 +47,7 @@ binaries                      // directory containing the binaries (optional)
    // folder are recommended to be unpacked at the same location as the binary
    // < modelIdentifier >.* is unpacked. If not it is likely that, if the FMU
    // has dependencies on those items, it will not be able to find them.
-resources                     // resources needed by the FMU (optional)
+resources                     // resources needed by the FMU
    // data in FMU specific files which will be read during initialization;
    // also more folders can be added under resources (tool/model specific).
    // In order for the FMU to access these resource files, the resource directory
@@ -222,7 +222,7 @@ _[The sub-directory "licenses" can be used to bundle all license files._
 _It is strongly recommended to include all license and copyright related information in the licenses folder of an FMU (especially but not only for contained open source software) - the `license.{txt,html}` file can serve as an entry point for describing the contained licenses._
 _This will help the users to comply with license conditions when passing source or binary code contained in an FMU to other persons or organizations.]_
 
-In directory `resources`, additional data can be provided in FMU specific formats, typically for tables and maps used in the FMU.
+In the mandatory directory `resources`, additional data can be provided in FMU specific formats, typically for tables and maps used in the FMU.
 This data must be read into the model at the latest during initialization (that is, before <<fmi3ExitInitializationMode>> is called).
 The actual file names in the ZIP file to access the data files can either be hard-coded in the generated FMU functions, or the file names can be provided as string parameters via the `fmi3SetString` function.
 _[Note that the absolute file name of the resource directory is provided by the initialization functions.]_


chrbertsch:
> I wish we had discussed this during the Design Meeting. But I see no other solution.

I added some more reviewers, but also everyone else is welcome to comment or review.

pmai:
I can see three distinct solutions to this problem:

1. Make resources mandatory as in this PR
2. Let the fmuResourceLocation argument be NULL (or some other special value), if no resources folder is present in the FMU
3. Do away with the special treatment of resources alltogether, and replace fmuResourceLocation with fmuLocation pointing to the root of the unpacked FMU

Choice 2 seems slightly more light weight, yet potentially more error-prone (though FMUs have to guard against this case for historical reasons anyway). It could also allow support for platforms that cannot provide access to the resources folder, allowing the FMU to decide whether it can operate without it or not - if we want to do that.

Choice 3 would address the current situation while also allowing potentially more flexibility in the future, and would ensure that additional DLLs/SOs or other resources that need to be co-located with the main DLL/SO would be properly unpacked, and can be accessed via fmuLocation and not via platform-specific ways of locating SOs/DLLs based on the base DLL/SO location.

But Choice 1 also seems defensible, if somewhat kludgey (that said, importers still have to lie for cases that the FMU illegally does not provide the resources folder, so it does not make the situation that much clearer).

My 2 centi-euros, if you will.

andreas-junghanns:
Choice 3 requires unpacking the entire FMU always and that was what we specifically tried to avoid in the beginning (trying to support light-weight platforms). Especially when we have FMUs that support multiple interface types and multiple platforms, this seems like a huge waste.

Choice 2 is what we have to do now if there is no resource directory (because it is optional). If we do not like mandating the existence of a potentially empty resource directory, we should clarify that NULL is acceptable (albeit any pointer in C can legally have NULL so to make this explicit is just to give us more reasons to yell at bad implementations).

I am leaning towards Choice 1 simply because this makes things more clear in the API, even if it shifts the burden to the importing tool to make sure it exists when the FMU does not provide the directory (or issue an indignation error).

pmai:
@andreas-junghanns I agree that trying to avoid unpacking the whole FMU is the theoretical foundation of only making resources accessible, but I think in practical terms it does not really work that way. I.e. I could not whole-heartedly recommend an implementation to try to only unpack resources, because quite a number of FMUs will fail to work when this is done, sometimes for not-so-bad reasons:

Currently, if an FMU needs to load dependent DLLs/SOs from its main DLL/SO, it has basically two ways of trying to make that work:

1) Locate the DLLs/SOs inside its binaries folder, hoping that the whole binary directory is unpacked intact and that it can find that location through platform-specific functionality,
2) Locate the DLLs/SOs inside its resources folder (probably in per-platform subdirs), thus negating the advantage of keeping the to-be-unpacked stuff small (also does not work well when the DLLs/SOs would be needed prior to instantiate call, but that is mostly only a theoretical concern).

The 3rd way of relying on the OS to do relative loading at load time does not work very well especially on Windows, unless the importing implementation goes through some hoops to make that work.

Choice 3 would alleviate this to some degree, and seems to match current practice anyway.

Note that I'm not propagating much for Choice 3, just pointing out some of the benefits.

I'm mostly neutral, leaning slightly to prefer choice 2 (since that seems minimaly invasive and fairly clear, and FMUs already have to deal with this case), or choice 3 (since that seems to solve some other issues), but choice 1 seems also OK with me (though it forces implementations to invent some URI when the resources folder is missing).

andreas-junghanns:
We currently require that:
"The binaries must be placed in the respective <platformTuple> directory with the general format `<arch>-<sys>{-<abi>{<abi_ver>}{<abi_sub>}}`."

I did not find the explicit requirement to unpack all binaries of the current platform. This should be added, agreed. But to move binaries into the resource folder seems like a drastic step that would need good reasoning to support.
So we are stuck with the relative, platform depenend loading trick we all are using.
But since loading DLLs is platform specific anyway, should we really worry and change things now?

andreas-junghanns:
@pmai commented, but did not mark as "approved" but seemed to not disapprove.
@masoud-najafi requested the change, so this is implicitly approved by him.
Missing, @t-sommer.

I think this is good enough, I will merge.

andreas-junghanns:
> Choice 2 is what we have to do now if there is no resource directory (because it is optional). If we do not like mandating the existence of a potentially empty resource directory, we should clarify that NULL is acceptable (albeit any pointer in C can legally have NULL so to make this explicit is just to give us more reasons to yell at bad implementations).
> 
> I am leaning towards Choice 1 simply because this makes things more clear in the API, even if it shifts the burden to the importing tool to make sure it exists when the FMU does not provide the directory (or issue an indignation error).

I change my mind here: There are target platforms without access to file systems and it must be ok to start an FMU without resourceLocation (== NULL).

Issue #807, opened 2020-02-25
Replace alias variables with <Alias> elements


t-sommer:
As discussed at the Berlin 2020 Design Meeting (meeting minutes at https://github.com/modelica/fmi-design/blob/master/Meetings/2020-02-25-FMI-Design-Berlin/minutes.adoc)

Proposal:

The aliases are only information attached to a variable (no actual variable). It's up to the importer howto handle (display, log, etc.) them.

```xml
<Float64 name="torque.tau_0" valueReference="2" description="Accelerating torque acting at flange"
      declaredType="Modelica.SIunits.Torque">
   <Alias name="torque.tau_1" description="Alias 1" displayUnit="kN.m"/>
   <Alias name="torque.tau_2" description="Alias 2" displayUnit="nN.m"/>
</Float64>
```

Only the following attributes are allowed for `<Alias>` elements:

- `name`
- `description`
- `display unit`

All attributes follow the same conventions as for variables.

chrbertsch:
Create also non-normative text about what the exporting tools have to do (e.g. find the common range of min, max, ....)

CSchulzeTLK:
* There is one variable, and aliases (additional names with varied displayUnit and description) which belong to them
* The namespace for ModelVariables and aliases is merged. So there must not be an alias which has the same name as a variable. 
* just the variable can be set with `set{VariableType}`, because its the only one with a valueReference
* Alias of time variable? No restriction necessary.
* if causality is
  * parameter
    * the aliases are treated similar to a calculatedParameter
  * input/output
    * also possible, no reason to restrict
* clock variables don't have aliases? => crosscheck the clocks section
* all attributes are equal, except:
  * name
  * displayUnit
  * description
* i.e. all are treated as if they e.g. had the same `<Unit name>` and the same `<Unit><BaseUnit>` definitions.
* exporting tool has to merge the other attributes
  * min/max/nominal/quantity/unit/variability...
  * no longer separate min max values (orignially allowed to provide submodel information in systems with connected submodels)
  * no longer contraticting start values for constants
* it is obvious which parameter name should be displayed in the importing tool

=> replace the word "aliasing" in the rest of the specification

chrbertsch:
Design Meeting: 

if causality is
    parameter
        the aliases are treated similar to a calculatedParameter

--> this is true but shall not be explicitly written.

clocks could have aliases with the new concept

If an alias allowed to have a different quantity or only a different unit (e.g. "m" and "s") --> No
[...]

PR #834, opened 2020-03-01
Add <Alias> element to model variables

t-sommer:
according to #807

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -642,7 +642,7 @@ _where `tolerance` is, for example, the relative tolerance defined in `<DefaultE
 |`unbounded`
 |If `true`, indicates that during time integration, the variable gets a value much larger than its nominal value `nominal`.
 _[Typical examples are the monotonically increasing rotation angles of crank shafts and the longitudinal position of a vehicle along the track in long distance simulations._
-_This information can, for example, be used to increase numerical stability and accuracy by setting the corresponding bound for the relative error to zero (relative tolerance = 0.0), if the corresponding variable or an alias of it is a continuous <<state>> variable.]_
+_This information can, for example, be used to increase numerical stability and accuracy by setting the corresponding bound for the relative error to zero (relative tolerance = 0.0), if the corresponding variable or an is a continuous <<state>> variable.]_
 
 |Item
 |Items of an enumeration has a sequence of `name` and `value` pairs.
@@ -906,7 +906,7 @@ The normalized string `variableName` is used to identify the terminal member var
 _[The `name` attribute of the model variable is equal to the `variableName` of the `TerminalMemberVariable`._
 _An importing tool may use the `variableName` attribute to identify a specific alias to get additional information, such as minimum, maximum, and nominal values.]_
 
-One variable can be part of several terminals, which is considered aliasing according to the rules of <<aliasing-of-variables>>.
+One variable can be part of several terminals.
 
 _[Tools which apply sequence based matching should also provide unique member names, so that all importing tools which use name based matching use the same member names._
 
@@ -1177,7 +1177,7 @@ Every variable is uniquely identified within an FMU instance by this name.
 [[valueReference,`valueReference`]]
 A handle of the variable to efficiently identify the variable value in the model interface and for references within the `modelDescription.xml`.
 This handle is a secret of the tool that generated the C functions.
-It is required to be unique for an FMU with exception of variables that have identical values (such variables are also called alias variables). `Clock` type variables do not have alias variables and are therefore unique.
+It is required to be unique for an FMU.
 This attribute is `required`.
 
 |`description`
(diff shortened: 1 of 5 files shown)


andreas-junghanns:
Why can´t have numeric variables not have a unit?

t-sommer:
> Why can´t have numeric variables not have a unit?

Because currently only Float32 and Float64 can have units.

pmai:
I could not attend the design meeting, so maybe this was discussed, but is there a good reason why clock variables cannot have aliases, i.e. have to be treated specially?

Because if not, the ugly repetition of the aliases stuff could be moved to fmi3VariableBase, where it seems to me to belong as much as the dimensions, etc.

(Regardless of whether I like the new refactoring itself or not).

t-sommer:
> Because if not, the ugly repetition of the aliases stuff could be moved to fmi3VariableBase, where it seems to me to belong as much as the dimensions, etc.

I don't think this is possible because you would still have to distinguish non-scalar variables (no `negated` attribute) and floating point variables (that have a `unit` attribute).

However, if you have a better solution, please feel free to push your changes.

docs/2_2_common_schema.adoc
```diff
@@ -906,7 +906,7 @@ The normalized string `variableName` is used to identify the terminal member var
 _[The `name` attribute of the model variable is equal to the `variableName` of the `TerminalMemberVariable`._
 _An importing tool may use the `variableName` attribute to identify a specific alias to get additional information, such as minimum, maximum, and nominal values.]_
 
-One variable can be part of several terminals, which is considered aliasing according to the rules of <<aliasing-of-variables>>.
+One variable can be part of several terminals.
```
friedrichatgc: Is this really true for all variables? Can a input variable be part of several terminals?

docs/2_2_common_schema.adoc
```diff
@@ -1159,6 +1159,19 @@ It should be noted that changing a <<structuralParameter,`structural parameter`>
 
 It is not possible to use a variable of type `Clock` in arrays, since array elements do not have individual <<valueReference>> and the access to <<clock,`clocks`>> requires a unique <<valueReference>> information.
 
+A variable can have any number of `<Alias>` elements that define a variable alias.
+Each variable alias has a required attribute `name` whose value must be unique among all variables and variable aliases and an optional attribute `description`.
+Variable aliases of numeric variables can optionally be defined with the attribute `negated = true` which indicates that the variable's value must be multiplied with `-1` to retrieve the value of the variable alias.
```
friedrichatgc: Do we really want to reintroduce negated alias variables?

chrbertsch: I am against negated variables and the corresponding attiribute.
This complicates the work of the importer. 
Why not have a separate variable with the negative value?

t-sommer: The whole point of aliases (as I understand it) is to make tracing variables more efficient. We decided that aliases are just additional information and can be completely neglected by the importer.

chrbertsch: > The whole point of aliases (as I understand it) is to make tracing variables more efficient. We decided that aliases are just additional information and can be completely neglected by the importer.

But especially then I would get rid of the negated variables. 
(The reason to have them in FMI 2.0 to my understanding was to have e.g. two outputs referring to the same value reference, but with opposite sign, enabling the fulfillment of sign conventions, e.g. for the forces on both sides of a spring)

pmai: The re-introduction of negated aliases also threw me for a loop: They were _removed_ in 2.0, so I think we should have a good reason to reintroduce them in 3.0. Given that the aliases are now completely ignorable add-on information, if we allow negated stuff, why not other linear (or non-linear for that matter) transformations (alternative units already kind of give you this ability partially through misuse)?

This seems to complexify stuff needlessly: Without negated aliases, this would at least allow the non-floating-point variables to use the same elements (and the FP stuff already has more specialties), thus reducing schema complexity as well.

Given that 2.0 did not support negated aliases, and I haven't seen any major complaints, why re-introduce them?

clagms: Can we not mimic negated variables by just using display units?
For instance:
```
<UnitDefinitions>
  <Unit name="kg.m2">
    <BaseUnit kg="1" m="2"/>
    <DisplayUnit name="-kg.m2" factor="-1"/>
  </Unit>
</UnitDefinitions>
...
<Float32 name="J" valueReference="1073741824"
  description="Moment of load inertia" causality="output" variability="continuous"
  unit="kg.m2" start="1">
  <Alias name="-J" description="Negation example" displayUnit="-kg.m2"/>
</Float32>
```

t-sommer: I find the `negated` attribute more elegant than the "unit solution" because it also covers variables that have no unit.
However, if this this is the only thing that's holding you back from approving this PR, I can remove it. We could still add it later.

docs/2_2_common_schema.adoc
t-sommer: Please open a new issue if you think this is should be changed. This PR only removes the reference to alias variables.

docs/2_2_common_schema.adoc
nickbattle: How would a `negated` option work with min/max attributes for a variable?

PR #840, opened 2020-03-09
Add alias element to variable types.

clagms:
Defines Alias variables for the XML and provides an example.
If these changes make sense, I will upload the adocs with the description and constraints on the Alias element.
Please do not merge the PR until I've uploaded the adocs with the description. I just need some extra pair of eyes on the XSD, so that there are no big issues with the way the Alias element has been defined.
Refer to Alias.xml for examples on how the Alias element is to be used.

Remark: I had to change the definition of fmi3String. Before, the number of occurrences of the Start element was controlled by the parent xs:sequence element. This made it cumbersume to add the Alias element (as I would need another xs:sequence element, because we don't want alternating sequences of Start and Alias elements).
I believe the new definition is equivalent, and I've checked all xml files against the new schema.

### DIFF
--- a/.circleci/validate_xml.py
+++ b/.circleci/validate_xml.py
@@ -7,6 +7,7 @@ schema = etree.XMLSchema(file='schema/fmi3ModelDescription.xsd')
 parser = etree.XMLParser(schema=schema)
 
 xml_files = [
+    'alias.xml',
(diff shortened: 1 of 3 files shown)


clagms:
Closed because just noticed that there's a PR doing the same thing 😄

PR #848, opened 2020-03-11
Fix missing changes of PR #591 (changed ModelStructure)

friedrichatgc:
The ModelStructure child container elements Outputs, Derivatives and
InitialUnknowns were removed with PR #591 in the xsd file and replace by a list of
Output, Derivative and InitialUnknown elements.
This commit applies this needed changes to the document.


andreas-junghanns:
Markus: Please check those two tiny additional fixes that I propose.

I saw two changes to the \dot in math formulas: no idea if those are right, sorry.
But I will accept your change set anyhow to move things forward and hope others will check the \dot changes that know what they mean.

friedrichatgc:
@andreas-junghanns Thanks for the two last changes. Don't know why I have missed these, I have searched the document for Outputs and Derivatives.
I have double checked the \dot changes and I'm sure these are right.

PR #849, opened 2020-03-11
CI adaptions

friedrichatgc:
Create a docker image usable to build the examples.
Use this docker image for the example build in CI.
Added .sh/.bat scripts to build the doc and examples locally (with docker installed).
See CONTRIBUTING.adoc for details of the local build.

Required adaptions before a merge of this commit:
- Create a modelica or FMI organization on DockerHub (analog to the modelica organization on GitHub).
- Push the docker image build_examples to this DockerHub organization.
- Adapt the scripts to use this docker image (for now my DockerHub account "fm12" is used)
- Maybe automatically rebuild the docker image by CI and push it to DockerHub
  (The DockerHub login credentials needs to be configure in CI for this.)
  But the image should rarely be changed, so a pure manual update of the image is also possisble.

### DIFF
--- a/.circleci/build_examples.sh
+++ b/.circleci/build_examples.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+ERROR=0
+
+echo
+echo "Validate XML"
+python3 /documents/.circleci/validate_xml.py || ERROR=1
+
+echo
+echo "Lint AsciiDoc files"
+# this script prints non-ascii characters to the console
+PYTHONIOENCODING=UTF-8 python3 /documents/.circleci/lint_docs.py || ERROR=1
+
+echo
+echo "Compile headers"
+gcc /documents/headers/fmi3Functions.h || ERROR=1
+
+echo
+echo "Run CMake generator"
+cmake /documents/docs/examples/c-code -Bbuild || ERROR=1
+
+echo
+echo "Build examples"
+cmake --build build || ERROR=1
+
+echo
+echo "Run examples"
+./build/jacobian || ERROR=1
+./build/co_simulation || ERROR=1
+./build/model_exchange || ERROR=1
+
+if [ $ERROR != 0 ]; then
+  echo
+  echo "!!!!! At least one test failed, see above !!!!!"
+else
+  echo
+  echo "All tests passed."
+fi
+exit $ERROR
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,46 +1,12 @@
 version: 2
 jobs:
   build_examples:
-    docker:
-      - image: debian:stretch
+    machine: true
     steps:
       - checkout
       - run:
-          name: Install sudo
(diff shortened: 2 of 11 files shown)


friedrichatgc:
@t-sommer I think it would be good if all CI tests can be done also easily on a local machine. This is already possible for the build and check of the document using docker images but not for the build of the examples which is done only by CI. This draft PR adds a docker image usable to build the examples and uses this image also by CI.
Moreover the scripts builddoc.[sh|bat] and buildexamples.[sh|bat] are added to run the doc build/check and examples build locally using the same docker images as used by CI.
Please see also the PR commit message and review.

t-sommer:
Our original plan was to use (copy) the examples from https://github.com/modelica/Reference-FMUs/tree/master/examples where they're build and run on all platforms (Windows, Mac, and Linux).
I'm currently updating the Reference FMUs to catch up with the latest changes of the API and schema.

chrbertsch:
Regular Design Meeting

Torsten: Validation of C-Code in Reference FMUs. In fmi-standard we validate xsd and html.
The examples are already there. I am still working on the FMUs coping with the API changes.
Claudio: it is good to do this locally. I have a script
Andreas: We need the link between validation and the documentation
Torsten: one could use submodule. But people would have to understand it
We can reference and cross-link issues in both repositories.
Claudio: I will also look into this

clagms:
Pros:
- Scrips like these can and should be run by everyone before committing, as a smoke test.
- The circle ci stuff gets a simpler command with this PR. I like this. @t-sommer , does the use of a single command impact the feedback that circleci gives?

Cons:
- Even using docker, there's a fair chance that the scripts will diverge for multiple operating systems. For instance, the scripts given will create an arbitrary amount of container on whomever runs these frequently. An alternative way is to create a named container and then remove the container immediately after (this allows for the reuse of the same name for each command).
For example (in powershell):
```
$BUILDERID = 'fmi-builder'
$FMIStandarDir = (Get-Location).ToString()
docker run -v ${FMIStandarDir}:/documents/ --name $BUILDERID asciidoctor/docker-asciidoctor asciidoctor --base-dir /documents/ --backend html5 --failure-level WARN --verbose docs/index.adoc
docker rm $BUILDERID
```
- Another con is the fact that different people might want to do different things regarding the build. For instance, I might always use my little powershell script to generate the images from the schema, to always check if there is an inconsistency (this is important, as someone might change an xsd and not explain the changes in the text...).

As such, I think that we could keep the circle ci simplifications (provided they don't have an impact), but maybe avoid having scripts for different platforms on the repository. 
I think these can be shared informally among the participants. This helps keep circle ci as the source of truth and minimize inconsistencies.

t-sommer:
As discussed in the Regular Design Meeting today, I don't see the benefit of sharing these convenience scripts vs copying & pasting the commands from the [contribution guide](https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#building-the-specification-document).

The generation and validation of the HTML are already performed with Docker images, so you should get exactly the same results.

Issue #387, opened 2018-10-17
Changing parameter values in the modeldescription.xml file


modelica-trac-importer:
**Reported by cbertsch on 15 Oct 2016 11:06 UTC**
I would like to clarify, whether it is allowed to change paramter values in the modeldescription.xml file and what the consequences of this are. I know of tools, that do this, and it is unclear to me, what the importing tools shall do with such a modified FMU.

Changing parameters in the modeldescription.xml file could lead to making the GUID invalid (as the 2.0 FMI standard states: 
_Argument fmuGUID is used to check that the modelDescription.xml file (see section 2.3) is
compatible with the C code of the FMU. It is a vendor specific globally unique identifier of the
XML file (for example it is a “fingerprint” of the relevant information stored in the XML file). It
is stored in the XML file as attribute “guid” (see section 2.2.1) and has to be passed to the
fmi2Instantiate function via argument fmuGUID. It must be identical to the one stored
inside the fmi2Instantiate function. Otherwise the C code and the XML file of the FMU
are not consistent to each other. This argument cannot be null._

Additionally, the parameter values are also stored within the C-Code and/or DLL and I think that it is not specified, that these internal values have to be overwritten with the values from the XML when importing the FMI. Is this correct? I think that some importing tools will do this and other not. Thus a clarificatoin is suggested for FMI 2.0.1

----

Migrated-From: https://trac.fmi-standard.org/ticket/387

modelica-trac-importer:
**Comment by ottot on 19 Oct 2016 12:51 UTC**
In the table on page 53 it's quite clear that you only have to call fmi2SetXXX if you want a different value then the one stored in the xml file. This kind of implies that you cannot change the xml file and expect that to have any effect, as you say maybe it should be stated more clearly that your not supposed to change the xml file.

```
Initial or guess value of variable. This value is also stored in the C functions
 [Therefore, calling fmi2SetXXX to set start values is only necessary, if a 
different value as stored in the xml file is desired.]
```

modelica-trac-importer:
**Comment by cbertsch on 20 Oct 2016 18:51 UTC**
I agree. However, do we consider an FMU with manipulated parameter values in the modeldescription.xml as "corrupt"? (I would do so, as they could produce completely different results)
Can the compliance checker detect this? I think it should!
Should importing tools reject such FMUs or which parameter values should they take?

modelica-trac-importer:
[...]

PR #850, opened 2020-03-12
add statement to explicitly forbid changes in the xml

chrbertsch:
according to discussion in #387

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -3,7 +3,7 @@
 All static information related to an FMU is stored in the text file `modelDescription.xml` in XML format.
 Especially, the FMU variables and their attributes such as name, unit, default <<initial>> value, etc. are stored in this file.
 
-It is not allowed to make changes to the `modelDescription.xml` file.
+It is not allowed to change the `modelDescription.xml` file.
 _[Reason: The `modelDescription.xml` file has to be consistent with the binary or source code implementations._
 _Specifically, changes to the start values are not allowed.]_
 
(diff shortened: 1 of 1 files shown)


chrbertsch:
I was not sure whether 2.2 or 2.3 is the right place for statement; here it is in 2.2 dealing with the xml. One could also argue that changing modeldescription.xml takes place within the zip file, which would be in 2.3.  Please comment.

andreas-junghanns:
I am not sure if we should have two places where we point this out. But if in doubt, two are better, right? Should we make one into a non-normative text?

andreas-junghanns:
> I was not sure whether 2.2 or 2.3 is the right place for statement; here it is in 2.2 dealing with the xml. One could also argue that changing modeldescription.xml takes place within the zip file, which would be in 2.3. Please comment.

Thematically, both sections would work, but I would not know where to place this in 2.3 - so your choice seems more natural. And earlier is likely better than later in the document.

andreas-junghanns:
I should have submitted this as draft - and I have no idea how to get this into the draft status.

chrbertsch:
> I would put the statement in 2.3 if it was only in one location - 2.2 is concerned with the schema, not the purpose of the file. But both would be fine.

This brings us to the question whether we in 2.3 should more generally describe what ist allowed to change in an FMU
* modeldescription.xml --> no!
* source directory --> no?
* binaries --> new platform binearies may be added, others removed?
* documentation -->information might be added? existing documentation be changed?
* resources: ? Might e.g. characteristic map files stored in resources be changed/exchanged?
* extra --> yes! as specified

pmai:
> This brings us to the question whether we in 2.3 should more generally describe what ist allowed to change in an FMU

There is no support for changing any content of an FMU, with the sole exception of the extra directory entries.

This obviously does not mean that you cannot modify stuff inside an FMU when you know what you are doing and the circumstances warrant it, but there is no support for this and no requirement on FMU generators to enable such things...

I.e. rewriting the modelDescription.xml by e.g. removing variables works, because there is no way for the FMU or the importer to even find out about this, so for them this is a valid transformation. Nevertheless there is no support in the standard for this and no warranties are spelt out. And e.g. an FMU signing scheme might prevent this from occurring and thus break this transformation.

In general I don't think we should worry about this or spell out too much on this, since it only leads to confusion.

The special case of start values warrants additional language not because this wasn't clear already, but because we have seen common misuse in the wild, and thus should be explicit.

chrbertsch:
Regular Design Meeting: 

@andreas-junghanns : I will merge it.

Issue #818, opened 2020-02-26
co-simulation, basic co-simulation, hybrid co-simulation, scheduled co-simulation


CThuleHansen:
There are 3 interfaces, but  4 different names. The names are
1. co-simulation
2. basic co-simulation
3. hybrid co-simulation
4. scheduled co-simulation.

For example, `canNotUseBasicCoSimulation` refer to a defined set of features and certain interface functions. In essence, Basic Co-Simulation has been in a fluctuating state between FMI for Co-Simulation, Co-Simulation and Basic Co-Simulation. 

From the restructuring we have the following:
1. co-simulation is the CONCEPT.
2. Basic co-simulation is related to the the same thing as the modeldescription `canNotUseBasicCoSimulation`, i.e. the functionality and interface similar to FMI2.
3. Hybrid Co-Simulation is related to HYBRID Co-SIMULATION
4. Scheduled Co-Simulation is related to Scheduled co-simulation

This ticket is created to ensure that a check will be made to ensure consistency.

andreas-junghanns:
I started with a first pass to see what we need to use. Here is a suggestion slightly different to the first:

1. co-simulation: as the general concept
2. Co-Simulation (xCS): meaning all three Co-Simulation interfaces (that follow below)
3. Basic Co-Simulation (BCS)
4. Hybrid Co-Simulation (HCS)
5. Scheduled Co-Simulation (SCS)

We need 2. to have a good way to name all three FMI CS interface types - in many situations the text refers to all of them. I will create a PR with a first set of change proposal. Please check. Many places are still in need of fixing as descriptions for many things are still open.

I started this because we need to define a consistent way of using these terms.

t-sommer:
`canNotUseBasicCoSimulation` is obsolete and has been removed from the schema.

PR #851, opened 2020-03-12
Fix usage of term and writing of Co-Simulation and all variants there…

andreas-junghanns:
…of (#818)


chrbertsch:
Regular Design Meeting: 

Andreas: We shall merge this soon, in case of corrections create new PRs. Not close the Issue #818, because we will have to look into this longer.
Claudio: I will review this. (I will make this also consistent later with the glossary later)
Pierre: I agree to merge this soon.

clagms:
Shall we fix `docs\examples\c-code\VanDerPol\modelDescription.xml`?
I'm getting this on the xml validation:
```
Validate modelDescription.xml
Element 'BuildConfiguration': This element is not expected. Expected is ( VendorAnnotations ).
Element 'CoSimulation': This element is not expected. Expected is one of ( BasicCoSimulation, HybridCoSimulation, ScheduledCoSimulation, BuildConfiguration, UnitDefinitions, TypeDefinitions, LogCategories, DefaultExperiment, Terminals, GraphicalRepresentation ).
C:\srcctrl\andreas-junghanns\fmi-standard\docs\examples\c-code\VanDerPol\modelDescription.xml fails to validate
```

clagms:
There are multiple sentences referring to the schema that are outdated. Some refer to the change in the schema from `<xs:element name="CoSimulation" ... />` to `<xs:element name="BasicCoSimulation" ... />` while others are older inconsistencies.

Examples are:
```
_[If only one FMU is simulated, as `instanceName` attribute `modelName` or `<ModelExchange/CoSimulation modelIdentifier="..">` from the XML schema `fmi3ModelDescription` might be used.]_
```
```
The modelDescription has to include the Element with name `CoSimulation` to be allowed to set the `fmuType` to `fmi3CoSimulation`.
```

I feel that to fix this is not the main goal of the PR, so shall I open an issue for this @andreas-junghanns ?

clagms:
Some changes I am unable to review at this point, because I don't know enough yet. I'll approve these due to the urgency of getting this PR merged:
```
The data exchange between subsystems is **largely** restricted to discrete communication points.
```
```
** *Step Mode* (**Basic Co-Simulation and Hybrid** Co-Simulation): The variables listed as elements `<ModelStructure><Output>` with a floating point type and <<variability>> = <<continuous>> or <<discrete>>.
```

clagms:
The sentence `The concepts defined in this chapter are defined that are common for Model Exchange and Co-Simulation interface types (Co-Simulation for short)` seems to be conflicting with the use of xCS.
xCS is defined in the preamble, so maybe we should minimize redundancy and do not introduce a synonym in the above sentence. 
In short, maybe remove the `(Co-Simulation for short)` part.

clagms:
The sentence `or a co-simulation slave following one or the three Co-Simulation interface types Basic` seems to imply that the co-simulation slave can support one OR 3 co-sim interfaced types, but, e.g., not 2. Is this the intention?

clagms:
> // TODO: Is this true also for SCS?

I don't think so. As far as I remember, the intent is that each co-simulation interface type has its own mathematical model. If two mathematical models are redundant, then it is the later co-simulation interface type (e.g., SCS) will state that its mathematical model is similar to the one in Basic Co-simulation, and not the other way around (e.g. Basic Co-simulation, as a kind of super type, should abstain from making remarks related to the other kinds, which are kind of subtypes).

andreas-junghanns:
> 
> 
> Shall we fix `docs\examples\c-code\VanDerPol\modelDescription.xml`?
> I'm getting this on the xml validation:
> 
> ```
> Validate modelDescription.xml
> Element 'BuildConfiguration': This element is not expected. Expected is ( VendorAnnotations ).
> Element 'CoSimulation': This element is not expected. Expected is one of ( BasicCoSimulation, HybridCoSimulation, ScheduledCoSimulation, BuildConfiguration, UnitDefinitions, TypeDefinitions, LogCategories, DefaultExperiment, Terminals, GraphicalRepresentation ).
> C:\srcctrl\andreas-junghanns\fmi-standard\docs\examples\c-code\VanDerPol\modelDescription.xml fails to validate

Thorsten told me he is fixing these examples and I did not touch them (end the codes extracted into the standard document) and I am glad for that - I think this should be a different PR and that should fix other issues connected to the new interface type elements as well.

andreas-junghanns:
> 
> 
> The sentence `or a co-simulation slave following one or the three Co-Simulation interface types Basic` seems to imply that the co-simulation slave can support one OR 3 co-sim interfaced types, but, e.g., not 2. Is this the intention?

No, clearly not: The sentence should say "all or any subset of the interface types".

andreas-junghanns:
Your suggestions are plausible and won´t be the last changes. Please rewrite those sentences as you suggest - thank you!

clagms:
I've made those changes now.

andreas-junghanns:
ok, then I suggest to merge - Torsten, agreed?

Issue #569, opened 2019-04-15
Replace XSD screenshots with code listings


t-sommer:
The screenshots of the XSD schema elements are out-of-sync with the changed schema files.
They could be replaced with XML examples that can be validated during CI.

For example

![Boolean_Schema_large](https://raw.githubusercontent.com/modelica/fmi-standard/master/docs/images/Boolean_Schema_large.png)

could be replaced with

```xml
<Boolean name="bool_in"
  valueReference="8" 
  declaredType="switchState"
  start="false"/>
```

This would also save a lot of space (compared to the current PNGs), improve the rendering on small screens and provide valid (copy & paste friendly) examples.

chrbertsch:
I have always seen these pictures as part of the normative specification.
So how can they be replaced by an example? 
Will then the XSD pert describing e.g. the attributes of the boolean variable be included in the specification? But this is not so well readable as the generated pictures.

t-sommer:
Technically they are normative (not in square brackets). If we keep them, someone needs to update the screenshots.

Another possibility would be to rebuild them as ASCII-art. Something like

```
+ Boolean
  + declaredType (normalizedString)
  + start (boolean)
``` 

For the complex types this would be quite an effort though...

chrbertsch:
Regular Design Meeting: 
Karl: the Screenshots give a good overview. The screenshots must be consistent from the xsd filed.
Torsten: Right now they are inconsistent. Should they be removed
Klaus: https://stackoverflow.com/questions/32563774/automatically-generate-graphical-documentation-for-xsd-like-xmlspy-does
Torsten: Someone has to do it ....
Klaus: mark it as not up-to-date. Having the pictures in the document is nice.
Torsten: Add placeholders and add the screenshots later. 
Klaus: If we do not have screenshots, are they really important?
Karl: If you are new to it, they are nice.
Klaus: it is important, what is mandatory and what not

chrbertsch:
FMI Design meeting:
Torsten S.: I will do this.  (not super urgent)

nickbattle:
In support of this, I made some notes of the problems I could see. Note that in some cases the **text** following the diagrams is also yet to be updated, so it's not just a matter of the screenshots being updated:

2.2.1 The top level diagram is out of date (no BuildConfiguration, TypeDefinitions don't have SimpleType any more, no ScalarVariables any more, types and variables are Float64 etc rather than Real, ModelStructure no longer as 3 subfields, etc.)

[...]

PR #852, opened 2020-03-19
Auto generate xsd images

clagms:
This is a preview of the solution to #569.

Please have a look and give me feedback, so I can implement the full solution.

Changes and approach:
- I've modified the tool XSDDiagram to enable the automatic generation of the pictures from the schema (see the ./bin/ folder).
- Generated images are in ./docs/images/gen/ folder.
- Automatically marked attribute images with an "outdated" label (see limitations below).

Advantages:
- Automated generation of images (so they are kept consistent).

Limitations:
- Edit: I patched XSDDiagram to generate attributes (see dgis/xsddiagram#28 and dgis/xsddiagram#27).
- I haven't made sure it works on linux.

Note that the tools placed in ./bin/ are temporary and just so you can give me some feedback.


t-sommer:
The images in `docs/images/gen` are not being copied into the distribution archive in CI (all broken).
Also, without the attributes, they're kind of useless since all the information is stored there.
I think we should go with the semi-automated approach proposed by @CSchulzeTLK.

clagms:
Thank you @t-sommer !

>The images in docs/images/gen are not being copied into the distribution archive in CI (all broken).

Ups, my bad. Working on correcting this.

>Also, without the attributes, they're kind of useless since all the information is stored there.

I fixed the tool to produce the attributes (see https://github.com/dgis/xsddiagram/pull/28)
It now produces something like this:
image

>I think we should go with the semi-automated approach proposed by @CSchulzeTLK.

In the light of the fix above, it is now possible to plot the attributes. Are you sure that you want to go for the semi-automated approach? It's bound to lead to inconsistencies, and I don't want to be the one doing it manually :P
Just let me know so I don't spend more time on this!

chrbertsch:
> 
> 
> See `Limitations` in the first comment.

@clagms 👍  I like your approach. 
Is it possible to influence e.g. the spacing ?

andreas-junghanns:
I agree with Christian: if we can reduce spacing a bit to waste less space, this is a great solution because we drop even more redundancy - a process started by TorstenS a while back.
I also prefer the slightly cleaner look.

clagms:
> @clagms 👍 I like your approach.
> Is it possible to influence e.g. the spacing ?

Yes, this option is part of the tool, but is not exposed in the command line. I can expose it.

With the compact option, the drawing goes from this:
image

To this:
image

clagms:
Commit 78327ab9d3c818a46c012aa1345fbf9f0d680929 generates compact diagrams.

clagms:
> The images in `docs/images/gen` are not being copied into the distribution archive in CI (all broken).

@t-sommer I might need a bit of help with this.
I'm running the following command on the master branch, which is essentially copied from ci config:
```
docker run -v ${FMIStandarDir}:/documents/ -w=/documents/ --name $BUILDERID myubuntu zip fmi-standard.zip README.adoc CHANGELOG.adoc LICENSE.txt docs/index.html docs/fmi-spec.css docs/images/* headers/*.h schema/*.xsd
```
And I'm still not getting the images in the zip file. So this is an already existing problem, and was not introduced by the current PR, right?
If that's the case, I won't try to solve this here, to avoid mixing changes.

chrbertsch:
Regular design meeting:

@clagms : I will continue on this, perhaps with some help of @t-sommer 

Torsten: It we good enough to check in the images created with an open-source tool and document how to create them. We do not change the xsd. The people who change the xsd will know to do it.

clagms:
@t-sommer, @beutlich, I think I'm done with the PR. Could you have a look?

beutlich:
Nice to see that you forked xsddiagram and even prepared a release. What about downloading https://github.com/clagms/xsddiagram/releases/download/XSDDiagram_1.2_FMI/XSDDiagram_1.2_FMI.zip and extracting the binary in the powershell script for a fully automatic workflow (as I proposed earlier)?

beutlich:
I do not understand what the motivation was to move the images from docs/images to docs/images/gen. If generated or obtained differently should not matter.

t-sommer:
@clagms, why did you change the XSDs?

clagms:
> I do not understand what the motivation was to move the images from docs/images to docs/images/gen. If generated or obtained differently should not matter.

@beutlich This is a good question and I should have pointed it out:
- It should be clear to anyone who tries to edit those figures that they are automatically generated. Putting then in a gen folder is common practice to this effect.
- The script clears the folder and check whether there are overrides (an override can happen when a different XSD element with the same name is being generated). Overrides are nasty mistakes that might otherwise easily go undetected by someone who adds a new xsd picture to the doc.

Based on the above, is it clear why there must be a separate folder for the generated images?

clagms:
> @clagms, why did you change the XSDs?

@t-sommer This is explained in c3ef6c7fc56d36945451774f3fd41280bd954d94 :
- Removes a redundant declaration of an element that is the same everywhere it is used. This prevents inconsistencies in the future by unintentional modification of one the elements.
- Makes my life easier, as the figure generation tool tries to find all elements with a given name, and draws them. The tool does not compare two elements to see if they are duplicates. I worked around this for other "duplicate" elements, like "Float64" and similar, by specifying the schema where they are declared. In general, I think we should avoid declared new elements with the same name (in every schema), as this can lead to confusion for people who read the schema. For instance, the Float64 element inside the TypeDefinitions, should have a better name, such as Float64Type, to avoid confusing with the Float64 element insite the model variables.

clagms:
> Nice to see that you forked xsddiagram and even prepared a release. What about downloading https://github.com/clagms/xsddiagram/releases/download/XSDDiagram_1.2_FMI/XSDDiagram_1.2_FMI.zip and extracting the binary in the powershell script for a fully automatic workflow (as I proposed earlier)?

@beutlich I can do that with no problem. However, I got the impression from the web meeting that @t-sommer is not very keen on having binaries in the repository. That's why I made the release and the instructions, so that anyone who wants to automate this on their side, can do so while not having binaries in the repo.

beutlich:
> having binaries in the repository

Sure. The workflow proposed in https://github.com/clagms/fmi-standard/pull/1/files#diff-f892b8c44b29d86a02974bc00e9c24c7R18-R23 therefore is not about having binaries (EXE files) in the repository, but how to get and use them in an automates way for the easiest way of reproducing the PNGs. If you add these EXE files to the gitignore, you can be sure to not commit them by accident. So, after all, it is about extending the powershell script.

beutlich:
> Based on the above, is it clear why there must be a separate folder for the generated images?

Actually, it is not. You simply can add a Readme file to doc/images, stating, that this directory will be used as output directory for xsddiagram, and that PNGs should not be edited manually. Then you simply can use doc/images directly as output directory and get rid of the intermediate step using doc/images/gen.

clagms:
> > having binaries in the repository
> 
> Sure. The workflow proposed in https://github.com/clagms/fmi-standard/pull/1/files#diff-f892b8c44b29d86a02974bc00e9c24c7R18-R23 therefore is not about having binaries (EXE files) in the repository, but how to get and use them in an automated way for the easiest way of reproducing the PNGs. If you add these EXE files to the gitignore, you can be sure to not commit them by accident. So, after all, it is about extending the powershell script.

Ah, I think I understand what you mean now @beutlich . Do you want to change the script to automatically download the exe file? Can you make that change?

clagms:
> > Based on the above, is it clear why there must be a separate folder for the generated images?
> 
> Actually, it is not. You simply can add a Readme file to doc/images, stating, that this directory will be used as output directory for xsddiagram, and that PNGs should not be edited manually. Then you simply can use doc/images directly as output directory and get rid of the intermediate step using doc/images/gen.

How would you change the script in order to avoid overwriting a generated png @beutlich ? If you can make that change, I'm fine with this! Beware that there are many other figures in the images folder, that are not generated. If you decide to create the README file, make sure that people will know which figures are generated, and make sure to keep that consistent as more generated figures are added.

beutlich:
> Do you want to change the script to automatically download the exe file? Can you make that change?

See https://github.com/clagms/fmi-standard/pull/2, where I added (and improved) it again.

beutlich:
Regarding docs\images\gen. I have no strong opinion here. However, there are also drawio generated SVG files in doc\images along with their XML configuration. I had simply expected, that generate_figures.ps1 generates or regenerates all schema PNGs in doc\images and does not care if the file already exists. (We have version control for that reason.)

clagms:
> See [clagms#2](https://github.com/clagms/fmi-standard/pull/2), where I added (and improved) it again.

Thank you so much @beutlich ! I've merged this!

>Regarding docs\images\gen. I have no strong opinion here. However, there are also drawio generated SVG files in doc\images along with their XML configuration. I had simply expected, that generate_figures.ps1 generates or regenerates all schema PNGs in doc\images and does not care if the file already exists. (We have version control for that reason.)

It's a matter of personal opinion, but I'd rather be more defensive, and ensure that no override happens unintentionally. For example, if the following instructions are used:
```
Export-Schema Float64 3 -schema fmi3Type.xsd 
Export-Schema Float64 3
```
There will be an overwrite. I'm afraid the source control will not be of much help here, since it will only show that one element has been modified. If the person who added the second instruction above does not pay close attention, an override has been made, and the doc will have two pictures which are the same, but one of those references are wrong. 
One requires visual inspection of the doc to later detect this error and that seems too much to ask.

An alternative is to rewrite the script into accepting an array of parameter records, and checks that no duplicates are generated.

>However, there are also drawio generated SVG files in doc\images along with their XML configuration.

I was unaware of this. I recommend that those be put in a separate gen folder (not the same as the xdiagram one.. we can reorganize this). It's very easy to edit SVG using inkspace, so I wouldn't be surprised if at some point someone does some improvements on the generated figures.

CONTRIBUTING.adoc
```diff
@@ -55,8 +55,6 @@ When you make changes to the XSD Schema, it is your responsibility to ensure tha
 
 To re-generate the images, follow the steps:
 
-. Download the latest XSDDiagram binaries from https://github.com/clagms/xsddiagram/releases.
-. Unzip and place the binaries in the folder `<fmi-standard>/bin/xsdiagram` so that, e.g., the path of the binary `XSDDiagramConsole.exe` is `<fmi-standard>/bin/xsdiagram/XSDDiagramConsole.exe`
 . On a powershell terminal placed at directory `<fmi-standard>`, run `.\bin\generate_figures.ps1`.
```
beutlich: "placed at directory" no longer is necessary as the script works from any directory now.

beutlich: Simply:

Run `PowerShell -Command "path\to\generate_figures.ps1"`.

CONTRIBUTING.adoc
```diff
@@ -55,7 +55,7 @@ When you make changes to the XSD Schema, it is your responsibility to ensure tha
 
 To re-generate the images, run the following:
 +
-PowerShell -Command "<fmi-standard>\bin\generate_figures.ps1"
+  PowerShell -Command "<fmi-standard>\bin\generate_figures.ps1"
```
beutlich:
```sh
PowerShell -Command "<fmi-standard>\bin\generate_figures.ps1"
```

CONTRIBUTING.adoc
clagms: Thank you @beutlich . Fixed now!

CSchulzeTLK:
I am searching for the information if an element is optional. Is this visible in the images?

CSchulzeTLK:
... just found it

Issue #854, opened 2020-03-20
Add feature matrix for interface types


t-sommer:
to give an overview of the differences and similarities of the four interface types

- Model Exchange
- Basic Co-Simulation
- Hybrid Co-Simulation
- Scheduled Co-Simulation

w.r.t. to the provided features

- solver included / external
- event indicators
- early return / intermediate update
- clocks
- direct feedthrough
- (please add more)

chrbertsch:
Regular Design Webmeeting: 

Torsten : Have a general overview in the introduction.
It helps people e.g. to select the right interface.

Collection: 
* you either have no continuous states or a solver inside the FMU except for ME
* no direct feedthrough in BCS, clarify for HCS (see #793)
* event indicators
* event iteration
* continuous states

--> @t-sommer : I will create a PR

andreas-junghanns:
The beginning is made, during web design meeting we discussed what we want to have in the text and table (references to concept descriptions). A new issue will be created to follow this: #865

PR #862, opened 2020-03-23
Add feature matrix for FMI interfaces

t-sommer:
solves #854


andreas-junghanns:
Web Design Meeting: we merge this because this is the beginning of a great journey! We will use the table to have references into the text (likely the different Mathematical Model sections) where the features/concepts are described for the different interface types.

Issue #738, opened 2020-02-10
Attribute "previous" missing in fmi3Variable.xsd


friedrichatgc:
The specification defines an attribute called "previous" for clocked state variables. However, the fmi3Variable.xsd file does not contain a definition for such an attribute.

andreas-junghanns:
AndreasP: This might have slipped your attention. Can you comment at least before you leave?

andreas-junghanns:
Web design meeting: we cannot see why such an attribute is needed (we think it is a leftover of some version of the clocks). We will drop this from the text. See issue #853 for the same issue.

chrbertsch:
@KarlWernersson : will check this again

andreas-junghanns:
@KarlWernersson : any progress here?

KarlWernersson:
@andreas-junghanns : Not yet

chrbertsch:
(Andreas J, Christian B)
@KarlWernersson : would you please propose a PR for the design meeting?

andreas-junghanns:
Karl: This is from the synchronous Modelica. "previous" is the previous value of a state before the clock tick during super-dense time.
Klaus: Why do we need this if we know the current value?
Otto: During algebraic loops, you reset the discrete states, using the previous.
Klaus: Why not have this in master?
Karl: The original FCP describes this...
Klaus: What is previous?
Karl: It is the value of the state of the previous super-dense time instant.

Andreas: Conclusion: We currently see no use-case for the previous value that the master cannot store the previous value itself. Unless we find a use-case where the master cannot know the "previous" value, we should not shift the burden on the FMU to provide this value "just-in-case".

PR #866, opened 2020-03-27
Remove non-existent previous attribute documentation

### LINKED ISSUE #853: Please clarify the name of the attribute "previous" in Discrete time state
This issue comes as a result of a discussion on the solution to #738 .
The name "previous" conveys little meaning as to what this attribute should contain.

In the standard, it says: "Discrete-time states are listed in the ModelStructure and have the variable xml-attribute previous referring to the variable holding the previous value of the state."
I guess that "previous value" means the value of the discrete time state at the previous tick (in the same sense that $x_{j-1}$ refers to the previous value of $x_j$).
Is that correct? 

Or does "previous" mean the value at the previous super dense time instant?

pmai:
Same as future perfect it was found not to be...
Fixes #853 and #738


andreas-junghanns:
I found a few more dead xrefs and removed those.

chrbertsch:
Regular Design Meeting:

Markus: I will review it. 

Klaus: there was a discussion in the hybrid CS group, do we remember it?

Andreas: The text does not explain it.  If someone needs it, it could be re-added.

Klaus: there were three interpretations

Markus: none of it really makes sense

Pierre: it is not in the Model Description. We should remove it now, but ask @KarlWernersson .

andreas-junghanns:
I sent an email to Karl, but got no response. I take this as a "don´t care" or "don´t care now". Merging now.

KarlWernersson:
Had to take care of sick daughter yesterday. So i was not available.
Please let me know, why was previuous droped?
Im still trying to catch up on what has happend while I was away

andreas-junghanns:
@KarlWernersson: The way the previous attribute was (not) described let us to believe that this was a leftover from an intermediate stage of the hybrid Co-Sim group. If you want to reintroduce it, please create a PR with explanations and all such that we know the semantics, XSD changes, use cases etc.

andreas-junghanns:
Here are a few questions that need to be answered:
“previous state” needs better definition:
 - previous doStep time?
 - Left-side value at events?
 - event iteration last super-dense time step?
Use case:
 - Why would the solver need to know the previous state from the model?
 - The solver knows whatever it needs from previous by memorizing itself, not so?

Issue #868, opened 2020-03-27
Remove old interface type enums and fix instantiate function descriptions


andreas-junghanns:
We removed the enum types for the instantiation function and split it into 4 different functions.
And we need to describe the new functions with their new arguments.

PR #869, opened 2020-03-27
Remove old interface type enums; fix function description for instantiation functions (#868)

andreas-junghanns:
according to #868


chrbertsch:
Regular Design webmeeting. 

We shall merge this soon before #872 

Torsten S: I will look into this.

chrbertsch:
perhaps only a small issue: @andreas-junghanns : in your changes you refer to "Co-Simulation Interface Types". But in the sense of xml/xsd in https://github.com/andreas-junghanns/fmi-standard/blob/fmi3_remove_old_enums_and_fix_function_descriptions/schema/fmi3InterfaceType.xsd
there is one Co-Simulation type; we only have different Co-simulation elements in the modeldescription.xml, see https://github.com/modelica/fmi-standard/issues/876, right?

andreas-junghanns:
> 
> 
> perhaps only a small issue: @andreas-junghanns : in your changes you refer to "Co-Simulation Interface Types". But in the sense of xml/xsd in https://github.com/andreas-junghanns/fmi-standard/blob/fmi3_remove_old_enums_and_fix_function_descriptions/schema/fmi3InterfaceType.xsd
> there is one Co-Simulation type; we only have different Co-simulation elements in the modeldescription.xml, see #876, right?

I was looking for an easy way to name all three Co-Sim types at once (instead of listing all three all the time which makes the text unreadable). Do you have a better proposal?
Maybe we can change this later? Can we merge these changes, soon, please? I hate merge conflicts because fixing those is error prone and might introduce new problems...

PR #872, opened 2020-03-30
Unify formatting of XML elements and XML attributes

andreas-junghanns:


andreas-junghanns:
This is still inconsistent in a larger sense - now only a subset of XML tags is formatted as `<tag>`.
Do we want to format all XML tags like that?

t-sommer:
Can't we get rid of the `<fmiModelDescription>` prefix? It's the root element and always present.

andreas-junghanns:
> Can't we get rid of the `<fmiModelDescription>` prefix? It's the root element and always present.

It might make the text easier readable, but harder to understand because the reader must every time figure out where the attribute or element belongs to. The way we have it now, it is longer, but helps to instantly place the attributes within the XML structure.

t-sommer:
> but harder to understand because the reader must every time figure out where the attribute or element belongs to

Their names are unique within the "root" element, so what's the problem here? It's like the `<body>` tag in HTML: it's always within the enclosing `<html>` element.

andreas-junghanns:
> 
> 
> > but harder to understand because the reader must every time figure out where the attribute or element belongs to
> 
> Their names are unique within the "root" element, so what's the problem here? It's like the `<body>` tag in HTML: it's always within the enclosing `<html>` element.

The problem is that no all of us know the standard so well that we immediately know where a certain XML lives. And yes, I could look it up every time, but what is the point? We can help in many cases with adding a bit of context and that makes the text more easily understandable for most readers.

chrbertsch:
Regular Webmeeting: 

Torsten S: This is good. The Dot-notation is inconsistent.
I will merge it. 

Andreas: I will update the formatting guides

Markus: How do we format attributes?

Torsten S: as in the XML

Andreas: I will continue with this PR to make formatting more consistent. There are still inconsistencies.

andreas-junghanns:
I finished checking all XML elements and all XML attributes for their formatting and use.
For some, this is essentially impossible, like `value` or `min`, I tried my best.

Can you check the changes now (and my merge)?

chrbertsch:
1.) Line 77 of https://github.com/andreas-junghanns/fmi-standard/blob/fmi3_fix_formatting_of_fmiModelDescription/docs/2_1_common_api.adoc

Must it not be:

` ... defined in <fmiModelDescription><ModelExchange> or <fmiModelDescription>><{Basic|Hybrid|Scheduled}CoSimulation> together with pass:[_] at the end (see <<ModelExchange>>, <<basic-co-simulation-schema>>, <<hybrid-co-simulation-schema>>, <<scheduled-co-simulation-schema>>,) ... `

instead of 
`... defined in <fmiModelDescription><ModelExchange> or <fmiModelDescription><CoSimulation> together with pass:[_] at the end (see <<ModelExchange>> and <<basic-co-simulation-schema>>). ...` ?

2. ) How shall we format element names in tables?

e.g., in 2.2.1 
grafik

Shouldn't this be `<ModelExchange>` etc. ?

andreas-junghanns:
> 
> 
> Looks good and can be merged from my side.
> 
> Only found 2 more cases where I think changes should be applied, see [#872 (comment)](https://github.com/modelica/fmi-standard/pull/872#issuecomment-609391097)

I have fixed the first of your comments.
The second issue is on my list - see my email.
But I do not want to start this without the two big PRs being merged to avoid merge conflicts.

PR #879, opened 2020-04-06
Change the definition of Boolean and String to avoid using an abstract

nickbattle:
Amended fmi3Type.xsd to avoid abstract type errors for Boolean and String.

### DIFF
--- a/schema/fmi3Type.xsd
+++ b/schema/fmi3Type.xsd
@@ -133,8 +133,20 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 					</xs:complexContent>
 				</xs:complexType>
 			</xs:element>
-			<xs:element name="Boolean" type="fmi3TypeDefinitionBase"/>
-			<xs:element name="String" type="fmi3TypeDefinitionBase"/>
+			<xs:element name="Boolean">
+				<xs:complexType>
+					<xs:complexContent>
+						<xs:extension base="fmi3TypeDefinitionBase" />
+					</xs:complexContent>
+				</xs:complexType>
+			</xs:element>
+			<xs:element name="String">
+				<xs:complexType>
+					<xs:complexContent>
+						<xs:extension base="fmi3TypeDefinitionBase" />
+					</xs:complexContent>
+				</xs:complexType>
+			</xs:element>
 			<xs:element name="Binary">
 				<xs:complexType>
 					<xs:complexContent>


schema/fmi3Type.xsd
```diff
@@ -133,8 +133,20 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 					</xs:complexContent>
 				</xs:complexType>
 			</xs:element>
-			<xs:element name="Boolean" type="fmi3TypeDefinitionBase"/>
-			<xs:element name="String" type="fmi3TypeDefinitionBase"/>
+			<xs:element name="Boolean">
+				<xs:complexType>
+					<xs:complexContent>
+						<xs:extension base="fmi3TypeDefinitionBase" />
```
pmai, line 139:
```suggestion
						<xs:extension base="fmi3TypeDefinitionBase"/>
```

schema/fmi3Type.xsd
```diff
@@ -133,8 +133,20 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+					</xs:complexContent>
+				</xs:complexType>
+			</xs:element>
+			<xs:element name="String">
+				<xs:complexType>
+					<xs:complexContent>
+						<xs:extension base="fmi3TypeDefinitionBase" />
```
pmai, line 146:
```suggestion
						<xs:extension base="fmi3TypeDefinitionBase"/>
```

nickbattle: Ah yes, thanks. Copy/paste layout problem from the issue.

andreas-junghanns:
Sorry, my XSD is a little rusty and my vote would carry hardly any weight... I trust Pierre to know what is good for us.

t-sommer:
@pmai, do we need a new tag for this change?

schema/fmi3Type.xsd
t-sommer: @nickbattle, please check "Allow commits by maintainers" when you open your next PR.

nickbattle: @t-sommer Yes, I would if I could, but for some reason when I create PRs via the INTO-CPS fork that option does not appear where it should. We're not sure why. Any idea?

pmai:
@t-sommer tag as in version tag? Depends on your view. In terms of the version number in the XML, it probably does not matter, since this is a strict superset of the tagged version, and this was obviously just a bug. In terms of people using the right schema for validating their test FMUs, probably a new tag would be a good idea...

Issue #877, opened 2020-04-05
Fix formatting of table cells


andreas-junghanns:
We have a number of tables where we have cells that miss proper formatting.

PR #880, opened 2020-04-06
Fix table cell formatting (#877)

andreas-junghanns:


andreas-junghanns:
This was even more than I thought: please check especially those tables that I had to reformat all lines.

andreas-junghanns:
Please review so we can merge, the next PR I am preparing will be another big one.

Issue #884, opened 2020-04-08
canNotUseBasicCoSimulation, providesHybridCoSimulation and providesScheduledCoSimulation 


andreas-junghanns:
These attributes are mentioned in the text, but do not exist in the schema files.

andreas-junghanns:
I do not like two things here:
canNotUseBasicCoSimulation contains a negation which makes it harder to understand.
canNotUseBasicCoSimulation could more consistently also be named providesBasicCoSimulation (like the other two), but have a default of "true" to express our expectation that BCS is always implemented.

nickbattle:
How do these flags relate to the elements that define the various FMU types? For example, does `canNotUseBasicCoSimulation` mean that there cannot be a `<BasicCoSimulation>` element? So why is presence or absence of the `<XXXCoSimulation>` elements not enough?

t-sommer:
I think the attributes are superfluous with the introduction of the `<*CoSimulation>` elements.

andreas-junghanns:
> I think the attributes are superfluous with the introduction of the `<*CoSimulation>` elements.

Ideally, this would be the cleanest solution.
But, right now, the text says (or at least implies) that if you have HCS or SCS elements in the XML, BSC can be used as well, without the XML element present.

I think/suggest this is unnecessarily complicated. If the FMU can do BCS, add the element. No implicit complications.

t-sommer:
Solved with #887.

PR #885, opened 2020-04-08
Add missing attributes to CoSim element (see #884)

andreas-junghanns:
This is a draft still because this should be discussed, see #884.


andreas-junghanns:
We will solve this by removing the attributes, different PR to follow

PR #887, opened 2020-04-09
Remove obsolete attributes from docs

t-sommer:
- canNotUseBasicCoSimulation
- providesHybridCoSimulation
- providesScheduledCoSimulation

#solves 884


andreas-junghanns:
Christian: we removed three XML attributes here that we think are not needed anymore since we have three different kinds of CS elements in the model description. I would merge this if you approve.

Issue #889, opened 2020-04-09
Clocks with Dimension allowed?


IZacharias:
According to the schema, a clock variable can be an array and have several dimensions:
![image](https://user-images.githubusercontent.com/33029000/78892458-ae247100-7a69-11ea-9c0f-33aafc6c9070.png)

This must not be allowed for clock variables (or does anyone know of a reasonable use case for this?). So the dimension element must be removed from the clock element.

t-sommer:
Currently the `<Dimension>` sequence is defined in the base class `fmi3VariableBase`. We could either state in the docs that this is not allowed or force it via the XSD.

t-sommer:
@pmai, what do you think?

andreas-junghanns:
If the schema does not get more complicated, I would prefer forcing this via XSD.

chrbertsch:
Regular FMI Design meeting

Pierre: xsd-Schema-wise we cannot override it. 
Do we have a problem with the arrays?

Irina: I think so

Pierre: Are these variables or scalar variables?

Irina: for structural parameters we have restrictions.

Claudio: What does it mean for a clock to be active? Perhaps this is where the problem lies.

Pierre: Its just multiple clocks

Claudio: What do we do with event indicators in ME?

Torsten: this is unrelated. I would follow Pierres argumentation. Also in Modelica one can express arrays of clocks.

Pierre: I would even allow for structural parameters to be arrays.

Torsten S: What about the dependencies? What happens when the clocks are arrays.

Irina: It would not fit with the API to have arrays.

...

Torsten B.: In the wider sense a clock is like a variable

...

Pierre: for the setClock API this is similar to the setBoolean API. This should work for clocks.
What about dependencies?

Claudio: For an array of input clocks ...

Pierre: as for booleans one could set the components of a clock array to individual values
If we cannot treat clocks as variables, we should define them to be something else. We should not only fix the syntax.

Irina: We would remove the variables from the varibles and make them a different type.

Andreas: We would need  a special way to identify clocks. 

TorstenS: We could define base classes

Pierre: I would allow arrays of clocks.

Andreas J: We should answer: Why should clocks not be arrays?

Pierre: I would avoid special cases. Perhaps in the next release we need them as arrays.

Torsten S: What are is an array of clocks? you could set them only at the same time. The information is in value reference.

Pierre :The API must support activating only part of the clocks of an array

Torsten: The clocks have a double meaning for clocks and model partition

[...]

PR #891, opened 2020-04-09
Minimal changes to fully support clock arrays

pmai:
This changeset includes the minimal changes necessary to fully allow the
use of arrays of clock variables. Further refactoring could be done
based on this to merge the Set/GetClock functions into the getter and
setter sections of the document, and other cleanups might be wanted, but
this should demonstrate what is basically needed to do this.

The special treatment of index 0 follows the convention used for
dependency information, but could be dropped if not wanted.

This addresses issues raised in discussion of #889 and should
serve as further discussion basis of how to proceed.


t-sommer:
@pmai, can we merge #894 first to get rid of the C code in the fmi-standard repo?

docs/2_2_common_schema.adoc
```diff
@@ -1402,6 +1402,13 @@ It is not allowed to combine <<outputClock,`output clocks`>> with periodic or st
 
 The attribute <<clockReference>> must not be used if <<causality>> = <<output>> for a variable.
 
+|`clockElementIndex`
+|
+[[clockElementIndex,`clockElementIndex`]]
+The optional attribute <<clockElementIndex>> is used in conjuction with the <<clockReference>> attribute to define the element index of the <<clock,`clocks`>> this variable is assigned to for array clock variables.
+The element index is the 1-based index of the element (with serialization of indeces for multi-dimensional arrays in the usual row-major order).
```
t-sommer, line 1409: Shouldn't the indexing be consistent with the rest of the API (i.e. C-style)?

docs/2_1_common_api.adoc
```diff
@@ -845,14 +845,17 @@ include::../headers/fmi3FunctionTypes.h[tag=SetClock]
 ----
 
 Sets <<clock,`clocks`>> activation status by providing the value references of the corresponding <<clock>> variables and their values.
+Note that <<clock>> variables, like any other variables can be scalar or array variables.
```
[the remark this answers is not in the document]
andreas-junghanns: > Do we have to mention this for each data type, if it's consistent?

Maybe this is better in non-normative text as a hint. But the confusion that we had in the Design Meeting indicates that this is worth mentioning.

docs/2_2_common_schema.adoc
andreas-junghanns: This is consistent with `elementIndicesIndependents` - but I agree this is odd, non-C-style and will likely lead to errors. Using 0 as the catch-all is arbitrary. Since size_t is unsigned, -1 is out. MAX_SIZE_T does not exist (or does it?). But can we define it in fmi3PlatformTypes.h?
Better ideas, anyone?

pmai: @t-sommer, @andreas-junghanns Not that I like this convention, and for the clocks we might not actually need this, but if we do this thing in the dependencies API, we'd better stay consistent to it. If we want to change this, I think we should do this in a separate pull request, where we change both sites.

Don't think that I have any better ideas though, since the -1 (UNSIGNED MAX) thingy seems problematic, since it does not work very well for the XML part, really.

docs/2_1_common_api.adoc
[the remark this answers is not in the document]
pmai: The reason I remention it is that the clocks stuff is in a seperate section, away from the other variable types, hence there easily is confusion. If that seems problematic, I'd rather opt to move the Set/GetClock stuff to the getter / setter section, however that will pull it from the other clocks API...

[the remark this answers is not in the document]
t-sommer: I have no strong opinion here, just want to avoid duplication.

pmai:
> @pmai, can we merge #894 first to get rid of the C code in the fmi-standard repo?

Sure can do, in that case we can drop the 2f1ed09 commit because it is no longer needed (or rather is needed against the reference fmu repo...

schema/fmi3Variable.xsd
```diff
@@ -381,6 +381,7 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 		<xs:attribute name="canHandleMultipleSetPerTimeInstant" type="xs:boolean"/>
 		<xs:attribute name="declaredType" type="xs:normalizedString"/>
 		<xs:attribute name="clockReference" type="xs:unsignedInt" use="optional"/>
+		<xs:attribute name="clockElementIndex" type="xs:unsignedLong" use="optional"/>
 		<xs:attribute name="intermediateAccess" type="xs:boolean" use="optional"/>
 	</xs:complexType>
```
IZacharias, line 387: An additional change in fmi3Variable.xsd is necessary: The fmi3Clock's start attribute 
`<xs:attribute name="start" type="xs:boolean"/>`
must be a list (same as for other variables):
```
<xs:attribute name="start">
	<xs:simpleType>
		<xs:list>
			<xs:simpleType>
				<xs:restriction base="xs:boolean"/>
			</xs:simpleType>
		</xs:list>
	</xs:simpleType>
</xs:attribute>
```

clagms:
@pmai, have a look at text ``It is not possible to use a variable of type `<Clock>` in arrays``.

pmai:
> @pmai, have a look at the following:
> 
> It is not possible to use a variable of type `<Clock>` in arrays
> It is not possible to associate more than one <<clock>> to a variable.

I'm not sure what you mean to say here?

schema/fmi3Variable.xsd
pmai: Yes indeed, will be added.

pmai:
> @pmai, have a look at the following:
> It is not possible to use a variable of type <Clock> in arrays
> It is not possible to associate more than one <> to a variable.
> 
> I'm not sure what you mean to say here?

Disregard, these were quotes of the current tree, thanks, will fix those as well.

pmai:
Actually on the second point:

> It is not possible to associate more than one <<clock>> to a variable.

I think this is still true for the st clocks, so the sentence should stand; however for the communicationPoint clocks, it states that multiple clocks can be associated, however I don't see how this is supposed to work with the current schema, so I think this needs fixing one way or the other in another ticket/PR.

pmai:
I will rebase this against current master (and drop the examples commit) prior to merging, but for now it stays against the old master to allow CI testing to remain useful.

chrbertsch:
Regular design Webmeeting: see discussion https://github.com/modelica/fmi-design/tree/master/Meetings/2020-04-17-Regular-FMI-Design-Webmweeting#httpsgithubcommodelicafmi-standardpull891

t-sommer:
@pmai, do you want this merged before tagging the `v3.0-alpha.4`?

andreas-junghanns:
FMI Design Meeting: decision to merge this PR now and work through the remaining issues later, when they come up.

clagms:
Detailed discussion.


Not much changed since last meeting.

Irina summary of potential issues with clock arrays: I don't see it as necessary to have clock arrays.

Andreas: I don't see a reason to see clocks differently.

Claudio: we can study these issues later when we do the mathematical model.

Pmai: I think if there are issues later, I think they will be syntactic. There are going to be some. Maybe there will be problems with dynamically changing array sizes.

Pmai: problems might show up not during the mathematical models, but will 

AJ: Let's merge this and then try to find problems with it.

TS: Why don't we let the master deal with support for this.

AJ: Importers and Exporters have the option of not supporting this.

PMain: I don't think this adds that much complexity. If we have tools that support it, the better.

AJ: The useful usecase is the possible instantiation of similar models. 

PMai: all clocks in an array are independent.

AJ: Clocks are a kind of clocks. They have a diff meaning than normal variables.

CG: Clocks don't hold values. When you set it, it looses its value immediately. 

(Lots of discussion).

CG: Is it possible to save and restore the state. We probably need to revisit these in the mathematical model.

PMAI: We need examples for each of these features.

Issue #795, opened 2020-02-23
Maybe use numbered environments in the standard and adopt more rigorous language


clagms:
In the standard, there are multiple helpful environments, such as non-normative text, example, etc.
In technical document, these are normally done with numbered environments.
For example:
```
Definition 1. FMU. An FMU is...
Example 1. Consider the following master algorithm.
Remark 2.1. fmi3Get is not allowed after a fmi3Set without a doStep in between.
```

The beauty of using these is that they can be referred to explicitly throughout the text, without having to refer to a section, without using hidden links, and reducing the redundancy (and therefore inconsistencies) in the text. The current links are problematic as it is not clear to the reader whether they exist or not (without hovering the mouse over them). 
The same applies for references to figures and tables, which should always be explicit (that is, don't use expressions like "Table below").

We could also adopt a more rigorous language (this is not a personal critique to whomever worked hard on the standard... it's all too easy not to be rigorous :p), being careful to define new terms before they are used (using definition environments), or using a table for all terms used.

I could offer some time to improve the text in this way, once there is some consensus about portions of the standard. Naturally, we will need multiple pairs of eyes checking for ambiguities. Words such as "this", "it", should be avoided, and the use of synonyms should be forbidden, except when defining a new term. 

This is related to #650 but affects the content in a more drastic way, and #459

chrbertsch:
@clagms : before applying this to the whole document: Could you prepare individual examples or PRs how this would look like (for figures, equations, ...)?

andreas-junghanns:
@clagms: Are you still planning to propose this? We tried a while back and the PR proposal is still around and the result wasn´t just ugly, it was totally unreadable for humans. Navigation is impossible, the TOC explodes. I am not sure either it helps you, because if we add something the numbering changes, right? Please comment and preferably close this ticket.

clagms:
I am still planning this @andreas-junghanns , but no actual work done here yet. I'm sorry I haven't been very responsive on this. I'm trying to formalize the hybrid cosim interfaces first.

>We tried a while back and the PR proposal is still around and the result wasn´t just ugly, it was totally unreadable for humans.

Can you please point me to that PR? I can take a look and it will help me get a feel for what you are saying.

>Navigation is impossible, the TOC Explodes.

Maybe I'm missing something, but I don't understand why the navigation is impossible.
I think that these environments do not have to show up in the TOC.
So, for instance, if we have the text:
```
[...]

PR #895, opened 2020-04-13
Expand glossary and unify terminology for sections 1-2 (inclusive)

clagms:
This PR expands the glossary, tries to improve clarify and unify terminology (https://github.com/modelica/fmi-standard/issues/795).

Updating the glossary forces me to read the whole document, so this PR will invariably contain improvements here and there that are not related to the glossary, but are too small to create an issue. If I'm not sure that there is an improvement to be made, I won't make the change (see https://github.com/modelica/fmi-standard/issues/898).

This is just a draft, as the changes are ongoing, but span the whole document.
Example changes made so far are:
* Identified some more terms that I think should be in the glossary. I haven't given a definition to them yet. This is because I keep finding similar terms, so I need to identify all the synonyms first.
* Minor improvements in the English.
* Replacement of synonym terms when it is obvious (to me).
* Fixed typo in Overview, where the implementation of caching was a separate guiding principle.
* Identified and fixed ambiguities up to section 2.1.3.
* Removed details from overview section into the corresponding subsection of chapter 2.
* Preamble of chapter 2 now explicitly refers to the all subsections in a consistent manner.

Issues that need addressing, but I'm not sure how and would like to hear your opinion, are listed in https://github.com/modelica/fmi-standard/issues/898.


clagms:
Some more issues that need to clarified:
1. In sentence `triggering execution of synchronous and asynchronous external events for all exposed model parts of a set of FMUs`, it's not clear why are we referring to a set of FMUs. There's a lot of implicit information in this sentence. Are those FMUs all implementing the Scheduled Cosim interface? I think it's best is we keep the description between a master algorithm and a single FMU, to be consistent with the other cosim interface descriptions.

clagms:
Thank you @andreas-junghanns! I understand that mixed changes are bad, but it's impossible for me to note and not fix every single thing that's bad as I go through the whole document, only to fix them later in a separate PR(s). There's so many of those that just keeping track of them will double my workload. I'm already adding the outstanding issues to this PR as comment, to see if I can get a quick feedback on how to fix them.

What do you propose I do?
I can put the argument/parameter mismatch in a separate commit, but I doubt that I'll be able to do so for everything I encounter...

clagms:
Commit https://github.com/modelica/fmi-standard/pull/895/commits/6c86c0ec52bb3566a7f557ec622e4ad6f7005336 fixes the following:
>You have also started renaming "arguments" to " "parameters".
This should be a separate commit for sure and should fix this throughout the entire document.

andreas-junghanns:
> What do you propose I do?
> I can put the argument/parameter mismatch in a separate commit, but I doubt that I'll be able to do so for everything I encounter...

To have "minor fixes" in one commit is ok, but not major changes that effect large part of the document according to a schema.

To merge such major mixed change sets makes tracking changes impossible

And it seems unfair to expect us to split your PR into multiple PRs.

What you can do to make it simpler:
Work with a set of branches and switch between branches depending on which change you want to make and commit that change into the appropriate branch. Once you are through, each branch can turn into a PR. This seems to work well, I can show you how I do this with command-line-gvim or how TorstenS taught me how to do this with fork/atom.

clagms:
Thank you for the clarification @andreas-junghanns.

I've put the arguments/parameters in a separate commit. 
From now on, stuff that's outside the scope of glossary i'll put in a separate branch.
Right now I'm unable to do this for the arguments/parameters commit though. 
Also, it kind of falls under the umbrella of glossary :p
Is it ok if that stays in this PR?

Can you have a look at the minor issues I'm adding to the PR comment?

andreas-junghanns:
@clagms: Please mark this as ready for review. I just reviewed because you asked and I did not check the status - but ideally, we check a PR only once at the end to reduce iterations.

clagms:
@andreas-junghanns, I'm nowhere near the end of this PR 😆 
I'm still in section 2. 
And I'm just identifying terms that should be defined (not defining them yet).
Do you want me to submit the PR just for section 2?
I can break it section by section (with some possible exceptions).

Overall, I think I need a week of work to get a reasonably complete glossary...

andreas-junghanns:
> 
> 
> @andreas-junghanns, I'm nowhere near the end of this PR 😆
> I'm still in section 2.
> And I'm just identifying terms that should be defined (not defining them yet).
> Do you want me to submit the PR just for section 2?
> I can break it section by section (with some possible exceptions).
> 
> Overall, I think I need a week of work to get a reasonably complete glossary...

If you need a week then it is even more important to break it into parts, either by topic (preferred!) or by section - otherwise all other work on the text is essentially stopped.

clagms:
Sounds good @andreas-junghanns. 
I'll break this PR into sections then.
Let me just finish section 2 (this should take today and tomorrow at least, as that is a big section).
I won't define the terms yet.

>otherwise all other work on the text is essentially stopped.

Can you explain why all the work in the text is stopped?

Maybe I misunderstood, but if a PR is merged before the current PR, all I have to do is pull all changes from the latest master in the upstream, merge them on my side, and push them to the glossary branch.
Right now, this is what I'm doing every time I get to work on this branch:
```
(current branch is clagms/fmi-standard/glossary)
git fetch -v upstream   (gets changes from modelica/fmi-standard)
git merge -v upstream/master    (merges modelica/fmi-standard/master into current branch)
git pull  (gets extra commits made by other people into this PR)
(do changes, commit, etc.)
git push origin
```

I've always assumed that it is the responsibility of the PR to pull all changes from the repo before a merge is made, so that there are no conflicts...

andreas-junghanns:
> Can you explain why all the work in the text is stopped?

If your PR touches every 4th line then when I merge my PR or your PR, we have a high chance of merge conflicts. Merge conflicts are no fun to resolve. Merge conflicts are hard to resolve and resolving them costs time and might introduce regressions or errors. I have had merges with 25 merge conflicts and that can easily take hours to figure out and resolve (and no merge tool can really help!). I do not want to do such work that is totally unproductive. Seeing monster PRs in the works, I stop my work waiting for the PR to be merged to then work on my stuff again.

Resolutions: Many smaller commits are easier to merge because we can merge them more often. If they are sorted by topic, resolving merge conflicts is MUCH easier because you usually understand more quickly how to resolve such conflicts.

This is modern collaboration 101 :P

clagms:
This is a good point @pmai!

I checked this when replaced the terms (see the glossary term "parameter" for a note I left).

To double check this, I went over the uses of "parameter" in the doc, and the only places where I think a reader might confuse the two are the following:
- `derivatives of the model with respect to parameters during continuous integration are not included`
- `...or the initial values are computed internally as a function of the parameters`
- `Additional parameters for establishing the network communication (for example, identification of the remote computer, port numbers, user account) are to be set via the GUI of the master.`

In many places (but not all), we already include a reference to the definition of a parameter (something like <<parameter>>) when referring to the parameters of the FMU. This helps avoid ambiguity.

Should we revert the change, or just clarify the above items?

andreas-junghanns:
Pierre was clear, I think: We simply use two words wherever we mean a "function parameter". This makes the standard easily readable.

Linking to a definition would require jumping around in the document needlessly.

clagms:
>Pierre was clear, I think: We simply use two words wherever we mean a "function parameter".

@andreas-junghanns, do you mean that we should replace "parameter" by "function parameter" everywhere? This term is used a lot on lists.
For example, applying this transformation leads to many lists like the following:
image

To reduce cluter, we can remove the term "parameter" from these lists altogether, as long as immediately before the list there's a sentence that makes it clear that each following item refers to the parameters of the function. As far as I saw, this is already clear.

@andreas-junghanns, @pmai, what do you think about replacing "parameter" by "function parameter", but first removing the term "parameter" where it is obvious that it is not needed?

andreas-junghanns:
(The list shows that your replacement of all arguments with parameter missed some, see visible and loggingOn.)

Clarity comes at a price: more words (when we cannot find a specific term). I am not sure how to solve the problem you pointed out that we name two different things parameter unless we name at least one with an additional term. If we can drop the term (as you suggest), means that it must be clear which sense of the term is meant - then why be more specific. Still, I like your idea to remove words, as long as it is clear what is meant.

clagms:
>(The list shows that your replacement of all arguments with parameter missed some, see visible and loggingOn.)

That was intentional @andreas-junghanns , because we refer to the value of `visible` and `loggingOn` in a boolean expression. Maybe I misunderstood the terms argument and parameter? They are all too easy to mix :p

andreas-junghanns:
> 
> 
> > (The list shows that your replacement of all arguments with parameter missed some, see visible and loggingOn.)
> 
> That was intentional @andreas-junghanns , because we refer to the value of `visible` and `loggingOn` in a boolean expression. Maybe I misunderstood the terms argument and parameter? They are all too easy to mix :p

I understand and I would try to optimize readability. If we use both, argument and parameter, because we are sophisticated and know-it-all, this won´t help the reader. I agree that parameter is the better term according to comp-sci traditions. Can we either have only parameter, or drop it in the lists that only talk about parameters (as you propose with an intro sentence)?

clagms:
I'll drop it from the lists, and change the other "argument" to "parameter".

clagms:
Come to think of it, if the distinction between argument and parameter is not important for us, then I think that actually always using the term "argument" to refer to a "function parameter" is much better than any other proposal so far. This makes it clear that "parameter" refers to model parameters, and "argument" refers to C stuff.
Should I make this change instead @pmai and @andreas-junghanns?

andreas-junghanns:
> Come to think of it, if the distinction between argument and parameter is not important for us, then I think that actually always using the term "argument" to refer to a "function parameter" is much better than any other proposal so far. This makes it clear that "parameter" refers to model parameters, and "argument" refers to C stuff.
> Should I make this change instead @pmai and @andreas-junghanns?

What you are proposing is how it was before you changed it (for other reasons, admittedly!).
So you could undo your changes in those lines and then we have the "arguments" back.

andreas-junghanns:
... come to think of it: we should explain why we name parameters arguments in the glossary, right?

clagms:
>What you are proposing is how it was before you changed it (for other reasons, admittedly!).

Happens so many times 😄

clagms:
Fixed with https://github.com/modelica/fmi-standard/pull/895/commits/db753c9a85097a28fbf9d8f6f1a93ca9af67e0ef. 
Thank you for the discussion!

clagms:
@andreas-junghanns , please go ahead and review this PR. I'm done with minor fixes.

andreas-junghanns:
@t-sommer: Will you have a look? We need to merge this Monday afternoon latest.

Ideally earlier, because I would like to start a number of additional PRs but I want to wait until this is merged.

If you cannot find the time, please let me know, I will then merge regardless: I am confident that this is a big step forward and I double checked every change and improved on top.

Issue #902, opened 2020-04-17
Examine seperation of source/build configuration into seperate XML file and backport 2.0


pmai:
As discussed today, 2020-04-17, (see https://github.com/modelica/fmi-design/tree/master/Meetings/2020-04-17-Regular-FMI-Design-Webmweeting#separation-of-build-configuration-from-modeldescription-xml)  it might be a good idea to split out the new source/build configuration into a seperate XML file:

* This might allow backporting of this enhanced configuration into FMI 2.0.x
* Would allow seperate evolution of build configuration spec from FMI version

This is part of a larger effort to see which features for FMI 3.0 can be split out into back-portable and seperately evolvable features. Similar discussion is ongoing for Ports & Icons, and other stuff might be identified as well.

Basic idea would be to have the current configuration sections of the modelDescription.xml as a seperate file, maybe under sources or intra/.../ or so, which can still reference the same source files as are in the FMI 2.0 config (or might be seperate ones). In this way FMI 2.0.x FMUs can support both configurations, where possible, enabling better compilation results/more platform support for tools supporting the new configuration mechanisms, with the old FMI 2.0 way as a fall-back.

Issue #870, opened 2020-03-28
Store "Ports & Icons" information in a separate XML


t-sommer:
As discussed at the Regular Design Meeting on 2020-03-28, it makes sense to move the "Ports & Icons" (P&I) information to a separate XML document that can be used and versioned independently of FMI 3.0.

Pros:

- P&I does not require changes to the API (unlike the other features introduced in FMI 3.0)
- the development would not held be back by the (slow) release cycle of FMI
- P&I can be retro-fitted to FMI 2.0 (or even FMI 1.0)
- the complexity of FMI 3.0 can be reduced

Cons:

- (please add points here)

**UPDATE**: I've changed the title from ~Continue "Ports & Icons" as a layered standard~ because it was misleading. This issue is _not about removing_ P&I from FMI but keeping it separate from the FMI 3.0 schema, so it can be used elsewhere (e.g. in FMI 2.0.2).

chrbertsch:
One could define a  new xml file with the "ports and icons" content and place it in a defined subdirectory of the /extra directory of an 3.0 FMU.

Additional Pro: 
- we would have blueprint for the usage of the /extra directory and the designs of layered standard, which would also promote these new concepts.

The goal would be to release this layered standard together with FMI 3.0 (or even earlier for FMI 2.0, the only requirement  for this would be 2.0.2 release that we could release almost immediately)
Thus for exporting and importing tools and end users, there is practically no difference compared to "ports and icons" being optional features of the "core" FMI 3.0 standard.
If exporting tools consider this feature important (which I do!) , they will ship their first 3.0 FMUs with the "Ports and Icons" content in the /extra directory.
 
The only parts that I would leave in the core FMI standard are the diagram.png, diagram.svg, icon.png and icon.svg files as defined in https://fmi-standard.org/docs/3.0-dev/#fmu-distribution

If there are enough people within the FMI project committing to contribute to this layered standard, we could  continue its development within the FMI project within a working group. (An alternative would be to create a new Modelica Association Project, but perhaps this will create too much overhead)

chrbertsch:
Regular Design Meeting: 

We shall put this on the agenda for next weeks FMI regular design meeting.

Torsten S: We should consider enabling this for FMI 3.0
Christian S: This should be part of FMI 3.0
Torsten S: we could make a it a layerd standard and state in FMI 3.0 that this layerd standard must be supported. It would be good to have a separate document
Christian S: We could have to have a separate xml. 
[...]

PR #904, opened 2020-04-18
Split of Terminals and Icons into new file

pmai:
This PR is a demonstration PR for how the split of the terminal and icons information into a separate file that would also work for FMI 2.0 could be handled (see #870). It is a draft PR until such a proposal is accepted in principle, and all necessary changes have been added to the PR to move it into shape.

Based on this foundation, the build configuration information could also be moved into a separate file (see #902).


t-sommer:
@pmai, did you update / re-generate the XSD diagrams?

pmai:
@t-sommer, not yet, and the documentation changes are not complete yet either, I’m waiting on feedback for the general approach and on whether to merge the terminal graphical info as I did, and/or to regularise the graphic info stuff some more...

pmai:
I think this PR is the place to develop the separate file and its semantics, as well as stuff like do we want to have a GUID/instantiationToken/something else reference in there to ensure stuff is in sync.

Once the technical details have been ironed out, and the overall approach to separating stuff is in place and agreed to, we'd do the move to separate documents/repos/....

But first I'd like to see everyone agree on the technical issues, the FMI 2.0 compatibility, etc.

t-sommer:
Wouldn't it be sufficient to give the XML file its own version number? For FMI 2.0.x we could reference the description and XSDs in the respecitve 3.x release.

andreas-junghanns:
> 
> 
> Wouldn't it be sufficient to give the XML file its own version number? For FMI 2.0.x we could reference the description and XSDs in the respecitve 3.x release.

Mmmhh, I thought we better not give version numbers for the XSD to decouple the versions.
I think we should search for a why to just say: There is the place where more information can be, how to interpret it depends on (according to Pierre´s proposal) the standard identified by the reverse notation domain.

pmai:
My comment on FMI 2.0 compatibility was more geared towards playing around with how the terminals and icons actually work with a 2.0 FMU (which I think they just do, but I think we'd better check more deeply).

CSchulzeTLK:
A few questions / remarks...

1. *schema/fmi3Terminal.xsd*:
The larger group had decided to separate the graphical representation completely from the interface description. The rationale to keep them separate was to enable simple tools to provide a graphical representation - even if they don't care about terminals. I don't see a big issue, because the fmu icon is still separate. Re-merging the graphical representation of the terminal with the fmi3Terminal element would enable graphical representation of nested terminals (=> requires at least some changes in the text).
2. *schema/fmi3TerminalsAndIcons.xsd*:
Do we need the *instantiationToken* here? This would only be used by FMI. But from my understanding you also target other use cases.

Why do we need the changes in *schema/fmi3Annotation.xsd* and *schema/fmi3InterfaceType.xsd*?

pmai:
> A few questions / remarks...
> 
> schema/fmi3Terminal.xsd:
> The larger group had decided to separate the graphical representation completely from the interface description. The rationale to keep them separate was to enable simple tools to provide a graphical representation - even if they don't care about terminals. I don't see a big issue, because the fmu icon is still separate. Re-merging the graphical representation of the terminal with the fmi3Terminal element would enable graphical representation of nested terminals (=> requires at least some changes in the text).

That is one additional change I made and requested feedback on, since for a stand-alone format it kind of made more sense to me to keep the terminal graphical representation with the terminals (which as you say also allows nested graphical representations, and avoids having to duplicate the terminal structure in two places; since tools only caring about the icon can completely ignore the terminals still, this seemed to me to better match the intent). This however is not necessarily tied to this PR, so we can leave that change out as well.

> schema/fmi3TerminalsAndIcons.xsd:
> Do we need the instantiationToken here? This would only be used by FMI. But from my understanding you also target other use cases.

That's another question I would have to the group: I added the instantiateToken to have some measure of checking whether the terminals file is in-sync with the modelDescription.xml file; however instantiateToken is not really very good for this task, and it indeed would require additional acrobatics to work with other component formats (however those are likely needed for the other aspects of it as well, so not that much of a problem). We can leave it out, or we can see whether some other criterion should serve as a redundant in-sync check.

> Why do we need the changes in schema/fmi3Annotation.xsd and schema/fmi3InterfaceType.xsd?

Those are needed/suggested so that we do not have to include fmi3InterfaceType into the new top-level file schemas, which don't need the FMI interface type stuff, but want to re-use the annotation element. If we go further in terms of the layered standard approach stuff, it would maybe make sense to adopt the SSP "common" approach of having some common core definitions as a seperable schema file (or even namespace).

schema/fmi3Terminal.xsd
```diff
@@ -53,7 +53,7 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 				</xs:complexType>
 			</xs:element>
 			<xs:element name="Terminal" type="fmi3Terminal" minOccurs="0" maxOccurs="unbounded"/>
-			<xs:element name="GraphicalRepresentation" minOccurs="0">
+			<xs:element name="GraphicalRepresentation" minOccurs="0" maxOccurs="1">
```
pmai, line 56: This is not necessary, since maxOccurs=“1” is the default...

pmai:
This seems more or less ready for inclusion; I think we might move the section on terminals into a separate 2.x section, as is being proposed for the build config PR, but I'd like to do this afterwards in a separate PR if we decide to do this, since this would produce unneccesary merge conflicts for now.

CSchulzeTLK:
The XSD image for the graphical representation of the FMU is not as expected. It includes both graphical representation elements (FMU and terminal), because they have the same name. The text is just about the FMI element.

https://github.com/modelica/fmi-standard/blob/d728db3aa59c77e14165a66f7b503d120ef19d81/docs/images/schema/GraphicalRepresentation.png

I failed to let the xsd tool generate just one of the elements. So we can change the text, the xsd2png tool, or the element names.

pmai:
> I failed to let the xsd tool generate just one of the elements. So we can change the text, the xsd2png tool, or the element names.

Yeah, that seems to be a restriction of the XSDDiagram functionality. I think we should ignore this for now, since the diagrams need fine-tuning for release anyway, and this can be fixed in extremis by cropping. I wouldn't change the schema nor the documentation for just tooling issues for now.

CSchulzeTLK:
I first searched for an argument to remove e.g. the 2. element in the graph (after automatically adding and sorting them). Maybe this functionality can be added to the tool.

However, renaming the elements seems to be less effort.
We could argue, that it is never a good idea to have the same element names for different things...

clagms:
I would also prefer that unique names are used, but the work involved in fixing the tool is clouding my judgement 😛

I understand that tooling issues are, in the grand scheme of things, ephemeral and the standard will outlast all those tools and all of us (and transcend it's physical form to become pure energy 😋).
However, having unique names would also: 
- simplify document references, 
- make it easier on the reader to understand which element we're referring to. 

We have duplicate names in other elements. 
For instance, Float64 can be a type or a value. 

While I wouldn't go so far as to ask that the "Float64" type becomes "Float64Type", I would ask that new elements be unique (for the above reasons).

Does this persuade you @pmai and @CSchulzeTLK ?

andreas-junghanns:
Let´s not discuss tool issues.
But I find it a bit strange that we have two elements in the XSD that have the same name.
I prefer all elements in the XSD to have unique names.
Or is there a compelling reason to stick to both elements having the same name?
(In my world, I would require all elements in an XSD to be unique, but that is just me, I guess.)

And by the way, Float64 is new, why not rename?

pmai:
XML/XSD is context sensitive, so it can actually express interesting things, even though too many people seem unable to deal with this (cf. all the broken XML tools and XML2OO bridges) ;).

I think the overall issue of whether we want to switch to unique element names in order to appease the gods of linearity should be handled in a new issue/PR.

andreas-junghanns:
> I think the overall issue of whether we want to switch to unique element names in order to appease the gods of linearity should be handled in a new issue/PR.

We will add this issue to our endless list of beer topics too because I think there is more than linearity at stake here (mainly sanity).

pmai:
@CSchulzeTLK can you provide examples and tie them into the text where they make sense? I'm only the mechanics of splitting it out guy, not a ports and icons connaisseur...

andreas-junghanns:
The GraphicalRepresentation.png contains two elements `GraphicalRepresentation` (just above 2.2.7.1.2. CoordinateSystem), the latter also shown in TerminalGraphicalRepresentation.png later in the document. Is this an export error?

Just below 2.2.7.1.2. CoordinateSystem, we repeat the CoordinateSystem element from above in CoordinateSystem.png - is this intended? It looks kind of odd.

Looks like we have not VendorAnnotations in all Annotation elements. My understanding of XSD is patchy at best, but why do we still have a VendorAnnotations element in lines 39-41?

I see more VendorAnnotation elements appear elsewhere? Is this a secret incursion of VendorAnnotations by the SSP project?

No other observations right now.

pmai:
> The GraphicalRepresentation.png contains two elements GraphicalRepresentation (just above 2.2.7.1.2. CoordinateSystem), the latter also shown in TerminalGraphicalRepresentation.png later in the document. Is this an export error?

yesno, this is the tool limitation issue that was discussed above (the XSDdiagramm tool has no syntax to specify elements other than with a name, which of course is not unique; for other cases we could work around this because the duplicate element names occur in separate XSD files which are not mutually included, but that is not true for this case).

> Just below 2.2.7.1.2. CoordinateSystem, we repeat the CoordinateSystem element from above in CoordinateSystem.png - is this intended? It looks kind of odd.

This seemed good for uniformity (cf. build configuration), one might want to tweak this by rearranging things...

> Looks like we have not VendorAnnotations in all Annotation elements. My understanding of XSD is patchy at best, but why do we still have a VendorAnnotations element in lines 39-41?

??? FMI still sticks to the VendorAnnotation moniker, SSP uses just plain Annotations (which I find more pleasing, but ignore this for now).

I tried to follow the modelDescription way of doing things, which has one top-level VA, and one in each modelvariable (cf. terminals).

One could have more or fewer than these, but that probably needs to be done across the standard.

> I see more VendorAnnotation elements appear elsewhere? Is this a secret incursion of VendorAnnotations by the SSP project?
> No other observations right now.

We don't know any VendorAnnotations ;)

andreas-junghanns:
Pierre: Will you merge into master?

pmai:
> terminalsAndIcons.xml is now inside /icons. I am happy (for now).

Does this really make sense? Anyway the documentation now says icons, however the folder is called icon, and the text needs some adjustments. And the build spec breakage needs resolving.

I'll merge once this is sorted out.

andreas-junghanns:
Aha - I was not aware that we are not enforcing icons to go under /icon(s)!
Should we not require this? Otherwise icons will end up everywhere, e.g. in /resources.

pmai:
> Should we not require this? Otherwise icons will end up everywhere, e.g. in /resources.

I think it is required, in the sense that you can only use relative URIs (relative to icon/terminalsAndIcons.xml) that are not allowed to go up (i.e. ..), thus all icons must reside under icon (that was true formerly as well, it's just that we can now formulate this more baseURI-correctly, when terminalsAndIcons.xml also resides in icon).

andreas-junghanns:
So, icon or icons?

pmai:
> So, icon or icons?

I left it at icon, since that's what it was in the rest of the document. If we want to rename the directories, this should be done in separate PRs.

PR #905, opened 2020-04-18
Move build configuration to separate XML file

t-sommer:
see #902


t-sommer:
@pmai, I'm not happy with the names of the XML and the root element, but couldn't come up with something better. Do you have ideas how to call them?

chrbertsch:
Generally I like the proposal very much.

But if we keep the description of the Build configuration in the FMI Standard (now as section 2.5), how can we reach the goal of "Would allow separate evolution of build configuration spec from FMI version"?

Wouldn't it be better to move the specification of the build configuration in a separate document and in in the FMI 3.0 standard document just state : "FMI 3.0.0" is released supporting FMI_Build_Configuration Version x.y.z?

andreas-junghanns:
> Wouldn't it be better to move the specification of the build configuration in a separate document and in in the FMI 3.0 standard document just state : "FMI 3.0.0" is released supporting FMI_Build_Configuration Version x.y.z?

I agree, we need a separate document (standard?) with its own version number, its own XSD (outside the current /schema directory. Maybe we should have separate repositories with these new sub-standards (FMI-SD and FMI-PI, for Source-Distribution and Ports-and-Icons, respectively).
This would be part one.

Part two is how we open up FMI to point to these (and other) new layered standards.
Right now we only have /extra, but that was not meant exactly for this purpose and I think this is sub-optimal. /extra is the folder we gave up control!
The /intra proposal sounds better, but even here we should define more clearly how this should work and where additional, future layered standards can be placed.

Proposal, based on the /intra idea:
---
Information according to layered standard, endorsed by the FMI project will be placed inside /intra.
The content of sub-directories inside /intra with name /intra/XYZ are defined in the layered XYZ standard and will contain an /intra/XYZ/XYZ.xml file to describe the content of /intra/XYZ and its semantics.
A version number inside /intra/XYZ/XYZ.xml is mandatory to allow for decoupled development of FMI and it endorsed layered standards.
The name space for XYZ is under control of the FMI project and a vote of the FMI Steering Committee is required to allow a standard to be placed inside the /intra/XYZ directory.
Such standards may be developed by organizations outside the Modelica Association.
The /extra directory is the place to add information of tools and according to standards not (yet) endorsed by the FMI project.
---

Open question: should we allow /intra content to be changeable by tools other than the exporter (i.e. should potential check sums include or exclude /extra)?

pmai:
> I agree, we need a separate document (standard?) with its own version number, its own XSD (outside the current /schema directory. Maybe we should have separate repositories with these new sub-standards (FMI-SD and FMI-PI, for Source-Distribution and Ports-and-Icons, respectively).

While I agree that this is the way to go (it being the way we did this for SSP), I think for now I would concentrate this PRs on demonstrating the technical parts, i.e. getting the seperate files to a level where we can see whether it would work correctly, and could be transported to FMI 2.0.

Once that's done and we have agreement on proceeding in this way, I'd go about splitting into separate documents, repos, etc.

> Right now we only have /extra, but that was not meant exactly for this purpose and I think this is sub-optimal. /extra is the folder we gave up control!

Yes and no: /extra is not without control, we reserved org.fmi-standard etc. prefixes in there exactly for this purpose; so I think stuff that can be changed from the outside could still go into /extra, even if it originates from FMI working groups.

> The /intra proposal sounds better, but even here we should define more clearly how this should work and where additional, future layered standards can be placed.

The distinction between /intra and /extra from my point of view should just be whether you are expected to be able to change its content from the outside or not.

I would not distinguish between layered standards done by the FMI project and those done from outside in the FMU directory structure above the distinction given through the reverse domain notation (i.e. FMI developed standards will have a prefix of org.fmi-standard, others not). Even standards developed by others but endorsed by FMI will be left under the original prefix for compatibility.

Or in other words: Origin and endorsement by FMI wg should have no changing impact on directory layout, since these are political/administrative acts and should not lead to code changes just because we do/don't do something in the administrative domain, IMHO.

I would also not mandate sub-structure or file-content in those layered standards, since that is under the purview of that layered standard (whether it even needs any top-level XML file or not will depend on the standard; whether we will stick forever to XML or not is not something we should cast in stone right now).

Or put it another way, I'd propose to take over the complete SSP way of doing this (which lead to the /extra directory in FMI), with the addition of having /intra for stuff where we don't support changes from the outside. The SSP standard contains a section that delineates how layered standards are supposed to work, and how they can be adopted into the SSP core without any changes.

andreas-junghanns:
Ok, keeping all in /extra is more lenient and I agree for technical reasons.
I am wondering about getting support from the P&I WG - will they accept being treated the same as anyone outside?
For the discussion Tue/Wed, we should have a complete sketch - also about the separation - or technical questions will remain and potentially forestall agreement to this proposal.
That would be sad, right?

pmai:
The SSP standard section 2.1 and 2.2 provide a detailed approach, e.g.

> Layered standards can fall into three categories:
> - Layered standards can be defined by third parties, making them third-party layered standards, without any representations by MAP SSP for their suitability or content, or even knowledge by MAP SSP about their existence.
> - Layered standards can be defined/adopted and published by MAP SSP itself, making them MAP SSP layered standards.
> - Finally, layered standards that have achieved enough adoption or importance to be included into the base standard set can be incorporated into a new minor release version of the base standard as an optional or mandatory appendix, making support for this layered standard optional or required for conformance with the newly published minor release version of the base standard.

The last para should be clarified to mean that support for that version (or newer backward-compatible ones) is required by that SSP/FMI version.

Stuff developed by the FMI working group (at least initially) would reside under org.fmi-standard, e.g. the ports and icons stuff could be org.fmi-standard.terminals-and-icons, etc.

It would be versioned and developed separately. If we were to include it into the base standard as per category 3 above, FMI a.b would "require" optional support for org.fmi-standard.terminals-and-icons version x.y or better.

otronarp:
> A version number inside /intra/XYZ/XYZ.xml is mandatory to allow for decoupled development of FMI and it endorsed layered standards.

Shouldn't the directory be versioned as well so that you can export an FMU that supports both 1.0 and 2.0 of the layered standard?

Like `/intra/XYZ/1.0/XYZ.xml` and `/intra/XYZ/2.0/XYZ.xml`

Otherwise with more parts of the standard forked of to layered standards living their own life it's going to be increasingly difficult to use an FMU unless it can contain support for multiple versions of layered standards.

pmai:
For that reason SSP left the versioning semantics and sub-directory structure to the layered standard, so that standards that want to support multiple versions in parallel could do so (either via sub-directories, or e.g. by including the version number in the reverse domain notation prefix, e.g. org.fmi-standard.terminals-and-icons.v1). Since different layered standards might have different needs here, we left it to the layered standards.

Alternatively for FMI we could however also mandate semantic versioning and include the major release in the mandatory prefix...

IZacharias:
I like the idea to describe the BuildConfiguration in a separate XML/XSD. But it must still be part of FMI 3.0, i.e. described in the FMI 3.0 standard. Removing it from the base standard would mean removing the source code support from the standard. 
Regardless of whether the feature is described in FMI 3.0 in a separate document or not, it is not a big effort to include the XSD and the description into FMI 2.1 in a backwards compatible way.

The BuildConfiguration feature has little potential to be subject to frequent changes and extensions. It is sufficient to develop and release it together with the base FMI Standard. This also gives a clearer picture for FMU importers/exporters and for users. Especially for customers it is unnecessarily confusing if they have to deal with several feature version information, i.e. the customers need to know about an FMU 3.0, which versions of which features it contains, in order to find out whether this FMU is supported in an importing tool.

t-sommer:
@pmai, can you please push the changes you proposed regarding the `<VendorAnnotations>` elements to this branch?

pmai:
This adds the minimal set of VendorAnnotations. One could add them to all the lower elements as well, but this is more a general issue for FMI and should be discussed/handled in a separate PR.

andreas-junghanns:
The terminalsAndIcons.xml is proposed in #904 to be on top level. Why is buildDescription.xml inside /sources? Or, why not move terminalsAndIcons.xml one level down to /icons?

No other observations for now.

pmai:
> The terminalsAndIcons.xml is proposed in #904 to be on top level. Why is buildDescription.xml inside /sources? Or, why not move terminalsAndIcons.xml one level down to /icons? 

It seems to me that sources seems like a good place to put the buildDescription.xml, since it is purely related to building sources. For terminalsAndIcons I don't see this as much, since most of it is not actually related to icons (one could argue that it should be split into two files, but that seems to be overdoing it).

So I think there are decent reasons for this.

PR #907, opened 2020-04-20
Fmi3 restructure first chapters

andreas-junghanns:
Please have a look at both commits separately:
 - the first is some minor moving around,
 - the second removes one entire file and move the acknowledgements to the appendix


andreas-junghanns:
I am happier with the overall flow of the (now) first chapter. A lot of redundancy is gone. Clarity as to what belongs where is clear at least in my head now - please confirm that this makes sense to more than just me.

t-sommer:
I think this is going in the right direction. However, when we're touching the introduction, we should also clean up the text. I would also keep some kind of abstract / preface before chapter 1 (similar to the [Modelica Language Spec](https://www.modelica.org/documents/ModelicaSpec34.pdf)).

PR #909, opened 2020-04-20
Remove duplicate state ref

andreas-junghanns:
Simple cleanup: some xrefs already contain the "Section".

Is this ok to be capitalized? All other references are lower case. Do we have to adapt the formatting of references to sections? How?


andreas-junghanns:
I kept this as draft  because I wanted to also fix the capitalized Section here, but I now think these are separate issues. Let´s merge this and later work on making the Section lower case.

PR #919, opened 2020-04-21
All sequence related changes

CSchulzeTLK:
memberName is optional for sequence but not for plug and bus,
remove "name/sequence-based",
fix alias handling


CSchulzeTLK:
@TorstenBlochwitz : I think you pointed out, that the variable type must match for matchingRule "sequence". However, this was explicitly not restricted:

> Matching is not restricted by variability, causality or variable type. Example: A fixed variable may be connected to a tunable variable, a variable of type fmi3Float64 may be connected to a variable of type fmi3Int32. However, it is recommended that the variable types and variabilities are equal.

Issue #921, opened 2020-04-21
Input Approximation


chrbertsch:
Today in the FMI Design meeting we agreed how to proceed with input approximation

see https://github.com/modelica/fmi-design/blob/master/Meetings/2020-04-21-22-Online-FMI-Design-Meeting/minutes.adoc#input-interpolation based on discussion https://github.com/modelica/fmi-standard/issues/794#issuecomment-617184280. 

Superseding tickets #603, #622, #623 (comment) 

Summary of decisions: 

- Input derivatives are coefficients of a specified polynomial that the FMU has to use on the current communication interval. We use the taylor polynomial similar to FMI 1.0 appendix D. 
- Additionally the FMU has an attribute on the maximum order of input deriviatives, replacing the attribute can interpolate inputs.
- The maximum order of input approximation shall be defined per FMU (and not per input and not generally in the standard.
- the capability flag CanInterpolateInputs shall be replaced by something like InputApproximationOrder

@friedrichatgc  and @chrbertsch will work on a PR

chrbertsch:
Closed according to decision at Online FMI Design meeting, see https://github.com/modelica/fmi-standard/pull/923#issuecomment-617825867

InputApproximuation API funcion shall be removed from FMI 3.0; the functionality can be realized  with Intermediate Variable Access and `fmi3GetOutputDerivatives`, see #934

Issue #836, opened 2020-03-03
Shall BCS support Invermediate Variable access and Early return?


chrbertsch:
Or should such "advanced" co-simulation features be moved to HCS?

These questions were raised in comments to different tickets (#797 #820,  #821) , so I created this new ticket to avoid a discussion split over several tickets.

Arguments:

Pro:
- This is what we decided and what is on the  current master branch. Why change it again?
- These two features do not require EventMode (with all the consequences like direct feedthrough at events). So BCS with this feature is more advanced than FMI 2.0 CS, but less advanced than HCS.
- IVA can be used in BCS also only for plotting with more information than only communication points (no advanced co-simulation)
- Early return is a simple extension of FMI 2.0 CS without complex event handling

Contra:
- keep BCS simple

... Plesae extend the list (you can edit this texts or comment)

jbernalr:
> So, do we allow advanced features like early return in Basic Co-Simulation?

I would say, according to my memory of the discussions, no we dont. 
What we do is, that we allow Co-Simulation with Early Return capability to be used as Basic Co-Simulation as well. (At least this is the functionality I would like to have as a user)

|   | Basic Co-Sim **FMU** | Early Return Co-Sim **FMU** |
| ------------- | ------------- |------------- |
| Basic Co-Sim **Master**  | Yes  | Yes (as basic CS)  |
| Early Return Co-Sim **Master**  | Yes (as basic CS)  | Yes |

chrbertsch:
> What we do is, that we allow Co-Simulation with Early Return capability to be used as Basic Co-Simulation as well.

Which FMU type does a "Early Return Co-Sim FMU" have? BCS or HCS?

jbernalr:
> Which FMU type does a "Early Return Co-Sim FMU" have? BCS or HCS?

Not a BCS. 

There may be however more than one of the remaining types supporting/using early return

chrbertsch:
> > Which FMU type does a "Early Return Co-Sim FMU" have? BCS or HCS?
> 
> Not a BCS.
> 
> There may be however more than one of the remaining types supporting/using early return

Which ones? ME and SCS do not have a doStep from which to return early, right?

APillekeit:
The state of the discussion up to now was: Intermediate Variable Access is available for all Co-Simulation modes since it is a realtivly orthogonal to the simulation interface types. It can be turned on and off in each co-simulation interface type and the fundamental way of simulating the FMU would not change, only additional information can be transported.

[...]

PR #923, opened 2020-04-21
Input approximation

chrbertsch:
addressing #921

-  Input derivatives are coefficients of a specified polynomial that the FMU has to use on the current communication interval. We use the taylor polynomial similar to FMI 1.0 appendix D.
-  Additionally the FMU has an attribute on the maximum order of input deriviatives, replacing the attribute can interpolate inputs.
-  The maximum order of input approximation shall be defined per FMU (and not per input and not generally in the standard.
-  the capability flag CanInterpolateInputs shall be replaced by something like InputApproximationOrder

### DIFF
--- a/docs/4_2_basic_co-simulation_api.adoc
+++ b/docs/4_2_basic_co-simulation_api.adoc
@@ -12,7 +12,7 @@ This section contains the interface description to access the input/output data 
 ===== Input Approximation [[api-input-approximation]]
 
 In order to enable the slave to approximate the <<continuous>> floating point <<input,`inputs`>> between communication steps, the the coefficients of a polynomial can be provided.
-The order of this approximation is given by the unsigned integer capability flag `MaxInputApproximationOrder`.
+The order of this approximation is given by the unsigned integer capability flag `inputApproximationOrder`.
 
 [[fmi3SetInputApproximation,`fmi3SetInputApproximation`]]
 [source, C]
@@ -34,21 +34,19 @@ include::../headers/fmi3FunctionTypes.h[tags=SetInputApproximation]
 
 <<fmi3SetInputApproximation>>::
 Restrictions on using the function are the same as for the `fmi3Set{VariableType}` function.
-Different input variables may have different interpolation order.
-<<input,`Inputs`>> and their <<derivative,`derivatives`>> are set with respect to the beginning of a communication time step.
 
 The input approximation polynomial on the interval latexmath:[[tc_i, tc_{i+1}]] is defined to be
 
 latexmath:[u(t)= \sum_{j=0}^{k} \frac{1}{j!} u_j (t-tc_i)^j]
 
-where k is the `MaxInputApproximationOrder`.
+where k is the `inputApproximationOrder`.
 
-It is up to the master to define the cofficients latexmath:[u_1] to latexmath:[u_k], while latexmath:[u_0=u_{tc_i}], which is set with the `fmi3SetReal64`.
+It is up to the master to calculate the cofficients latexmath:[u_0] to latexmath:[u_k]. latexmath:[u_1] to latexmath:[u_k] is set using <<fmi3SetInputApproximation>>, while latexmath:[u_0=u_{tc_i}] is set with <<fmi3SetReal64>>.
 
 _[One choice of the master could be to define them to reach a certain value for latexmath:[u_{tc_{i+1}}] with a linear interpolation._ +
 _Another choice of the master could be to use the output derivatives of an connected FMU at time latexmath:[tc_i]:_ +
-_Then the polynomial corresponds to a Taylor polynomial._ +
-_In order to get the output derivatives of an connected FMU, the master can invoke the following function]_
+_Then the polynomial corresponds to a Taylor approximation of the output of the connected FMU at time latexmath:[u_{tc_i}]._ +
+_In order to get the output derivatives of an connected FMU, the master can invoke the following function:]_
 
 [[fmi3GetOutputDerivatives,`fmi3GetOutputDerivatives`]]
 [source, C]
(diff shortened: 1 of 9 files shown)


chrbertsch:
Discussion (@friedrichatgc , @chrbertsch ):

Disccussed and fixed issues: 
- remove the orders argument of fmi3SetInputApproximation and a require setting of all orders 1...maxInputApproximationOrder order
- remove max  from maxInputApproximationOrder  and rename to InputApproximationOrder
- define memory layout of values array for array variables and order
- Can continuous input variables be float32? Do we then have to support Input approximation for 32bit. --> Does not make sense, as states can be only float64

Still open:
- does this feature make sense for SCS? If yes --> move description from BCS to "Common Concepts for Co-Simulation" 

@friedrichatgc : perhaps we should also have a scalar `order ` argument for `fmi3SetInputApproximation`:
- reduce data exchange
- symmetry to fmi3getOutputDerivatives
We would then have to define that all higher coefficients used by the FMU for interpolation are set to zero.  What's your opinion?

chrbertsch:
> @friedrichatgc : perhaps we should also have a scalar `order ` argument for `fmi3SetInputApproximation`:
> 
>     * reduce data exchange
> 
>     * symmetry to fmi3getOutputDerivatives
>     We would then have to define that all higher coefficients used by the FMU for interpolation are set to zero.  What's your opinion?

We could change again the flag inputApproximationOrder again to maxInputApproximationOrder

We then could add an `order` argument to  `fmi3SetInputApproximation`

```
typedef fmi3StatusTYPE(fmi3Instance instance,
                                               const fmi3ValueReference valueReferences[],
                                               size_t nValueReferences,
                                               float32 order, 
                                               const fmi3Float64 values[],
                                               size_t nValues);
```

with order having the meaning, up to which meaning *up to which order* the approximation coefficients are provided, with `order `<= `maxInputApproximation` Order. 
It is additionally mandatory to set all higher coefficients of the input interpolation, for which the `fmi3SetInputApproximation` had been called in the current timestep, to zero.

chrbertsch:
Discussion at the design meeting: 
What happens, if the master does not call fmi3inputApproximation? Or not on all value references?
If you do not call it, use zero. (at the beginning of each communication step, all the coefficients a_j are assumed to be zero.
Introduce an argument `order `to the call fmi3SetInputApproximation.
Is it allowed to call the function multiple times? yes.

Make sure, the the coefficients are only valid for the next interval.

For Hybrid co-simulation: what to do in case of events?

With IVA one could do the same, but not the other way round.
Andreas: Why keep this feature then?
Markus: For smooth transition from FMI 2.0. 
Claudio: can this not conflict with IVA?
Markus: this cannot be used jointly.

Torsten: Do we have a flag for eachInput, if setting intermdiate input values?
Andreas: Yes. intermediateAccess

Define a rule, that an importer may not use both mechanisms of intermediateVariableAccess and intermediateVariableAccess

Markus: We could also remove inputApproximation. the master could realize this. This creates a little bit more effort for the master, but less effort for the FMU. Perhaps its even simpler for the master.

Otto: in a distributed setup, passing the polynomial to the FMU could be more efficient.

Markus: one could pass the coefficients to the slave

Torsten: Master and FMU must support Intermediate update callbacks.

Klaus: in FMI 2.0 it was not well specified

Christian: Should we have a poll whether we remove inputApproximation?

There was a poll already

Christian: I think it we clarified later, that IVA is supported in BCS.

Klaus: Could not the solver in the FMU do something intelligent with the polynomial?

Markus: if you you have BDF solver, you can just provide a new value. This can be done with IVA.

Torsten B.: this leads to a problem: Can one call it with preliminary values?

Klaus: there is a flag.

Torsten: with IVA you have to call the function many times. Does your solver allow this?

Markus: the only case would be if the solver allows for exactly the same polynomial.

Klaus: We change responsibility to the master.

Claudio: you can decouple the calls ... 

Klaus: the master can pass values when it likes to

Markus: Regarding performance: the polynial could be implemented w/o a function call.

Klaus: the set and gets are more costly

We had a poll on removing inputDerivative #836, but with few people and some people have changed their mind

Claudio: If one needs a real input derivative, one has to declare a special input

Torsten: in #836 we had the argument that keeping the inputApproximation gives a smooth transition from FMI 2.0

Klaus: in FMI 2.0 it was not clearly defined, so we could remove it. We had big problems in FMI 2.0.

*Poll*:  
*Should we remove inputApproximation from FMI 3.0? (and use IVA instead)?*

Yes: Andreas, Christian B, Antoine, Claudio, Klaus, Markus, Otto
No: Torsten B. 
Abstain: Karl, Irina, 

--> Decision: remove  inputApproximation

chrbertsch:
Declined according to discussion above.
InputApproximuation API funcion shall be removed from FMI 3.0; the functionality can be realized  with Intermediate Variable Access and `fmi3GetOutputDerivatives`, see #934

docs/4_2_basic_co-simulation_api.adoc
```diff
@@ -6,49 +6,77 @@ This section contains the interface description to access the input/output data
...
+--
+* `valueReferences` is a vector of value references that define the variables whose input approximation coefficients shall be set.
+
+* `nValueReferences` is the dimension of the argument `valueReferences`
+
+* `values` is a vector with the values of the approximation coefficients. The or of the `values` elements is derived from a threefold serialization: the outer level corresponds to the index of the coefficient, the middle level to the `valueReferences`, the inner level to the serialization of variables. The inner level does not exist for scalar variables.
+
+_[ Example:_ +
+_Assuming we have inputs u1[2*3],  u2,  u3[2] and `inputApproximationOrder`=2._ +
+_Then the `values` are serialized in the following way:_
```
chrbertsch, line 32: Creating a comment just to be able to link this line

Issue #122, opened 2018-10-16
add reallocMemory to callback functions


modelica-trac-importer:
**Reported by andreas&#x2e;junghanns on 4 Jan 2013 16:14 UTC**
Currently, the callbacks specified for memory management are allocate and free.

If FMU code uses realloc() then this cannot be faked by malloc-memcpy-free because the previous size of the memory is missing for the memcpy action.

Suggestion: Replace allocateMemory with reallocateMemory and accept a first argument that can be NULL.

iakov.nakhimovski@modelon.com requested such a function as well in his email from Feb 16, 2012.

----

Migrated-From: https://trac.fmi-standard.org/ticket/122

modelica-trac-importer:
**Comment by iakovn on 8 Jan 2013 22:59 UTC**
I got a negative response on my mail mentioned in the ticket and investigated the issue a bit deeper. There are some strong reasons why `realloc` is prohibited in many systems. For instance, Steve Maguire's Writing Solid Code explains why `realloc` is horrible. 
> The real problem is it does so much, and any combination of parameters is valid; you can pass zero and non-zero in any combination.
>
> It does: malloc, free, shrink and relocate, where shrink would make an allocated block smaller and relocate would move to a new block.

His conclusion was that it was such a common source of error (because it is misunderstood) that it shouldn't be used directly. And I have to say, I agree.

There is no realloc in C++ neither (if one uses only new and delete). 
With this arguments I can understand why this callback in not present in the standard interface. Luckily, most numerical codes do not need `realloc` anyway.
If one needs to "fake" `realloc` then every `malloc` request may be extended with a header containing information on the memory block size before calling `allocateMemory`. This information may then be used to perform `memcpy` in the `realloc`.

To make a decision on the standard it might be worth too make a poll on which importing environments do not provide "native" realloc vs which FMUs need the functionality. Note that there is even a capability flag "canNotUseMemoryManagementFunctions".

modelica-trac-importer:
**Modified by rerbacher on 6 Feb 2013 11:07 UTC**

modelica-trac-importer:
**Comment by torstenblochwitz on 7 Feb 2013 14:55 UTC**
Decision: Follow the last remarks from Jakov: realloc will not be included.

modelica-trac-importer:
**Comment by karl&#x2e;wernersson on 26 Jun 2018 13:56 UTC**
Requested by array working group, should be reopened again for 3.0

[...]

PR #928, opened 2020-04-22
Remove memory management functions

t-sommer:
- fmi3CallbackAllocateMemory
- fmi3CallbackFreeMemory

and attribute "canNotUseMemoryManagementFunctions"

see #122


andreas-junghanns:
We should vote on this in the large group at 14:00 CEST, today.

PR #938, opened 2020-04-23
Clarify description of clocks

IZacharias:
- Add attribute relatedOutputClockReference to Clock variable (specification & schema)
- Clarify description of clockReference attribute
- Remove the double occurrence of input and output column from tables


andreas-junghanns:
I am trying to finally understand the clocks and stumble into issues.
E.g. "It is not allowed to include clocks of different clockType in one FMU."

What if we have multiple interface types inside one FMU, with ME and synchronous clocks and SCS with communication clocks? There is a sentence that says: "These clocks are ignored by Model Exchange." (referring to synchronous clocks) - but why have this if only one type is allowed?

andreas-junghanns:
Another issue: "Clocks can be periodic or non-periodic." Is non-periodic the same as aperiodic (see "The optional attribute relatedOutputClockReference is used to define a tick relationship from an outputClock to an aperiodic inputClock.")?

clagms:
2020-04-30-Regular-FMI-Design-Meeting Discussion:

TS: Please fix the build first, before merging.

andreas-junghanns:
The currently required attribute 'priority' will likely be for all clocks.
@KarlWernersson : Can you provide such priorities for Modelica clocks? Do they make sense there?

IZacharias:
> 2 issues:
> 
> add an example modelDescription.xml that validates against your xsd - we are wondering about the place of the priority attribute - should this not be inside the Clock element (and rather be defined in fmi3ClockAttributes)?
> the name of attribute relatedOutputClockReference seems long and should likely reference "union" somehow, if it defines a union. Better name might be triggeredBy.

- I will add an example XSD. 
- @andreas-junghanns I don't understand the question regarding the place of the priority attribute. The attribute is defined in fmi3ClockAttributes 
- I like the name triggeredBy. Another proposal: triggeringClockReference (at least three characters shorter than relatedOutputClockReference)

andreas-junghanns:
>     * @andreas-junghanns I don't understand the question regarding the place of the priority attribute. The attribute is defined in fmi3ClockAttributes

Agreed, this is ok (no idea, I must have been cross-eyed at the time).

andreas-junghanns:
`triggeredBy` is short and descriptive. To tell people that it is triggered by an output clock adds little but characters. I just made the (simple) change to see how the text reads and I like it quite a bit better.
Agreed?

@IZacharias : What keeps us from merging your improvements?

Issue #934, opened 2020-04-22
Remove inputInterpolation


chrbertsch:
According to decision at Online FMI Design meeting, see https://github.com/modelica/fmi-standard/pull/923#issuecomment-617825867

- Remove Attribute `canInterpolateInputs`
- Remove API function `fmi3SetInputDerivatives`
- Add short rationale and non normative text, how to realize Input approximation (similar to FMI 1.0 CS) with Intermediate Variable Access and `fmi3GetOutputDerivatives`
- Check and clarify arguments of `fmi3GetOutputDerivatives`

PR #939, opened 2020-04-23
Remove input derivatives

chrbertsch:
Remove InputInterpolation according to https://github.com/modelica/fmi-standard/issues/934


chrbertsch:
One open point with this PR is the `orders` argument.

I see two possibilities: 

1.  Leave it as in FMI 2.0, but describe its usage better. 
in order to inquire multiple input derivatives, one can pass a value reference multiple times to the `valueReferences` of `fmi3SetInputDerivatives`, and then put the the corresponding derivatives order numbers, at the corresponding place of of the `orders` array (e.g., to get first and second time derivatives of a variable with value reference `7`, define `valueReferences=[7, 7]`, `orders=[1,2]`. This has not been clear to me by reading the FMI 2.0 specification. Additionally, the serialization in case of array variables has to be explained. 
PRO: high flexibility, e.g. to inquire different time derivative orders for different variables, same as in FMI 2.0
CONTRA: Complicated in the standard case, where for each variable the same derivative orders are inquired: an array  needs to be created and passed, while in alternative 2. an integer suffices.

2. changing `orders` to a scalar variable`order` as in the previous, rejected PR https://github.com/modelica/fmi-standard/pull/923. Then all the time derivatives of the variables with the listed value reference up to `order`, see https://github.com/modelica/fmi-standard/pull/923/files#r415391371
PRO: Simpler in the standard case, where for all variables the same order of derivatives are inquired
CONTRA: Serialization of `values `array more complicated 

I slightly prefer 1.

@friedrichatgc , @antvl , @TorstenBlochwitz  and others:  could you please comment on which alternative you prefer?

clagms:
2020-04-30-Regular-FMI-Design-Meeting

CB: Which way we shall go for this?

KW: Have to think about this.

AJ: I won't use this API.

AJ: I'll go for option 1.

CG: I'll go for option 1.

Conclusion: CB will go for option 1.

chrbertsch:
@t-sommer : I do not understand the ci/circleci: build-errors. Could you please hava a look? Thanks!

andreas-junghanns:
The example for fmi3GetOutputDerivatives seems broken: y1[2*3] - what does that mean? y1 has 2 dimensions?

masoud-najafi:
Should the table 1.2.5 (Feature Overview of FMI Interface Types)  be updated and allow IVA for BCS?

chrbertsch:
Regular Design meeting: merge it.

PR #946, opened 2020-04-30
Harmonize format of attribute "version"

t-sommer:
in buildDescription.xml and portsAndIcons.xml

### DIFF
--- a/docs/examples/build_description_complex.xml
+++ b/docs/examples/build_description_complex.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<fmiBuildDescription version="1.0-alpha.4">
+<fmiBuildDescription version="3.0-alpha.4">
 
   <BuildConfiguration modelIdentifier="PlantModel" description="Build configuration for desktop platforms">
     <SourceFileSet language="C99">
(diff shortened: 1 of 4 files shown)


clagms:
2020-04-30-Regular-FMI-Design-Meeting:

PMAI: the questions here really is: what version number should they have? I used the same version number as the model description. TS things everything should start at 1.0. 

AJ: we decided to have these things going synchronous. Having independent version numbers is bad. Here we have to have the same v number as model description. We avoid tables of compatibility.

CB: I agree. 

PMAI: 3.0 is slightly less confusing.

TS: this attribute should be called fmi version if we want 3.0. 

PMAI: TS says, if we release a 3.0.1, all files get that version, even if we did not change anything on that part. 

AJ: Everything is baked into the same version, so this should be called fmiversion.

CB: Those xmls could evolved independently (to do bugfix).
What would you do if you use this XML file for a 2.0 fmu?

AJ: How would you use it? This number describes where this xml description is.

CG: How likely is this?

CB: Very likely for fmi version 2.0

PMAI: There are no clean solutions to this.

Irina, CG: I think fmiVersion makes sense.

CB: So the two versions are coupled. 

CB: Let's go with fmi version, that starts at 3.0.

TS: I'm cool with it.

**Conclusion**: Let's go with fmiVersion, starting at 3.0.

Issue #945, opened 2020-04-28
Change VendorAnnotations to Annotations?


chrbertsch:
In several tickets like https://github.com/modelica/fmi-standard/pull/904#issuecomment-620631343 it was mentioned, that  `VendorAnnotations `could/should be changed to `Annotations`. To me this sounds logical.

Pro: 

- Why should only "Vendors" created defined annotations for a certain validity scope and not some other organization, consortium, ....?
- This would be compatible with SSP (which also belongs to the "FMI"-family of standards
- ....

Contra: 

- Renaming late in the development process of FMI 3.0
...

Please add arguments

andreas-junghanns:
I think something along those lines happened in PR #905...

pmai:
> I think something along those lines happened in PR #905...

#905 did not rename VendorAnnotations to Annotations... or did I miss something?

clagms:
2020-04-30-Regular-FMI-Design-Meeting

PMAI: Annotations would allow you to place additional information, for eg., for layered standards. Could be used for additional metadata. 

PMAI: Should additional xml files have annotations?

CB: would this be consistent with the extra stuff?

PMAI: Yes.

AJ: Two things: 1 this is a renaming. 2. Is copying all over the place.

PMAI: the 2nd is most important.

AJ: This affects the hashing. 

CB: We should forbid changes to all (except extra).

AJ: We put the annotations everywhere.

AJ: this allows us to be ready for layered standards. These annotations are created by the fmu exporter. Anything except the extra directory is under the control of the exporter.

PMAI: there's an inconsistency with the names.

TS: KW and PMAI, can you put some examples of how the annotations are being used? Can you also show how the validation works?

AJ: I also have no idea on this is done.

PMAI: Each element is associated with a schema that is used to validate it.

KW: we have to be careful to avoid forcing the exporter to have container elements.

Conclusion on renaming:  everybody agrees.

Conclusion on putting it everywhere: everybody agrees.

andreas-junghanns:
related to #937 and #386

Issue #386, opened 2018-10-17
Only one child element within the Tool element of an fmi2Annotation possible


modelica-trac-importer:
**Reported by klausschuch on 7 Oct 2016 08:58 UTC**
The fmi2Annotations are unnecessarily restrictive about the number of elements within the "Tool" element.
The maxOccurs attribute is not specified explicitly and the default value is 1. It should be set to unbounded.

	
```
#!div style="font-size: 80%"
Code highlighting:
  ```xml
<xs:complexType name="fmi2Annotation">
  <xs:sequence maxOccurs="unbounded">
    <xs:element name="Tool">
      <xs:annotation>
        <xs:documentation>Tool specific annotation (ignored by other tools).</xs:documentation>
      </xs:annotation>
      <xs:complexType>
        <xs:sequence> <!-- default value of maxOccurs is 1 -->
        <!-- should be replaced by <xs:sequence maxOccurs="unbounded"> -->
          <xs:any namespace="##any" processContents="lax" minOccurs="0"/>
        </xs:sequence>
        <xs:attribute name="name" type="xs:normalizedString" use="required">
          <xs:annotation>
            <xs:documentation>Name of tool that can interpret the annotation. "name" must be unique with respect to all other elements of the VendorAnnotation list.</xs:documentation>
          </xs:annotation>
        </xs:attribute>
      </xs:complexType>
    </xs:element>
  </xs:sequence>
</xs:complexType>
  ```
}}}
```

----

Migrated-From: https://trac.fmi-standard.org/ticket/386

modelica-trac-importer:
**Modified by andreas&#x2e;junghanns on 6 Feb 2017 10:59 UTC**

KarlWernersson:
Although I agree with you, I don’t see this as something that should be changed in 2.0.1, it can currently be circumvented since any child nodes to the any node are unbounded.
Chaning milestone to 3.0

chrbertsch:
<xs:sequence maxOccurs="unbounded">

is already defined in https://github.com/modelica/fmi-standard/blob/master/schema/fmi3Annotation.xsd

andreas-junghanns:
related to #945 and #937

clagms:
2020-04-30-Regular-FMI-Design-Meeting

KW: This causes more levels of hierarchy without need.

Conclusion: allow any content. PMAI will create a PR.

PR #949, opened 2020-04-30
Regularize and expand Annotations

pmai:
This PR addresses the issues raised in #945 and #386. This mirrors mostly the SSP approach to annotations and regularizes differing versions in FMI (VendorAnnotations vs. Annotations).

Currently in draft form until all separate changesets have landed.


chrbertsch:
Looks really good. 

@pmai : could you please also update and extend the standard text (2.2.7) possibly re-using part of the SSP description (2.1.1., 4.2 of https://ssp-standard.org/publications/SSP10/SystemStructureAndParameterization10.pdf)

KarlWernersson:
Sorry, dinnt plan to change to ready for review, cant seem to find the undo button =(

andreas-junghanns:
Poll during web design meeting:
Do you agree with the proposed direction of this PR (Adding Annotations at every element that is not a collection):
In Favour: Andreas, Pierre, Christian, Irina, Klaus, Otto, Torsten B
Against:
Abstain: Torsten S

@pmai : Please proceed with this PR.

pmai:
This should now address all places where annotations might reasonably be used, and updates the documentation and images to match.

One might want to give the extension mechanisms like annotations and the extra directory a more central place in the outline, but I think this can be handled later on in the editing process if wanted.

So please review again, and comment where you think that I've now gone overboard with the places where annotations can be used...

andreas-junghanns:
My understanding of XSD is patchy, my approval is based on naive pattern matching.

PR #952, opened 2020-05-01
change directory name from icon to icons

### LINKED ISSUE #947: Rename folder /icon to /icons
While previously, the folder /icon contained essentially only one icon.
Now, we are adding more with the T&I changes.
Then, there will be many icons in /icon... not very consistent.

chrbertsch:
resolves #947


andreas-junghanns:
Discussion during web design meeting:
Pierre: this is a new directory anyway.

Issue #937, opened 2020-04-23
Remove redundant default attributes from XSDs


andreas-junghanns:
E.g. `maxOccurs="1"`

andreas-junghanns:
related to #945 and #386

PR #954, opened 2020-05-01
Remove XSD attributes with default values

andreas-junghanns:
If we want all attributes to be there, even with default values, we should be consistent and always have them everywhere. I think removing default settings is cleaner.


andreas-junghanns:
See #937

Issue #936, opened 2020-04-23
Remove xref to independent variable when it refers to an independent parameter.


clagms:
In multiple places of the text, there is an invalid reference to independent.
For instance:

>``The symbol without a subscript references <<independent>> <<parameter,`parameters`>>``

For more, find all occurrences of `<<independent>> <<parameter`.

Probably one needs to also double check all other references `<<independent>>`

chrbertsch:
I understand what *the* intependent variable is (time). But what is an "independent parameter" anyway? Should we not avoid "independent" for this anyway, and find a better description? What could this be?

clagms:
@chrbertsch I think that at some point there was a need to distinguish "calculated" parameter from "independent" parameter ("calculated" being synonymous with "dependent").

Perhaps we can indeed remove the "independent" qualifier altogether, and make sure the reader knows that "parameter" means a non-calculated parameter.

clagms:
> Removing "independent" from the parameter descriptions seems ok, until you find that we have also references to "dependent parameters".
> There is a definition:
> "Dependent <<parameter,`parameters`>> (variables with <> = <>) are..."

Afaik, a parameter can be dependent during initialization mode, if it is calculated from other parameter.

>So, are all dependent parameters "dependent"?

Here the "dependent" refers to the dependenciesKind, right?
If so, then the answer is no. 
Calculated parameters can have a linear relationship with their dependencies (in which case they would be "constant").
The rules in the standard seem to indicate that parameters cannot have "fixed", "tunable", or "discrete", dependencies, since those cannot be declared for initial unknowns.

andreas-junghanns:
> The rules in the standard seem to indicate that parameters cannot have "fixed", "tunable", or "discrete", dependencies, since those cannot be declared for initial unknowns.

If that was true, what variables would be "tunable"?
This seems obviously not true, so whatever makes you believe this must be fixed.

andreas-junghanns:
> > There is a definition:
> > "Dependent <<parameter,`parameters`>> (variables with <> = <>) are..."
> Afaik, a parameter can be dependent during initialization mode, if it is calculated from other parameter.

I was under the impression, that dependent parameters are called calculated parameters.
They change, when tunable parameters change that they depend on.
That can happen during event modes as well.

Back the original question: Are all dependent parameters "calculatedParameters"?
If yes, we should pick one term (I propose calculated because it is more descriptive) and use that everywhere.

clagms:
[...]

PR #956, opened 2020-05-01
Fix dependent parameter

andreas-junghanns:
To address #936

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1107,7 +1107,8 @@ Allowed values of this enumeration:
 
 [[parameter,`parameter`]]
 - `parameter`: A data value that is constant during the simulation and is provided by the environment and cannot be used in connections.
-<<variability>> must be <<fixed>> or <<tunable>>. These parameters can be changed independently, unlike <<calculatedParamter,calculated parameters>>.
+<<variability>> must be <<fixed>> or <<tunable>>.
+These parameters can be changed independently, unlike <<calculatedParameter,calculated parameters>>.
 <<initial>> must be <<exact>> or not present (meaning <<exact>>).
 
 [[calculatedParameter,`calculatedParameter`]]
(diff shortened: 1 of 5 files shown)


docs/2_2_common_schema.adoc
```diff
@@ -1572,19 +1573,19 @@ _Discussion of the combinations that are allowed_:
 
 >|_[green]#(1)#_
 |_<<fixed>> <<parameter>>_
-|_Non-<<tunable>> <<independent>> <<parameter>>_
+|_Non-<<tunable>> <<parameter>>_
 
 >|_[green]#(2)#_
 |_<<tunable>> <<parameter>>_
-|_<<tunable,`Tunable`>> <<independent>> <<parameter>> (changing such a <<parameter>> triggers an external event (ME) or takes effect at the next communication point (xCS), and <<tunable>> <<calculatedParameter>>/<<output>>/<<local>> variables might change their values)._
+|_<<tunable,`Tunable`>> <<parameter>> (changing such a <<parameter>> triggers an external event (ME) or takes effect at the next communication point (xCS), and <<tunable>> <<calculatedParameter>>/<<output>>/<<local>> variables might change their values)._
```
clagms, line 1580: Do you think we should keep the links to the <<parameter>> here? I think these are not meant to link to parameter, because they refer to the tunable parameters.

For instance, the fact that `changing such a <<parameter>> triggers an external event` links to parameter may make the reader thing that there's a parameter associated with each tunable parameter.

I understand if that's not the goal of this PR.

andreas-junghanns: A tunable parameter must be independent.

clagms: Sorry, forgot that the markdown will mess ``<<text like this one>>``.
Updated the original remark.
Anyhow, I understand that this is not the goal of this PR.

PR #961, opened 2020-05-01
change to three digit version string

### LINKED ISSUE #433: Shall we use a three-digit version string
**Reported by cbertsch on 31 May 2018 18:23 UTC**
For FMI 1.0 and FMI 2.0 we use a two-digit version string.
For FMI 3.0 it would be possible to use three-digit version string "FMI 3.0.0" and require, that an FMI 3.0.x importing compliant importing tool must be able to import all 3.0.y versions.

----

Migrated-From: https://trac.fmi-standard.org/ticket/433

chrbertsch:
resolves #433

### DIFF
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -253,7 +253,8 @@ include::../headers/fmi3FunctionTypes.h[tags=GetVersion]
 
 Returns the version of the `fmi3Functions.h` header file which was used to compile the functions of the FMU.
 The function returns `fmi3Version` which is defined in this header file.
-The standard header file as documented in this specification has version `"3.0"`, so this function returns `"3.0"`.
+The standard header file as documented in this specification has version `"3.0.0"`, so this function returns `"3.0.0"`.
+Importing tools supporing the FMI 3.0 standard, must be able to import FMUs with `fmi3Version`=`"3.0.x"` where x is a one- or mutliple- digit number.
 
 ==== Creation, Destruction and Logging of FMU Instances [[creation-destruction-and-logging]]
 
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -254,7 +254,7 @@ include::../headers/fmi3FunctionTypes.h[tags=GetVersion]
 Returns the version of the `fmi3Functions.h` header file which was used to compile the functions of the FMU.
 The function returns `fmi3Version` which is defined in this header file.
 The standard header file as documented in this specification has version `"3.0.0"`, so this function returns `"3.0.0"`.
-Importing tools supporing the FMI 3.0 standard, must be able to import FMUs with `fmi3Version`=`"3.0.x"` where x is a one- or mutliple- digit number.
+Importing tools supporting the FMI 3.0 standard, MUST be able to import FMUs with `fmi3Version`=`"3.0.x"` where x is a one- or mutliple- digit number.
 
 ==== Creation, Destruction and Logging of FMU Instances [[creation-destruction-and-logging]]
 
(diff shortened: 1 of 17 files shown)


chrbertsch:
> Should ":revnumber: 3.0" in index.adoc also turn into 3.0.0?

--> changed. 
> What about the tags, like `v3.0-alpha.3`?

-->are already v.3.0.0-alpha.3
could someone with experience with regular expressions please fix the xsd RexExp  "3[.](\d+)[.](\d+)" to allow for optionally "-alpha.x", or more generally for "-XXXX" to allow also "-beta.y", -RC1"

> What about `fmi-standard-3.0-rc.1.zip`?

Should be `fmi-standard-3.0.0-rc.1.zip`?

> For marketing, 3.0 is likely still ok.

Sure!
> All else would have to change, right?

Yes.

andreas-junghanns:
My last change removes the need for a bug-fix version. I am not sure if this ticket proposes to enforce the bug-fix version number... if not required, the current version is what we need. If we want to require the bug-fix version, we have to undo the last change-set (and fix the reference FMUs).

chrbertsch:
> My last change removes the need for a bug-fix version. I am not sure if this ticket proposes to enforce the bug-fix version number... if not required, the current version is what we need. If we want to require the bug-fix version, we have to undo the last change-set (and fix the reference FMUs).

I think if we follow the path of this PR, we should make the three-digit version mandatory for the exporters, but ensuring, that importers claiming to support FMI 3.0.0 will import al FMI 3.0.x FMUs. 
So I propose to undo the last commit

andreas-junghanns:
I reverted the last change - now we require the bug-fix version number.
The reference FMUs fail - as expected - and need fixing.

chrbertsch:
> The reference FMUs fail - as expected - and need fixing.

How do we do this generally? This is only a draft PR and the Reference FMUs reside in a different Github repo, so the CI will take the master branch to check. So I shall fix the reference FMUs only if and after we have accepted and merged this PR, right?

andreas-junghanns:
> 
> 
> > The reference FMUs fail - as expected - and need fixing.
> 
> How do we do this generally? This is only a draft PR and the Reference FMUs reside in a different Github repo, so the CI will take the master branch to check. So I shall fix the reference FMUs only if and after we have accepted and merged this PR, right?

@t-sommer  should let us know what his preferences are here.

andreas-junghanns:
Discussion web design meeting:
AndreasJ: I could see some use cases for the 3rd digit (adapting import behaviour knowing export bugs).
Pierre: This is theoretically broken but practically useful.
TorstenS: We should have all changes that change behaviour reflect a minor version change, not the bug-fix release.
Pierre: Clarifications could be patch-release-able. This is where the patch-number could be helpful.
Pierre: we should separate this from the add-feature discussion.
AndreasJ: Can it hurt to have the patch number?
Klaus: this might raise the expectation that people want to have 3.0.1 exports without really needing this.
Pierre: We will likely have a 3.0.1 so there could be a use for signaling that someone read the new release and made sure to adapt to the new clarification.
TorstenB: DCP had a similar discussion and there the specification may have a three-digit version, but the XML can only have a two-digit version number. A clarification is a change in contract and that would have to be then a new minor version.

Poll: Is a clarification a change of contract? A bug-fix release would allow only typo-fixes and changes in non-normative text. This forces more minor releases.
Agree: Klaus, TorstenS, TorstenB, Otto
Disagree: Andreas, Christian, Pierre
Abstain: Irina

Poll: Extend the version string to include the bug-fix number, e.g. "3.0.1":
In Favour: Andreas, Irina, Pierre, Christian
Against: TorstenS, TorstenB, Otto, 
Abstain: Klaus

Pierre: I will start writing a proposal for a Declaration of Version Independence
AndreasJ: I will join the fun! (I want to be a founding father as well)

chrbertsch:
There is no consensus to follow this path. 
During the discussion in this PR https://github.com/modelica/fmi-standard/pull/961#issuecomment-625843856,  came up the idea for a "declaration of version independence" (will be followed up in https://github.com/modelica/fmi-standard/issues/983), which would make this proposal obsolete. 
We could re-open this PR, if this would not succeed.

PR #941, opened 2020-04-24
Glossary model exchange

clagms:
This PR summarizes the changes made while adding terms to the glossary in the Model Exchange section.
This is part of a larger effort to update our glossary.

Some extra changes are made, because:
1. They are too small to have in a separate PR
2. I am convinced that they are small improvements to the text

Other issues that I found but did not know how to fix, are summarized in 
https://github.com/modelica/fmi-standard/issues/898
https://github.com/modelica/fmi-standard/issues/926

PR #962, opened 2020-05-03
update overview table

chrbertsch:
The overview table is incorrect partially. 
The formatting of the "except ..." parts is not perfect, perhaps someone has a better idea.

### DIFF
--- a/docs/1___overview.adoc
+++ b/docs/1___overview.adoc
@@ -170,7 +170,7 @@ The following table gives an overview of the features of the different interface
 
 |Clocks
 |icon:check[]
-|Communication Point Clocks: icon:check[] + 
+|Communication Point Clocks: icon:check[] +
 Other clocks: icon:times[]
 |icon:check[]
 |icon:check[]
--- a/docs/1___overview.adoc
+++ b/docs/1___overview.adoc
@@ -170,16 +170,16 @@ The following table gives an overview of the features of the different interface
 
 |Clocks
 |icon:check[]
-|icon:times[] +
-(Except Communication Point Clocks)
+|Communication Point Clocks: icon:check[] + 
+Other clocks: icon:times[]
 |icon:check[]
 |icon:check[]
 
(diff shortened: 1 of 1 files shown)


andreas-junghanns:
I changed the table cells to be more explicit (I think).
Do you agree?

chrbertsch:
> 
> 
> I changed the table cells to be more explicit (I think).
> Do you agree?
OK

masoud-najafi:
Is "Direct Feedthrough" defined somewhere?

andreas-junghanns:
> Is "Direct Feedthrough" defined somewhere?

I added this to the glossary with PR #941.

clagms:
> Is "Direct Feedthrough" defined somewhere?

I think the "direct" qualifier in feedthrough is redundant. Feedthrough is feedthrough. There isn't any "indirect feedthrough".

chrbertsch:
> I think the "direct" qualifier in feedthrough is redundant. Feedthrough is feedthrough. There isn't any "indirect feedthrough".

While in system theory one talks about the "feedthrough matrix", the behaviour when the feedthrough matrix is nonzero is also called "direct feedthrough", see e.g. https://en.wikipedia.org/wiki/State-space_representation
grafik

So I think, "direct feedthrough is the commonly used name, which we should used (and explain in the glossary)

(Perhaps this has been influenced by the usage of Mathworks, which in ML/SL talks about "nondirect feedthrough blocks" http://matrix.etseq.urv.es/manuals/matlab/toolbox/simulink/ug/how_si11.html like an integrator.) 

We should also link this to the useage of "dependencies" in the model structure, https://fmi-standard.org/docs/3.0-dev/#ModelStructure

andreas-junghanns:
I added both terms (direct feedthrough and feedthrough with reference to the first) into the glossary in PR #941

andreas-junghanns:
BTW: We should merge this PR - agreed?

Issue #971, opened 2020-05-05
Simplify State-Machine and remove calls


andreas-junghanns:
After several iterations on how to include the Hybrid Co-Sim features, the state machine and functions driving it are a bit messy. This ticket describes a proposal how to clean this up.

Observations:
A doStep() call can have the following outcomes:
 - fmi3Fatal: the end
 - fmi3Error: requires a SetState or Free/Initialize to continue simulation
 - fmi3OK: doStep finished the step Tc+h
 - XXX: doStep had to stop in the middle of the interval [Tc, TC+h] (for reasons inside or outside the FMU)
 - YYY: doStep just did not make any progress, the FMU state did not change

I propose:
 - XXX : this should be signaled by the new return value fmi3EarlyReturn
 - YYY : this should be signaled by the existing return value fmi3Discard

Reasoning:
 - XXX/fmi3EarlyReturn : a clean state reached in the middle of the co-simulation is just as good as Tc+h
 - YYY/fmi3Discard : fmi3Discard is a valid return value for other functions and signals a NOOP: nothing in the FMU changed (the FMU "Mu": https://en.wikipedia.org/wiki/Mu_(negative))

Consequences:
 - drop the state Step Discarded in BCS and HCS: we checked, there is no reason for this additional state (anymore)
 - drop fmi3EnterStepMode() in BCS (in HCS we have it still for Event Mode -> Step Mode)
 - drop fmi3GetDoStepDiscardedStatus() and add the arguments `*terminate` and `*lastSuccessfullTime` to the fmi3DoStep() function
 - add fmi3EarlyReturn to the return values of fmi3DoStep() and fmi3IntermediateUpdate() to signal early return status and request, respectively.

Discussion:
 - `*terminate` keeps its current semantics to signal if the FMU needs to terminate. This is different from fmi3ERROR: it might simply be the end of its model time.
 - `*lastSuccessfullTime` must always return the current time of the FMU after the doStep() return. This is not just interesting for the case of fmi3EarlyReturn, but also for fmi3OK to show that the internal time of the FMU might slightly differ from Tc+h for internal reasons (e.g. fixed-step solver not able to hit Tc+h precisely)

Issue #574, opened 2019-04-26
Return value for illegal function calls


KarlWernersson:
It is usually not well defined what an FMU should return one of its function is called with an illegal argument. With illegal argument i include things like

- invalid valueReference
- invalid state call.e.g  newDiscreteStates in continoustime mode, setting a fixed parameter after initialization
- providing a  nullState for setFMUstate
- etc.

Currently an FMU exporter has to either return a warning or an error.
a waring has the risk of getting ignored while an error forces the master to terminate even though it might be able to recover. 
I would argue that what behavior you want is more on the master side than on the exported FMU side.

if a master is non interactive during simulation, you would probably want to terminate since you have either programed your master wrong or the exported FMU is wrong and there is little you can do now to fix it.
However if the master is interactive, then a user using it could try do allot of dumb things and it would be sad to force termination instead of giving a proper message, e.g. "illegal call, please do something else"

We could call it
fmi3Illegal

chrbertsch:
Regular Design meeting: 
Masoud: we need something like this during initialization (e.g. like. This input would lead to a dvision by zero)
Karl: in this specific case, one could use discard. One  could use discard for setReal
Klaus: agree. Perhaps we should not mix these two issues: numerical reasons, not allowed function called.
Klaus: Strongly reminds me of DCP discussions ... distinguish between several log / error messages: what can the master expect that the slaves checkes. The master can e.g. know if the call of setting a value went wrong.
Kaska: We need a clear specification how a function call reacts to an error.
Klaus: Look at DCP specfication 3.4.7 , what happens for PDUs (these are essentially function calls)
Kaska: one should get the same error codes for different FMUs in the same error case
Karl: My proposal was more lightweight. Error code are good, but would need a lot of work.
Klaus: in DCP, the focus was on realtime systems. With error codes, more automatization is possible.


Poll (multiple options possible): 
* do nothing (keep it as in FMI 2.0, one could use error or discard)
** Antoine, Andreas, Masoud, Kaska
*define "fmiIllegal" + non standardized human-readable message (could be extended later with error codes)
** Karl, Christian B., Masoud, Klaus, Thorsten G. 
* define detailed Error codes for specific cases that we can imagine.
** (no one)

Karl will prepare to proposal for the Design meeting.

pmai:
[...]

Issue #535, opened 2019-02-13
Rework fmi3Discard mechanism for DoStep


pmai:
As discussed in the design meeting on 2019-02-13:

The current definition of fmi3Discard only allows restarting of the calculation if full-fledged fmi3Get/SetFMUState support is present. Since fmi3SetFMUState can already be called in terminated/error state, fmi3Discard does not provide more of a potential mechanism for resuming than fmi3Error (except for the indication of the actual reason for the error).

However it is rather more likely that implementations can more easily support reseting the FMU state to the state when entering the fmi3DoStep on a failed step, than having full fledged fmi3SetFMUState support.

Therefore the definitions should be changed, so that:

- fmi3DoStep returning fmi3Discard means the FMU has ensured that the state is unchanged from entering the fmi3DoStep call, so the FMU remains in step state, and a new fmi3DoStep call starting from the original time can be called.
- FMUs that cannot support this must return fmi3Error, which auto-transitions to terminated state, but still allows fmi3SetFMUState to restart the FMU.

To be decided: Add a new fmi3-Return Value to indicate that the failed fmi3DoStep is due to not being able to complete the step, and not some other error (e.g. fmi3StepFailed).

APillekeit:
FMI Design Meeting in Regensburg:
Discussion points:
- fmi3Discard gets a new meaning: last communication step is rolled back internally by the FMU. The master can call doStep() again from last communication point time.
- add fmi3StepFailed for the case that the FMU can not roll back internally. The internal time is undefined. 

The side effects have to be discussed in more detail.

chrbertsch:
@KarlWernersson :  will write down a solution discussed in Regensburg

KarlWernersson:
Disregarding asynchronous doStep, the concept of fmiDsicard after a doStep  in Fmi2 is actually an earlyReturn where you cannot continue the simulation from. When doStep returns discard you end up in stepFaild that we accidentally removed from the new statemachine in fmi3. In fmi2 two cases can get you into this state. Either an internal Termination is called. E.g.  the modelica function terminate, or anything else is causing the simulation to fail. The second case doesn’t in practical differ much from an fmiError except that you can retrieve the last successful time. And you are at least implicitly supposed to be able to retrieve the values at that time instance, while error also allows you to return values it does not give any of those implicit guarantees. But that diference is not realy cleary defined.

[...]

PR #972, opened 2020-05-05
Simplify state machines

andreas-junghanns:
Discussions with Torsten* revealed some potential for simplification. See #971.

Missing still the changes to the BSC and HSC state machines.
I tried to do that on draw.io, but strange things have happened there... I need TorstenS or his Mac, maybe both to figure this out...


andreas-junghanns:
@pmai   : Your issue #535 should be worked into this proposal as you suggested (fmi3Discard truly discards the doStep() and turns the function call into a NOOP).

@KarlWernersson : Your issue #574  should also be - at least partially - addressed (Illegal arguments can be signaled with fmi3Discard and that turns the function call to NOOP).

docs/2_1_common_api.adoc
```diff
@@ -203,27 +203,38 @@ this function has shown the prepared information message to the user.
 
 `fmi3Discard`::
 [[fmi3Discard,`fmi3Discard`]]
+With this return value, the FMU signals that the function called could not achieve the defined effect and was able to keep its state as before the call.
+This essentially indicates that the function call was a NOOP.
```
KarlWernersson, line 209: I had to google NOOP,
Maybe I who is un educated since I lack a CS background with very limited assembler experience.
Is it a commonly used expression or should it be changed?

docs/4_2_basic_co-simulation_api.adoc
```diff
@@ -92,19 +103,24 @@ The function returns:
 `fmi3OK`::
 if the communication step was computed successfully until its end.
 
+`fmi3EarlyReturn`::
+if the FMU computed successfully only a subinterval of the communication step.
+
 `fmi3Discard`::
-if the slave computed successfully only a subinterval of the communication step. The FMU enters the *Step Discarded* state, see <<state-step-discarded-co-simulation>> for more information.
+if the FMU could not compute any new (more advanced) state successfully, but was able to return its internal state to exactly the state when <<fmi3DoStep>> was called.
+This return code indicates that the call to <<fmi3DoStep>> was essentially a NOOP.
```
KarlWernersson, line 110: NOOP here to

andreas-junghanns:
Yes, we should avoid CS lingo - I will change NOOP into readable text.
Allowing fmi3Discard for all or more functions would deal with the case you raised in #574: If you receive illegal arguments, you can return with fmi3Discard.
If that was not clear from the text, I should improve it.

chrbertsch:
> The `intermediateUpdate` callback should not return `fmi3Status` but `fmi3Boolean` to indicate an early return request. Just like for the other callbacks (e.g. `logMessage`) it is pointless to signal a problem in the importer to the FMU (it is in control anyways).

@t-sommer : but does #977 not rely on these return values?

t-sommer:
#977 would become obsolete, if we adopt my proposal

andreas-junghanns:
> The `intermediateUpdate` callback should not return `fmi3Status` but `fmi3Boolean` to indicate an early return request. Just like for the other callbacks (e.g. `logMessage`) it is pointless to signal a problem in the importer to the FMU (it is in control anyways).

Such a simplification sounds too good to be true, is it?
 1. returning to continue the doStep: fmi3False
 2. returning to return early at a later time, normal early return, FMU will properly finish doStep and signal its internal time with lastSuccessfulTime
 3. returning to return immediately, because of error

How do we know 2. from 3.?
Do we (the FMU) have to know the difference?

t-sommer:
> Do we (the FMU) have to know the difference?

I don't think that's necessary. The FMU will pass back control to the master after the early return. The master can then terminate if something went wrong (without letting the FMU know the reason).

andreas-junghanns:
> > Do we (the FMU) have to know the difference?
> I don't think that's necessary. The FMU will pass back control to the master after the early return (`requestedReturnTime == intermediateUpdateTime`). The master can then terminate if something went wrong (without letting the FMU know the reason).

In case of error, the FMU can return immediately without trying to finish the doStep to a state of order that would be required for a proper early return. I think this is important, unless we want the FMU to always finish properly, also when an error was detected.

andreas-junghanns:
@t-sommer : now

PR #980, opened 2020-05-14
Fix xrefs (logMessage, nextEventTime*, earlyReturn)

andreas-junghanns:
just a few more xrefs added


andreas-junghanns:
All changes are syntactic improvements only (or meant to be).
Please check and if you agree, this should be an easy merge.

PR #982, opened 2020-05-15
Remove function-state tables

andreas-junghanns:


andreas-junghanns:
We decided to remove the function-state tables in one of the design meetings.
This PR simply removes these 4 tables.
Rational:
 - The tables are not really helping - the many footnotes with indicate that a simple table is not sufficient
 - we chose a different format: We now describe each state in the text with the allowed and forbidden functions

chrbertsch:
> we chose a different format: We now describe each state in the text with the allowed and forbidden functions

Maybe  I miss something, but I find only a small portion of the information that was contained in the table in the text. 
E.g. For ME in the "instantiated mode", the table said that 12 API function calls were allowed ("x"),   four conditionally allowed ("1") and 18 forbidden (blanks)

In the text description of the state "instantiated", only  the conditional cases are mentioned, not even spelling out the corresponding function names.

grafik

Isn't this loss of information?

andreas-junghanns:
These texts need fixing, but the tables were incomplete at best, so the information I removed would not have helped. If you need it, the tables in 2.0.1 are the next best thing to correctness we know.

PR #986, opened 2020-05-15
Discard for illegal argument

andreas-junghanns:


docs/2_1_common_api.adoc
```diff
@@ -213,6 +213,7 @@ The output argument values are not defined, but the computation can continue.
 When debug logging is enabled (`loggingOn == fmi3True`) the function <<logMessage>> was called by the FMU.
 Advanced master implementations can try alternative approaches to drive the simulation by calling the function with different arguments or calling another function.
 Otherwise the master has to treat this return code like <<fmi3Error>> and has to terminate the simulation.
+_[When detecting illegal arguments or the wrong function call at the current state, the FMU may return <<fmi3Discard>> instead of <<fmi3Error>> to allow the importer to attempt an alternative call.]_
```
chrbertsch, line 216: But the master cannot distinguish between a failed doStep due to numerical reasons and a wrong function call e.g. due to illegal arguments. 
Wouldn't it be better to introduce an fmi3Illegal return value clearly telling the master "it's your fault"?

andreas-junghanns: Of course there is a large number of cases for errors/issues that we [sc]hould distinguish and not just illegal arguments and numerical errors.
However, the point is not to declare the reasons for failing, but to declare if we can continue and how (or not). And it is a quality of implementation issue if you as master algorithm have a number of heuristics on how to continue when a fmi3Discard comes back.
And if you have no clue, treat fmi3Discard just like fmi3Error.

andreas-junghanns:
> 
> 
> Illegal arguments and illegal calling sequences are bugs IMHO. We should not encourage implementers to use fmi3Discard for these cases.

I see your point: bad habits do not deserve protocol.

I was also toying with the idea to add another logCategory, but the same argument applies here: The import tool cannot do much with the information of "bad treatment", the users even less (they cannot change the tool).

Your argument leads to: Illegal arguments are errors, even if the FMU can ignore them and preserve its state. I changed the PR to reflect this.

I am still not sure this is the best way to handle this because the fmi3Discard essentially declares "unchanged state in FMU" and the FMU, by rejecting illegal arguments, can keep its state unchanged.
But I also agree that if the importer calls with illegal arguments, what are the chances that the importer can remedy this?

The choice is clear (both states of the PR reflect them), current state is the stricter of the definitions.

@KarlWernersson , @pmai : Any words of wisdom?

KarlWernersson:
Currently fmiDiscard is allowed on most setXXX function and some getXXX function, + dostep

- To my understanding and from the few examples that exists, discard in setXXX is used to signal that it was illegal vale, in FMI 2 specifically inputs outside their min/max value is mentions, in fmi3 some more examples like fixed/tuneable parameters when they are not allowed.

- For the getter function its is only a subset of them used for calculation of continuous variables that can return discard and only for ME. The specification is not so clear on what that means but in my understanding it has been that internal computations fails with the current input/state selection please change them (i.e reduce step size if you use a variable step size solver, or posibly)  and maybe you will have better success.

- For doStep discard says that your step succeed part of the way. In fmi2 you couldent do much except terminate if the discard was due to internal termination, for FMI3 how to continue form that state has been expanded.

So, we have three use cases, bad Input values, failed continuous computation in ME, partially failed doStep in CS. Whichever of these three use cases is clear since its based on the function returning.

If we start using fmiDiscard for general illegal input arguments, you start to contaminate these use cases and it will be hard to distinguish theme on a master level. Just as @chrbertsch said.

So therefore I am against this change. (I might have had a different opinion earlier but i am now convinced that was wrong)

I proposed earlier to use separate return value
https://github.com/modelica/fmi-standard/issues/574

But i think that @t-sommer says

> Illegal arguments and illegal calling sequences are bugs IMHO. We should not encourage implementers to use fmi3Discard for these cases.

Is a good argument, also against a separate return value.

With that i woudd even go so far as to say we should remove the discard for setting parameters when not allowed that has been added for FMI 3.

andreas-junghanns:
> With that i woudd even go so far as to say we should remove the discard for setting parameters when not allowed that has been added for FMI 3.

@KarlWernersson : Do you agree with my revised PR? I followed @t-sommer´s proposal. If setting parameters is done in the wrong state, fmi3Error would have to be returned. If you agree with the current proposal, please approve.

PR #992, opened 2020-05-19
Add communication delay requirement for intermediateStepFinished==false

### LINKED ISSUE #991: intermediateStepFinished?
If I understand correct if the callbacks return intermediateStepFinished == false it indicates that the internal solver step is just an attempted step but you want to update input to help it on for the next attempt. 

Lets say now that you have two FMU's A and B both with inputs and outputs connected to each other. These are both fairly complicated models so several returns of intermediateStepFinished == false can occur.

If now FMU A return intermediateStepFinished == True, and FMU B return intermediateStepFinished == False, is there some way to recalculate the step for FMU A with updated input? As long as FMU B has not finished one would expect that the intermediate output from that fmu will change, but it is no longer possible for FMU A to update its input accordingly. 

Have i understood the use case for intermediateInput with intermediateStepFinished == false correctly?

Do we need to fix this? I.e do we need the master to control when to finish these step calculation similar to event iteration.

robbr48:
Resolves #991.

This requirement is necessary to avoid potential lockups, where FMU A has finished its internal step while FMU B is still in an intermediate step.


andreas-junghanns:
I am confused: "This must only be used on input variables with a communication delay, which must be greater than the internal solver step in the FMU." -- how would the master know what that delay is?

Or should we rather write as second sentence: "The FMU must call `intermediateUpdate` with `intermediateStepFinished == fmi3False` if at least one input change acts beyond the current communication point (`intermediateUpdateTime`)."?
At the very least, can we please avoid "This" in the sentences as such unclear back references can cause confusion.

@robbr48 : Could you also provide a concise scenario for this usage as a non-normative text explaining the intended use here?

robbr48:
@andreas-junghanns The delay is located in the master. It reads outputs from FMU A, inserts them into an interpolation table and uses them to compute interpolated inputs for FMU B with a specified delay. The delay must be greater than the internal step size in FMU B.

If the delay is physically motivated, the two FMUs can then be executed in parallel within a limited time frame without compromising numerical stability. This is the basic idea behind the transmission line modelling technique.

I am not sure I understand your second sentence. `intermediateStepFinished==false` only indicates that the outputs from the FMU are the result of a temporaray solver state and should not be forwarded to other FMUs or log files.

I agree that the "this" is confusing. I will remove it and try to write a more clear explanation.

andreas-junghanns:
Mmh, I think I understand this better now, but can you also spend a second paragraph on what happens if `intermediateStepFinished == fmi3True`, please? Maybe the contrast between both explanations highlights even better what is intended here?

chrbertsch:
We should also define what a "temporary solver step" is and what one can expect when calling fmi3GetXX for such a step. If I understand right, one cannot rely on such a result, but it could still change. Can the master find out somehow, if a previously received value from a temporary solver step is no longer valid?

klausschuch:
I think this PR is too restrictive (see my comment in #991).

However, after re-reading the current description, i think this sentence should be changed:
"If intermediateStepFinished == fmi3False, the Co-Simulation master shall only use intermediate output variables to compute intermediate input variables at the current state."
I do not even understand the intention of this sentence.

robbr48:
@chrbertsch In my opinion, "previously" received values from temporary steps are never valid. Outputs received during an intermediate update with `intermediateStepFinished==false` are only valid during the execution of that particular call to the callback function.

chrbertsch:
> @chrbertsch In my opinion, "previously" received values from temporary steps are never valid. Outputs received during an intermediate update with `intermediateStepFinished==false` are only valid during the execution of that particular call to the callback function.

Is there any use case what can be done with Outputs received during an intermediate update with `intermediateStepFinished==false` at all?

I personally do not see one, and would propose to remove this possibility in order to simplify it and remove invitations for misuse ...

@robbr48 : could you please also comment of https://github.com/modelica/fmi-standard/issues/993?

chrbertsch:
@KarlWernersson , @robbr48 , @klausschuch : 
Could you please check, if #1047 fixes this the intention of this PR also, also adressing @klausschuch 's concerns?
Please comment there. 

Can we then close this PR?

KarlWernersson:
I really dont like that we are so specific we expalin verry detailed what you can do, but not what they are.

Specificaly
_If <<intermediateStepFinished, `intermediateStepFinished == fmi3False`>>
the Co-Simulation master shall only use intermediate output variables to compute intermediate <<input>> variables at the current state for the same FMU.
The master *MUST NOT* pass information obtained with `fmi3Get{VariableTypes}` in this case to other connected FMUs nor use these information to derive any kind information (e.g. extrapolation) that is passed to other connected FMUs or other model parts.
However, the master is allowed to use the obtained with `fmi3Get{VariableTypes}` for calculations that are directly passed back to the same FMU `fmi3Set{VariableTypes}`._

First of is it really true that you use to output for the input of the same fmu? dont you connect them (maybe not directly but with some master connection)  and the master uses the intermediateOutput of the fmu's in to update and improve the itermediate inputs? Or have I misunderstood it. I haven't worked with TLM so I am not expert here.

Instead of telling people what and what they can do with these values, which they anyway can ignore.
inst it better to explain what they are and how to avoid to shoot yourself in the foot with them.

example (wording could be improved)

<<intermediateStepFinished, `intermediateStepFinished == fmi3False`>>
The solver as returned at an unfinished step calculation
if intermediateGET is allowed you can retirve the temporary values, since the step is not finished they are not final and should not be used for direct communication between fmu's.

If intermediateSet is allowed the master may try to give a different input for the step calculation.

Then we could have the explanation of TLM as usecase.

robbr48:
No, we don't forward the intermediate output directly back the same FMU as an intermediate input. That would make no sense. The intermediate inputs are calculated using both the intermediate outputs and delayed variables from the connected FMU.

I agree with your suggestion, except for the word "direct" in this sentence: 
> [...] since the step is not finished they are not final and should not be used for direct communication between fmu's.

These variables should not be communicated to other FMUs in any way.

chrbertsch:
@robbr48 and @andreas-junghanns :  Could we close this in favour of #1047  (so you could possibly adapt there)

chrbertsch:
@robbr48 Please make sure that  #1047 covers the TLM specific aspects

PR #995, opened 2020-05-22
Remove leftover from old alias definition

chrbertsch:


andreas-junghanns:
@chrbertsch : I changed two more sentences to clarify language (and remove a "of of"). Please approve or merge right away if you agree with the changes.

PR #1004, opened 2020-05-23
Fix description of causality "parameter"

chrbertsch:
as the current description contradicts parameter tuning and parameter propagation


chrbertsch:
@andreas-junghanns : in https://fmi-standard.org/docs/3.0-dev/#definition-of-model-variables 
there is the non-normative-text: "How to treat tunable variables"

> A parameter p is a variable that does not change its value during simulation ....Therefore, "tuning a (structural) parameter" during simulation does not mean to "change the parameter online" during simulation ....

This is not yet reflected in this PR. Do you have an idea, how we could link this?

PR #1013, opened 2020-05-24
Improve fmu distribution description

### LINKED ISSUE #1011: Improve 2.3 FMU distribution
- improve wording of first sentences  ("An FMU description consists of several files. An FMU implementation may be distributed  .... " " The implementation implements ..."
- the "optional" information move to next lines depending on zoom factor
grafik

- Currently only the possibility two have 2 different shared libararies is described in 
 
> In an FMU multiple interface types might be present. If in both cases the executable part is provided as a shared library, then two different or only one library can be provided. The library names are defined in the modelIdentifier attribute of elements <fmiModelDescription><ModelExchange|{Basic|Hybrid|Scheduled}CoSimulation>:
> 
> [Example of different libraries:

```
   binaries
     x86_64-windows
        MyModel_ModelExchange.dll   // modelIdentifier of <ModelExchange> =
                                    //    "MyModel_ModelExchange"
        MyModel_CoSimulation.dll    // modelIdentifier of <CoSimulation> =
                                    //    "MyModel_CoSimulation"
]

```
... but it can be four different types for ME, BCS, HCS, SCS, right?

- add sub-heading, to make reading and referencing easier, e.g. for
-- zip folder structure
-- platform tuple defintion
-- multiple shared libraries for different FMU types
-- documentation directory
--sources directory
-- resources directory
---  native model representations for needsExecutionTool
-- extra directory

chrbertsch:
resolves #1011

### DIFF
--- a/docs/2_3_fmu_distribution.adoc
+++ b/docs/2_3_fmu_distribution.adoc
@@ -68,9 +68,10 @@ extra                         // Additional (meta-)data of the FMU (optional)
 
 ==== Documentation directory [[documentation-directory]]
 
-===== License information [[license-information]]
-_[The subdirectory `licenses` can be used to bundle all license files._
-_It is strongly recommended to include all license and copyright related information in the licenses folder of an FMU (especially but not only for contained open source software) - the `license.{txt|html}` file can serve as an entry point for describing the contained licenses._
+===== Licenses subdirectory [[license-information]]
+This optional subdirectory can be used to bundle all license texts for the code, binaries or other material (documentation, content of resources folder) contained tin the FMU. If it is present, it must contain either a `license.txt` or `license.html` file as entry point.
+
+_[It is strongly recommended to include all license and copyright related information in the licenses folder of an FMU (especially but not only for contained open source software) - the `license.{txt|html}` file can serve as an entry point for describing the contained licenses._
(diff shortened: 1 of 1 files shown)


chrbertsch:
> I have only one remark that might require changes: you have now introduced sections with only non-normative text, e.g. the license directory. Even if the directory itself is optional, should we not recommend that iff present, it must contain either a license.txt or license.html?

@andreas-junghanns: Please review my last commit. 
(You are also (always) very welcome to improve it by committing directly to the branch.)

I personally would even go farther and 

- make both the `documentation` and `documentation/licenses` folders mandatory
- drop the `licenses.txt` and make only the `licenese.html` mandatory
- for binary FMUs go on with good exameple and make the BSD license text of any header files from the FMI standard, that was used to compile the FMU, mandatory to be contained in this folder ...

But am not sure if we can find consensus for these proposals ....

andreas-junghanns:
I am not sure we win if we force the existence of a /documentation directory - we would end up with a lot of empty files. If nothing is to say, non-existence is a clear signal.

The rest of your proposal as now reflected in the changes has my full backing.
As this is a substantial change set, I would like at least one more approval before merging.

andreas-junghanns:
> I am not sure we win if we force the existence of a /documentation directory - we would end up with a lot of empty files. If nothing is to say, non-existence is a clear signal.

I will merge this PR now, if we feel your proposal may get traction, we should open another issue.

Issue #1016, opened 2020-05-24
Review of 3.2 (ME API)


chrbertsch:
https://fmi-standard.org/docs/3.0-dev/#_application_programming_interface: 
If the super-dense time before a call to
-  [x] explain argument `x`; describe or reference internal order of the argument x of `fmi3SetContinuousStatesTYPE`, also referring to array serialization
![grafik](https://user-images.githubusercontent.com/16363372/82750631-b0d8ec80-9db1-11ea-9017-c67c1c824e6b.png)

- [x] In explanation of `fmi3NewDiscreteStatesTYPE`: Correct sentence " fmi3NewDiscreteStates was (tR,tI), then the time instant after the call is (tR,tI)." to "... after the call is (tR,tI**+1**)."

- [x] I do not understand the non-normative text below the definition of `fmi3EnterContinuousTimeModeTYPE`: 

> [This function might be used for the following purposes:
>     If the FMU stores results internally on file, then the results after the initialization and/or the event has been processed can be stored.
>     If the FMU contains dynamically changing states, then a new state selection might be performed with this function.]

- [x] Describe the order of the the continues states returned from 
`typedef fmi3Status fmi3GetContinuousStatesTYPE` ...

- ~~in the text a function `fmi3EnterContinuousTimeModeTYPE(fmi3Instance instance);` is defined, but in the state machine for ME, the function  `fmi3EnterStepMode `is displayed (which belongs to CS, right?):
![grafik](https://user-images.githubusercontent.com/16363372/82761123-86f6e880-9df8-11ea-9086-8167411b6657.png)~~

- [x]  Text below ME State machine: 

>  For example, the state chart indicates that when an FMU for Model Exchange is in state Continuous-Time Mode, a call to fmi3Set{VariableType} for a discrete input is not supported.

This was true for FMI2.0, but in FMI 3.0 this cannot be seen from the state machine. Replace this text with  a new one telling that the State machine displayes allowed transitions (and call of transition functions), but the allowed calls to functions in each state are defined in the text below.

- [x] Generally missing in 3.2.3.x: Detailed list of allowed and forbidden function calls. -> singled out as #1151 

~~- 3.2.3.6: missing point at the end.~~

- [x] 3.2.5. Clocks API:  "For a outputClock, the environment" --> "For an ..."

clagms:
[...]

PR #1025, opened 2020-05-26
fix fmi3EnterContinuousTimeMode in ME State machine

chrbertsch:
fix finding in #1016 

In the text a function fmi3EnterContinuousTimeModeTYPE(fmi3Instance instance); is defined, but in the state machine for ME, the function fmi3EnterStepMode is displayed (which belongs to CS, right?):
grafik
grafik


andreas-junghanns:
The name of the PR is focused only on one issue - I therefore merge this PR now.

PR #1030, opened 2020-05-27
Make resouces folder optional and describe return of NULL pointer

### LINKED ISSUE #1020: Is NULL allowed for argument resourceLocation of fmi3InstantiateXXX functions?
I think we should explicitly clarify if NULL is allowed or not for the argument "resourceLocation" of the fmi3InstantiateXXX functions.

chrbertsch:
Resolves #1020 
and discussion in #814

### DIFF
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -300,7 +300,7 @@ The following schemes must be understood by the FMU:
 +
 _[Example: An FMU is unzipped in directory `C:\temp\MyFMU`, then `resourceLocation` = `file:///C:/temp/MyFMU/resources` or `file:/C:/temp/MyFMU/resources`._ +
 _The <<fmi3Instantiate>> functions are then able to read all needed resources from this directory, for example maps or tables used by the FMU.]_ +
-A NULL pointer is returned for `resourceLocation`, if no resource location can be provided to the FMU, which may occur
+A NULL pointer is supplied for `resourceLocation`, if no resource location can be provided to the FMU, which may occur
 
 ** if the FMU does not contain a resources folder, or
 ** if the environment is not able to provide an URI to the resources folder _[e.g., if the environment does not have a file system. If the FMU in such a case cannot be sinmulated, as it depends on the resources folder, it shall terminate with an error.]_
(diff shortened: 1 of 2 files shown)


andreas-junghanns:
Note: I believe this is actually better described now than it was before.

docs/2_3_fmu_distribution.adoc
```diff
@@ -55,12 +55,12 @@ binaries                      // directory containing the binaries (optional)
...
    // has dependencies on those items, it will not be able to find them.
-resources                     // resources needed by the FMU
+resources                     // resources used by the FMU (optional)
    // data in FMU specific files which will be read during initialization;
    // also more folders can be added under resources (tool/model specific).
    // In order for the FMU to access these resource files, the resource directory
-   // must be available in unzipped form and the absolute path to this directory
-   // must be reported via argument "resourceLocation" of fmi3InstantiateXXX.
+   // shall be available in unzipped form and the absolute path to this directory
+   // can be reported via argument "resourceLocation" of fmi3InstantiateXXX.
```
pmai, line 63:
```suggestion
   // should be reported via argument "resourceLocation" of fmi3InstantiateXXX.
```

docs/2_1_common_api.adoc
```diff
@@ -294,12 +294,17 @@ It must be passed unchanged to the FMU. This argument must not be a null pointer
...
-
+** Mandatory -- `file` with absolute path (either including or omitting the authority component); +
 ** Optional -- `http`, `https`, `ftp`.
+
 +
-_[Example: An FMU is unzipped in directory `C:\temp\MyFMU`, then `resourceLocation` = `file:///C:/temp/MyFMU/resources` or `file:/C:/temp/MyFMU/resources`._
-_The <<fmi3Instantiate>> functions are then able to read all needed resources from this directory, for example maps or tables used by the FMU.]_
+_[Example: An FMU is unzipped in directory `C:\temp\MyFMU`, then `resourceLocation` = `file:///C:/temp/MyFMU/resources` or `file:/C:/temp/MyFMU/resources`._ +
+_The <<fmi3Instantiate>> functions are then able to read all needed resources from this directory, for example maps or tables used by the FMU.]_ +
+A NULL pointer is returned for `resourceLocation`, if no resource location can be provided to the FMU, which may occur
```
pmai, line 303:
```suggestion
A NULL pointer is supplied for `resourceLocation`, if no resource location can be provided to the FMU, which may occur
```

docs/2_3_fmu_distribution.adoc
pmai: Can seems to lenient, since this implies an implementation is completely free to chose whether to supply it, whereas we probably want to state that it really should, unless the stated reasons in the non-normative sections apply...

PR #1038, opened 2020-05-28
Reword Early Return and Event handling description

### LINKED ISSUE #665: Clarification over event management in Co-Simulation
Hi,

I have some concerns about event management in Co-Simulation:

- I understood how an FMU can signal an event to the co-simulation master, however what is not clear to me is how this event is propagated to the other FMUs. For example, are all the FMUs pushed to the Event mode, and the event is somehow provided as an input parameter? Or the event is just a way to set a new communication point to update variables values? 

- What is the real difference between a clock tick and an event? I think that what is not really clear is the general concept of Clock Support in Hybrid Co-Simulation mode, and how it really differs from the Early Return.

- Is the function fmi3NewDiscreteStates() used to manage super-dense time?

Thanks.

andreas-junghanns:
Resolves  #665 

Initially I thought this chapter had a extraneous part that needed to be moved to its proper place. But I think the idea of the chapter is rather to explain the early-return and event-handling mechanisms. I tried to fix this with this PR

### DIFF
--- a/docs/5_2_hybrid_co-simulation_api.adoc
+++ b/docs/5_2_hybrid_co-simulation_api.adoc
@@ -19,7 +19,7 @@ There may be an event iteration in *Event Mode* at a Newtonian time instant caus
 
 Based on the information provided by <<fmi3CallbackIntermediateUpdate>>, additional information about the discontinuity at that time instant can be obtained by the master from the FMU by calling <<fmi3NewDiscreteStates>> and <<fmi3GetClock>>.
 
-==== Handling a Successful Early-Return by the Co-Simulation Master in Hybrid Co-Simulation
+==== Handling Early Return and Events in Hybrid Co-Simulation
 
 If the FMU is successful in conducting an early return, <<fmi3DoStep>> returns with <<earlyReturn,`earlyReturn == fmi3True`>>.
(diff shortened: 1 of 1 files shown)


docs/5_2_hybrid_co-simulation_api.adoc
```diff
@@ -34,24 +34,28 @@ The Co-Simulation master can also call <<fmi3EnterEventMode>> at communication i
...
 The master can start a resynchronization of FMUs at an event time, if the <<currentCommunicationPoint>> has passed the event time, the master can roll-back the FMU and repeat the step with a suitable <<communicationStepSize>> (if the FMU supports the roll-back).
 
-In *Event Mode* and only after <<fmi3DoStep>> returned with <<earlyReturn,`earlyReturn == fmi3True`>> the function <<fmi3NewDiscreteStates>> may be called.
+[source, C]
+----
+include::../headers/fmi3FunctionTypes.h[tag=NewDiscreteStates]
+----
+
+In *Event Mode* and only after <<fmi3DoStep>> returned with <<earlyReturn,`earlyReturn == fmi3True`>>, and <<fmi3EnterEventMode>> was called, the function <<fmi3NewDiscreteStates>> may be called.
 Only the following output arguments may be considered:
```
otronarp, line 51: We discussed having some stronger wording here. Perhaps:  
Only the following output arguments are defined by the FMU:  
...items...
All other output arguments from <<fmi3NewDiscreteStates>>  are undefined and have no meaning in Hybrid Co-Simulation.

PR #1043, opened 2020-05-29
Change types in xsd

### LINKED ISSUE #940: Properly deal with duplicate XSD element names in Schema
This issue was created due to the discussion in https://github.com/modelica/fmi-standard/pull/904

Summary:
1. Some xsd elements have duplicate names (e.g., Float64, GraphicalRepresentation).
1. While item 1 may make conceptual sense, it leads to the following problems:
   1. The tool XSDDiagram has difficulties selecting the right element to export to png when there are two elements with the same name, in the same schema. 
   1. In the adoc, it may be hard to xref a xsd element by name if there are two elements for the same name.
   1. The reader may have trouble distinguishing xsd elements that have the same name (e.g., the Float64 type and the Float64 value).

Items 2.ii. and 2.iii. are "maybe"'s. Item 2.i. is a certainty. 
Moreover, it may lead to difficult to detect problems down the line, where a new XSD element $A$ gets added with the name of an already existing XSD element $B$, and suddenly the png of $B$ gets messed up.
The png of $B$ may be in the other side of the document, and may be even regenerate much later than the change that broke it, so it may be difficult to detect this problem until someone goes over the whole document.

andreas-junghanns:
Resolves #940


andreas-junghanns:
I could not regenerate the images, the PS script requires a the XSDDiagramConsole.exe that I did not know how to find. (please amend the contributing.adoc)

Still missing: renaming of one of the GraphicalRepresentation elements... on it...

andreas-junghanns:
@clagms : Please regenerate the images and check all my changes - this was a lot of flying-by-ear.

beutlich:
> PS script requires a the XSDDiagramConsole.exe

Strange. Didn't I previously add automatic download and extraction of XSDDiagramConsole.exe exactly for that reason.

nickbattle:
This change affects the type names, but it doesn't seem to affect the docs/examples which use those type names. Are the examples handled separately, or should they be included in the one PR?

clagms:
> This change affects the type names, but it doesn't seem to affect the docs/examples which use those type names. Are the examples handled separately, or should they be included in the one PR?

Good point @nickbattle . These should be fixed now.

>Strange. Didn't I previously add automatic download and extraction of XSDDiagramConsole.exe exactly for that reason?

Indeed. I've regenerated the figures now. Later we'll figure out what went wrong on @andreas-junghanns 's side.

@andreas-junghanns , this PR looks good. However, we might have to go over the text and search for all the elements that have been renamed, and see if they need to be renamed in the text. There are tables, table captions, figure captions, etc..

nickbattle:
@clagms You only fixed a couple of the examples files affected? The `variable_types.xml` example includes one instance of every type, but there are several other files affected too.

clagms:
> @clagms You only fixed a couple of the examples files affected? The variable_types.xml example includes one instance of every type, but there are several other files affected too.

@nickbattle Can you clarify which files still need to be fixed? I blindly fixed all files that are checked by validate_xml.py. I just looked at the `variable_types.xml` file, and, as far as I understood the changes that @andreas-junghanns did, only the elements under fmi3Type.xsd changed their names. The elements in variable_types.xml are under fmi3Variable.xsd. 
Did I miss something?

nickbattle:
@clagms Ah! My mistake, sorry. I thought it had changed the names of the types, not the names of the type names... as it were :-)  So this is probably OK, because there are only a couple of type definitions in the examples.

clagms:
Ehehe this just motivates avoiding duplicate names :P

andreas-junghanns:
@pmai : This is a tricky one - can you please have a look?

beutlich:
> Later we'll figure out what went wrong on @andreas-junghanns 's side.

Yes, would be interesting to see the actual error message from @andreas-junghanns 

> please amend the contributing.adoc

@andreas-junghanns https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#changing-the-xsd-schemas already mentions how to build the images.

clagms:
> Yes, would be interesting to see the actual error message from @andreas-junghanns

@beutlich , I'm almost sure it was caused by @andreas-junghanns many headsets, so it's a local problem 😛 . Your script is working very well (tested it in this branch).

beutlich:
Well, still could be, that we missed to mention some execution policies for the PS script. That's why I wondered.

docs/2_2_common_schema.adoc
```diff
@@ -21,74 +21,90 @@ In the graphical representation of the schema, optional elements are marked with
 The required data types (like: `xs:normalizedString`) are defined in https://www.w3.org/TR/xmlschema-2/[XML Schema Part 2: Datatypes Second Edition].
 The types used in the FMI schema files are:
 
-[cols="1,3,1"]
+[cols="1,3,1,1"]
 |====
 |XML
 |Description (http://www.w3.org/TR/xmlschema-2/)
 |Mapping to C
+|Mapping to FMI 3.0
```
pmai, line 29: Probably the FMI 3.0 column should be before the C column?

andreas-junghanns:
> Well, still could be, that we missed to mention some execution policies for the PS script. That's why I wondered.

I had issues with Windows policies, we might want to let people know what to do.
But also, the download failed with strange errors. I downloaded manually, then all was fine.

chrbertsch:
> > Well, still could be, that we missed to mention some execution policies for the PS script. That's why I wondered.
> 
> I had issues with Windows policies, we might want to let people know what to do.
> But also, the download failed with strange errors. I downloaded manually, then all was fine.

For me it just worked fine as described in CONTRIBUTING.adoc.

PR #1053, opened 2020-06-08
Correct docs/examples for VDMCheck3

nickbattle:
I've made the minimal "obvious" changes here, but in some cases I may have changed the intent of the example unintentionally.  Please check!  But these are now "clean" according to my model.

### DIFF
--- a/docs/examples/co_simulation_early_return.xml
+++ b/docs/examples/co_simulation_early_return.xml
@@ -11,6 +11,27 @@
     modelIdentifier="MyLibrary_SpringMassDamper"
     canHandleVariableCommunicationStepSize="true"
     canReturnEarlyAfterIntermediateUpdate="true"/>
+  <UnitDefinitions>
+    <Unit name="rad">
+      <BaseUnit rad="1"/>
+      <DisplayUnit name="deg" factor="57.2957795130823"/>
+    </Unit>
+    <Unit name="rad/s">
+      <BaseUnit s="-1" rad="1"/>
+    </Unit>
+    <Unit name="kg.m2">
+      <BaseUnit kg="1" m="2"/>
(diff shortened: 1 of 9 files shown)


clagms:
@andreas-junghanns , @t-sommer , I am clearly biased, but why don't we include @nickbattle 
 check in the circleci thing? Or at least include it was a command line util that should be run whenever people change/add examples? Kind of like the XSDDiagram generator?

Had we been using it we would have caught these before they made it to the master branch...

docs/examples/co_simulation_early_return.xml
```diff
@@ -11,6 +11,27 @@
...
+    </Unit>
+    <Unit name="kg.m2">
+      <BaseUnit kg="1" m="2"/>
+    </Unit>
+    <Unit name="N.m">
+      <BaseUnit kg="1" m="2" s="-2"/>
+    </Unit>
+  </UnitDefinitions>
+  <TypeDefinitions>
+    <Float64Type name="Modelica.SIunits.Inertia" quantity="MomentOfInertia" unit="kg.m2" min="0.0"/>
```
beutlich, line 30: Modelica Standard Library version 4.0.0 renamed Modelica.SIunits to Modelica.Units.SI. Is this a real issue for importers?

nickbattle: All of the docs/examples that include Modelica named types are "Modelica.SIunits", as are all of the examples in the Standard document itself. It might be wrong, but it's consistent.

t-sommer: @beutlich, please make a PR if you want to reflect the changes in Modelica 4.0. From the FMI point of view it's just a name.

PR #1058, opened 2020-06-13
Fixes in BCS description

### LINKED ISSUE #1019: Review of  chapter 4 (BCS)
- ~~Fix formatting of
grafik
--> fixed in https://github.com/chrbertsch/fmi-standard/tree/fix-formatting-issues~~
- ~~define "hybrid ODE" in glossary~~ added to PR #941
- [x] Add to non normative text
grafik
--> mentioning one communication timestep delay for a response, also adding ME as a possibility "pure discrete" (does this exist any more?) or with clocks ...

-  [x]  the equation 
grafik
is very hard to read. Some parts are not really defined. Is "j" a derivative order? what is m_ido and m_odo (guessing input derivative order, output derivative order, but not clearly defined

- [x] "hc_i" is a very bad variable name (not compliant with typical mathematical notation), especially when it is squared 
grafik
Why not rename to H_i in the whole documentation?
Same for "tc_i" why not remae to "T_i"?
_Comment: Addresed now by a separte issue_ https://github.com/modelica/fmi-standard/issues/1050
-  [x] As we removed input derivatives, 
grafik
is no longer valid  or required!!!!  --> replace with Intermadiate Variable Inputs !!!!

- [x]  same for 
grafik

-  [x ] in 4.1.1. Initialization Mode (https://fmi-standard.org/docs/3.0-dev/#_initialization_mode_2

>  If the slave is connected in loops with other models, iterations over the FMU equations are possible. Algebraic equations are solved in this mode.

Is this really intended for BCS? Or should we move this to HCS, where we may have iterations at events anyway?

--> addressed now in a separate ticket #1055 

- [x] Change the following to BCS, as we may have loops at events of HCS
grafik

-  [x] remove 
grafik 
from "Table 2. Mathematical description of an FMU for Basic Co-Simulation." --> step mode

- [x] check "Table 2. Mathematical description of an FMU for Basic Co-Simulation." for new IVA and early return features

- [x ] Better explain and improve graphics quality of  figure 12.
-- Do the plots for the different FMUs all have the same time axis?
-- What do the green, red, dashed, non-dashed lines mean?

addressed in separate issue #1056

- [x] Should "4.1.3. Early Return from Current Communication Step" not be completely moved to https://fmi-standard.org/docs/3.0-dev/#early-return

- [x] Why do "4.2.1. Variables access in Co-Simulation interface types" and "4.2.2. Computation in Co-Simulation interface types" refer to "Co-simulation interface types", but are in "4. Basic Co-simulation
Should we move these subsections to "2.4. Common Concepts for Co-Simulation" ? Or should we move "2.4. Common Concepts for Co-Simulation" as a full chapter 4 before the BCS chapter?

- [x] in "4.2.4. Code Example for Basic Co-Simulation": remove "10. The slaves do not support asynchronous execution of fmi3DoStep."

- [x] why does the listing start with "8.?"
grafik

- [x] it would be nice to add Intermdiate input values in 4.2.4. Code Example for Basic Co-Simulation
--> singled out as separate ticket #1057 

- ~~providesAdjointDerivatives missing in table of https://fmi-standard.org/docs/3.0-dev/#basic-co-simulation-schema
--> addresssed by PR #1021~~

- ~~improve formatting of
grafik
(at least ":")--> fixed in https://github.com/chrbertsch/fmi-standard/tree/fix-formatting-issues~~

- [x] Examples: Almost identical, can we reduce to one?
4.3.1. Example XML Description File and 4.3.2. Early Return from Current Communication Step  are almost identical (the first has units defined (which is not important for BCS, the second `canReturnEarlyAfterIntermediateUpdate= true`)

chrbertsch:
Resolves #1019

### DIFF
--- a/docs/2_4_common_co-simulation.adoc
+++ b/docs/2_4_common_co-simulation.adoc
@@ -154,7 +154,36 @@ If the FMU does not support the Co-Simulation interface selected in <<fmi3Instan
 
 ==== Early Return [[early-return]]
 
-An FMU is allowed to stop the execution of <<fmi3DoStep>> and return without reaching the predefined communication time, i.e. <<currentCommunicationPoint>> `+` <<communicationStepSize>>.
+:DOSTEP: fmi3DoStep()
+:CBIU: fmi3CallbackIntermediateUpdate()
+
+//=== Improving efficiency in multi-FMU environment when asynchronous mode is used
+
+_[ In the particular context of multi-FMU architectures, significant co-simulation speed-up may be obtained if the master can avoid waiting until the end of the slowest FMU step integration._
+_If an FMU prematurely stops its current step integration computation due to an unpredictable internal event before the normal end of the step calculation, all other concurrently running FMUs may be stopped as soon as possible in order to minimize the time needed for the Co-Simulation master to resynchronize all the FMUs at the same event time._
+
+_In this context based on parallel multi-FMU calculations, <<figure-early-return>> illustrates different possibilities to synchronize FMUs at the same event time._
+
+.Different possibilities to synchronize parallel FMUs at the same event time.
+[[figure-early-return]]
+image::images/earlyReturnFigure.png[width=100%, align="center"]
+
+_Each FMU starts integration from communication point latexmath:[t_{i}] to reach the next communication point latexmath:[t_{i+1}]._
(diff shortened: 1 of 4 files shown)


chrbertsch:
Could someone please review this issue so that we could merge it soon?
This is mainly about restructuring  and removing inconsistencies.

PR #1064, opened 2020-06-17
Improve explanation of inital value positive for structural parameters

### LINKED ISSUE #1003: Why must initial dimension sizes be positive?
In https://fmi-standard.org/docs/3.0-dev/#definition-of-model-variables it is stated

> All initial dimension sizes (i.e. prior to any configuration or reconfiguration) must be positive integers (i.e. not zero), so that no dimension is initially vanished. Changes to dimension sizes are constrained by the min/max attributes of the referenced structural parameters, which can be any non-negative integer, including zero. Specifying a minimum size of zero on a structural parameter allows any related dimension sizes to be changed to zero in Configuration Mode or Reconfiguration Mode, thus causing the respective array size to go to zero, which leaves the respective array variable without any active elements.

Why do we have the restriction, that no dimension may initially vanish, while later in re-configuration we allow this?

( I do not understand the argument 

>  [This allows importing tools to ignore structural parameters because that start value reflects the internal default setting of that structural parameter].

in https://fmi-standard.org/docs/3.0-dev/#structuralParameter

andreas-junghanns:
Resolves #1003

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1136,6 +1136,11 @@ These two options are mutually exclusive, i.e. for each `<Dimension>` element ei
 However different dimension sizes can be specified using different mechanisms and can have differing <<variability>> attributes.
 
 All initial dimension sizes (i.e. prior to any configuration or reconfiguration) must be positive integers (i.e. not zero), so that no dimension is initially vanished.
+
+_[This allows importing tools to ignore <<structuralParameter,`structural parameters`>> because that <<start>> value reflects the internal default setting of that <<structuralParameter,`structural parameter`>>._
+_The rationale for requiring positive start values for <<structuralParameter, `structural parameters`>> that this avoids importers having to deal with vanishing dimensions if they do not want to deal with them (or even with changing sizes at all)._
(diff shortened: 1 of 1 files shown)


andreas-junghanns:
@chrbertsch : please have a look and approve!

docs/2_2_common_schema.adoc
```diff
@@ -1136,6 +1136,11 @@ These two options are mutually exclusive, i.e. for each `<Dimension>` element ei
 However different dimension sizes can be specified using different mechanisms and can have differing <<variability>> attributes.
 
 All initial dimension sizes (i.e. prior to any configuration or reconfiguration) must be positive integers (i.e. not zero), so that no dimension is initially vanished.
+
+_[This allows importing tools to ignore <<structuralParameter,`structural parameters`>> because that <<start>> value reflects the internal default setting of that <<structuralParameter,`structural parameter`>>._
+_The rationale for requiring positive start values for <<structuralParameter, `structural parameters`>> that this avoids importers having to deal with vanishing dimensions if they do not want to deal with them (or even with changing sizes at all)._
```
pmai, line 1141:
```suggestion
_The rationale for requiring positive start values for <<structuralParameter, `structural parameters`>> is that this avoids importers having to deal with vanishing dimensions if they do not want to deal with them (or even with changing sizes at all)._
```

Issue #1042, opened 2020-05-29
Scrap event information at fmiEnterEventMode


KarlWernersson:
I apoligze for missing this earlier.

So a change from FMI 2 is that you now can provide information about the event type when you enter Event mode. By the arguments inputEvent, stepEvent, rootsFound, timeEvent.

I see a couple of problems with this. 

1. Besides time event all other events can be triggered by reactions to other events. So even if you start with a time event you can easily soon have all other events active in your event iteration, that if we want this information it should be in the newDiscreteState function and not the enter Event mode. Or alteals all but time event.

2. >  [These arguments are not mutually exclusive and may all be fmi3False if the caller cannot provide this information.]

This statement makes the information useless as the FMU cannot be sure that the information is correct and therefore needs to check for all event types any way.

3. If we make it mandatory we put allot of extra strain on the master, for what gain?

4. A stepEvent could be triggered by other events but we have no way of signaling this to the master so it can not return that information to the fmu if it happens during event iteration. In fmi2 this is handled internally while calling newDiscreteStates without any problem.. 

5. Any input changes are input events during event iteration. To really optimize we should maybe specify which input it is that changers?
A more practical optimization is to avoid calling newDiscreteStates on an FMU that don't have input changes and is not currently iterating internal events. (I.e an FMU with all the event arguments false)

6. If one have to provide which state event that is active through the rootsFound event iteration would certainly need more iterations. One can look at the CoupledClutches in the modelica standard library,  (image of the model, 4 inertias conected by three clutches)
https://www.claytex.com/tech-blog/investigating-events-in-your-dymola-model/attachment/coupledclutchesdemo/
Having just one fmu at time 0.4 we first have a time event, this time event triggers 2 state events, these state events triggers a third state event. With the proposed changes the first two state events cannot be handed until you finish your time event since you don't know about them. Instead you call getEventIndicators after the newDiscretStates and then you find your two first state events, call a new call to newDiscreteState to handle your stateEvent with rootsFinde[indx]= true, get a new root with eventIndicators, call newDiscreteStates again. 
With the current fmi2 interface the time event and the first state event can be handled in the same call. One could implement the function to handle all of them with the same call. 

So i propose that we scrap these arguments.

andreas-junghanns:
Less arguments? No arguments from me!

otronarp:
I think found (at least some of) the issues that is the reason why they were added: #588 and  #589. 

[...]

PR #1067, opened 2020-06-17
Remove parameter inputEvent from fmi3EnterEventMode()

t-sommer:
see #1042

### DIFF
--- a/docs/3_2_model_exchange_api.adoc
+++ b/docs/3_2_model_exchange_api.adoc
@@ -117,7 +117,7 @@ _[If `fmi3True` would be returned in this case, an infinite event loop may occur
 
 [[nextEventTime,`nextEventTime`]]
 * argument <<nextEventTime>>, and shall call <<fmi3EnterEventMode>> at this time instant.
-If integration is stopped before <<nextEventTime>>, for example, due to a state event, the definition of <<nextEventTime>> becomes obsolete.
+If integration is stopped before <<nextEventTime>>, for example, due to a <<state event>>, the definition of <<nextEventTime>> becomes obsolete.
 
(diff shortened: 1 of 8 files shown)


andreas-junghanns:
@t-sommer : I added more xrefs to your new anchors - please check!

Also, there is a `external event` in the text - what is that?

docs/3_2_model_exchange_api.adoc
```diff
@@ -68,7 +68,7 @@ The model enters *Event Mode* from the *Continuous-Time Mode* and discrete-time
 
 The followings arguments can be set to `fmi3True` to inform the FMU why *Event Mode* was entered.
 
-_[These arguments are not mutually exclusive and may all be `fmi3False` if the caller cannot provide this information.]_
+_[These arguments are not mutually exclusive and may all be `fmi3False`, e.g. for an <<input event>>.]_
```
klausschuch, line 71: I would not mention fmi3False here at all anymore because the rootsFound argument is not a boolean but an int array.
```suggestion
_[These arguments are not mutually exclusive.]_
```

docs/3_2_model_exchange_api.adoc
```diff
@@ -68,7 +68,7 @@ The model enters *Event Mode* from the *Continuous-Time Mode* and discrete-time
 
 The followings arguments can be set to `fmi3True` to inform the FMU why *Event Mode* was entered.
```
klausschuch, line 69:
```suggestion
The followings function arguments have to be given to inform the FMU why *Event Mode* was entered.
```

pmai:
I'd like to merge this soonish (as well as #1074) soonish, to finish API-changes for the alpha.4; @klausschuch does this work for you?

PR #1072, opened 2020-06-18
Add windows docker command to CONTRIBUTING.md

klausschuch:


clagms:
@klausschuch what is the need for a separate command line for windows?
I think the two are exactly the same because I interpret `$PWD` has a value that the reader of CONTRIBUTING.md has to provide. I don't interpret it has a literal command that the reader will copy and paste...
It's fine if you can provide commands for different platforms, but that new command you provided probably doesn't work by copy past on all terminals you can use on windows (like the powershell... or a bash with git and all that stuff on it).

The downside of having multiple commands is the potential to become inconsistent later...

klausschuch:
@clagms I thought it might be a good idea to have all commands in a form that can be just copied and pasted. Especially for users that use docker or asciidoctor only for building the fmi-standard. 
(I don't use these tools often either and I had to look up the syntax each time again.)
This works quite well for https://github.com/modelica/fmi-standard.org/blob/master/CONTRIBUTING.md .
(note: in step 1 even the `cd fmi-standard.org` command is given)

However, you are right windows is not specific enough.
It could/should be "Windows cmd" (not only "Windows").
The "Linux, Mac" section could be extended to "Linux, Mac, Powershell". (or we even mention the spcific shells)

clagms:
I see what you mean @klausschuch !
I never following those so never realized they actually work by copy and paste :)

However, I think that those commands should not be used for users that build frequently.
This is because I think the docker command:
```
docker run --volume="%cd%:/srv/jekyll" -p 4000:4000 -it jekyll/jekyll jekyll serve
```
creates a docker container every time it is run. The docker container stays in your system, and needs to be explicitly removed by you!

To build and then remove the created container, you have to give it a name, and then remove that container.

If it helps, below is a small powershell script that I have on my local repo ([ignored locally](https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally) so I don't commit it).
Name it build.ps1, set the execution policy accordingly, and run it:
```
Function Check-Docker(){
  & docker version >$null 2>&1
  if (-not $?){
    Write-Output "Docker not started. Please start docker daemon."
    Exit 1
  }
}

Check-Docker

# Build adoc file
Write-Output "Build adoc file"
$BUILDERID = 'fmi-builder'
$FMIStandarDir = (Get-Location).ToString()

docker run -v ${FMIStandarDir}:/documents/ --name $BUILDERID asciidoctor/docker-asciidoctor asciidoctor --base-dir /documents/ --backend html5 --failure-level WARN --verbose docs/index.adoc
docker rm $BUILDERID
```

clagms:
Actually I just found out that there's a flag that automatically removes the container once its job is done:
https://docs.docker.com/engine/reference/run/#clean-up---rm

I'll take this into account in the future.

pmai:
I think an even better approach would be to copy over the commands from the circle-ci config to the document, which directly run the proper HTML build. This would ensure that what the authors use corresponds to our checks.  This can be done for Win/cmd and unixoid systems.

pmai:
I.e. something like this. What do you think?

clagms:
None on my side @andreas-junghanns

Issue #1068, opened 2020-06-17
Clarify input events and intermediate input values


CSchulzeTLK:
I see an important use case in providing smooth input functions for CoSimulation FMUs in a way, which does not reset the internal solver step size at communication points. Simulation times using BDF solvers strongly depend on the communication step size in CS mode using FMI 1.0 and 2.0.

Ideally it would be possible to provide smooth input values during one, and over several the doStep-intervals. No input event would be triggered at the communication point. So a BDF solver doesn't have to be reset. I think it would be good to clarify this.

I remember earlier discussions, where the callback function in the IOV could be used for that. If the input callback function is used during the interval, then inputs always have to be provided that way (=must not be provided at the communication point from "outside").

Is that use case covered by "normal" CS? Is it covered by Hybrid CS?

CSchulzeTLK:
I'm missing some clarifications:
- The **master suddenly stops providing intermediate input** values after some point in (simulation) time
  - Since the u(t) function should be deterministic, at least old values should be accessible. I suggest **to "keep the old value"** (since we are doing this for basically all calls to set functions) **or forbid this**.
- Intermediate input values are provided, and additionally at the communication points the **fmi3Set{VariableType} is called with the different value provided as the intermediate input value**.
  - Examples (communication points are 2 and 3):
    - A: setInput(time=2) = 42, setIntermediateInput(time=2)=42, setIntermediateInput(time=2.5)=712, setIntermediateInput(time=3)=713, setInput(time=3)=1048
  - IMHO this **should reset the internal solver step size**
  - Inputs provided in the next doStep interval should provide the same input value at the starting communication point (because they should be deterministic)
- Intermediate input values are provided, and additionally at the communication points the **fmi3Set{VariableType} is called with the (numerically !) same value provided as the intermediate input value**.
  - Examples (communication points are 2 and 3):
    - A: setInput(time=2) = 42, setIntermediateInput(time=2)=42, setIntermediateInput(time=2.5)=712, setIntermediateInput(time=3)=713, setInput(time=3)=713
    - B: setIntermediateInput(time=2)=42, setIntermediateInput(time=2.5)=712, setIntermediateInput(time=3)=713
  - IMHO this **should NOT reset the internal solver step size**

CSchulzeTLK:
Conclusions of the Working Group Meeting:

We discussed adding the sentence:
"If the master sets an intermediate input once for a variable in a communication interval it always has set it when requested by the fmu during this interval."
[...]

PR #1047, opened 2020-05-31
Change BCS description w.r.t to IVA and add new flag in xml to indicate requirement on intermediate input smoothness

chrbertsch:
- in the mathematical model of BCS, there were many leftovers from input derivatives as noted in  #1019 
- Adreassing #1019 with an alternative formulation to #992 
- Adressing #993 
- Introducing a new flag in the xml to indicate smoothness requirements for intermediate inputs.

This is a draft PR and feedback and improvements are very welcome! Especially @antvl , @robbr48 , @MartinBenedikt, @klausschuch, @clagms  ...
 (Maintainers are also very welcome to directly commit to the fork.)

PR #1074, opened 2020-06-21
Replace flags with vr sets for intermediate update setup

pmai:
This change reflects decisions at the FMI Design Meeting 2020-06-18 to
switch the all-on/all-off flags during instantiation to be variable
sets, thereby allowing for more precise control, as well as clearer
semantics during intermediateUpdate.

(partially addresses https://github.com/modelica/fmi-standard/issues/1068 )


pmai:
> Looks very good.
> However, before merging this, we should first merge #1047 into master and into this PR (so perhaps the wording in 2_4 will have to change.)

I've looked at this, and there is only one small purely syntactic merge conflict in 2.4, which can easily be resolved on merging the PRs, so I don't think any pre-merging or order is imposed on the PRs (i.e. they can be merged in any order to master, with the conflict resolved on merging). I'll take care of this upon merging either one.

pmai:
@t-sommer does this work for you? I don't really like the argument names all that much (CamelCase and n/N and very long), but that seemed to follow the pre-existing pattern...

pmai:
Changed the function argument names to be slightly saner after short brainstorming with @t-sommer .

pmai:
@chrbertsch I think this was not intended for the schema but for the function argument name to fmi3CallbackIntermediateUpdate, right? And was it decided to rename that to intermediateVariableSetRequired or intermediateVariableSetRequested? Both appear in the ticket write-up.

I can add the function argument renaming, though this does not really affect the API (function argument names are syntactically irrellevant), to this PR.

chrbertsch:
> @chrbertsch I think this was not intended for the schema but for the function argument name to fmi3CallbackIntermediateUpdate, right? 

Yes, you are right. 

And was it decided to rename that to intermediateVariableSetRequired or intermediateVariableSetRequested? Both appear in the ticket write-up.

That confusion was un-intentionally, these names should mean the same. Personally I like intermediateVariableSetRequested better.

> I can add the function argument renaming, though this does not really affect the API (function argument names are syntactically irrellevant), to this PR.

OK, would be good.

Issue #613, opened 2019-09-06
Boundaries of simulation intervals in the co-simulation specification


sglumac:
In the section describing the basics of the mathematical model for co-simulation (4.1.1. Basics) there is a bullet list starting with the sentence: "FMI for Co-Simulation is restricted to slaves with the following properties". The 6th bullet in this list says that "Whenever the simulation in a slave is interrupted, a new time value <img src="http://chart.apis.google.com/chart?cht=tx&chl=tc_{i%2B1}, tc_i <= tc_{i%2B1} <= t_{stop}"/>, can be given to simulate the time subinterval <img src="http://chart.apis.google.com/chart?cht=tx&chl=tc_i < t <= tc_{i%2B1}"/>".

This statement may be contradictory a new time can be set as <img src="http://chart.apis.google.com/chart?cht=tx&chl=tc_{i%2B1} = tc_i"/> and latter says that the simulation interval ends with <img src="http://chart.apis.google.com/chart?cht=tx&chl=tc_{i%2B1} > tc_i"/>. If the former said "a new time value <img src="http://chart.apis.google.com/chart?cht=tx&chl=tc_i < t <= tc_{i%2B1}"/>" I think it would be ok.

Furthermore, in section 4.1.2 Mathematical Model under the equation (4.1) there is a statement "<img src="http://chart.apis.google.com/chart?cht=tx&chl=\mathbf{\Phi}_i"/> and <img src="http://chart.apis.google.com/chart?cht=tx&chl=\mathbf{\Gamma}_i"/> define the system behavior for the time interval <img src="http://chart.apis.google.com/chart?cht=tx&chl=tc_i <= t < tc_{i%2B1}"/> with <img src="http://chart.apis.google.com/chart?cht=tx&chl=tc_i = tc_0 %2B \sum_{k=0}^{i-1}hc_k"/>". I would interpret this that an FMU simulates the time subinterval <img src="http://chart.apis.google.com/chart?cht=tx&chl=tc_i <= t < tc_{i%2B1}"/> and not <img src="http://chart.apis.google.com/chart?cht=tx&chl=tc_i < t <= tc_{i%2B1}"/>.

I am interpreting this as an inconsistency in the boundary definition of the simulation subinterval. I apologize if this interpretation is due to my misunderstanding of the topic.

chrbertsch:
As no one commented on this, remove milestone 2.0.1. 
@antvl : could you please look into this and comment?

sglumac:
[...]

PR #1075, opened 2020-06-22
Fix inconsistency in math formulation of BCS: Issue 613

clagms:
Ongoing work in https://github.com/modelica/fmi-standard/issues/613
Still needs some discussion to make sure it is clear.

### DIFF
--- a/docs/1___overview.adoc
+++ b/docs/1___overview.adoc
@@ -219,7 +219,7 @@ This requires that the target operating system and processor be known.
 Once exported with binaries, the FMU can be executed by any simulator running on the target platform (provided the necessary licenses are available, if required from the model or from the used run-time libraries).
 
 Small run-time overhead::
-Communication between an FMU and a target simulator through the FMI does not introduce significant run-time overhead.
+Communication between an FMU and an importer through the FMI does not introduce significant run-time overhead.
 This can be achieved by enabling caching of the FMU outputs and by exchanging multiple quantities with one call.
 
 Small footprint::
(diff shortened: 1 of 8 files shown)


andreas-junghanns:
Are we assuming t_{i+1} must be larger than t_i? Always? I am trying to understand if an FMU could, even after an event iteration, detect yet another event at the very beginning of the next step...

andreas-junghanns:
Discussion at web design meeting: Should doStep() move the time at least infinitesimally?
Karl: For BCS early return at the same point as t_i was required because of interpolated inputs or because of violating boundaries.
Pierre: This needs clarification in the document (2.4.2 Early Return)
Klaus: You could enforce event handling from the FMU by Early Return with dt = 0.0.
Pierre: Do we want to allow an early return do to input issues - other reasons should be handled differently. If we prohibit dt = 0 does not allow returning immediately if the inputs are "wrong".
Claudio/Klaus: Isn´t it strange to reject bad inputs this way?
Pierre: Why do we have Early Return in BCS? If we know this reason, we can decide dt = 0.0.
Andreas: Internal event detected that needs to be communicated (crank-angle to raise injection computation).
TorstenS: We should merge BCS and HCS anyway: they are not really different anyway.
Pierre/Andreas: We kind of agree.
Karl: not allowing dt = 0.0 seems artificial. Unless we come forward with a good reason for not allowing dt=0.0, why restrict it.
Andreas: Outputs will have to be computed even for dt=0.0 because inputs might have changed.
Pierre: We should think this through and clarify this, and allow it if there is no good reason forbidding it is found.

Summary: With early return, dt=0.0 is possible. What are the consequences? (e.g. new outputs available!?, is this the same as super-dense time?)

Will outputs be computed?
Klaus: Why not reject bad input by returning fmi3Discard?
Karl: So my previous argument is wrong.

Use cases:
Pierre:
 1. bad inputs: Discard without output computation (no state change)
 2. react to new input: Early Return with output computation (with state change) - with dt=0.0 you can now implement direct feedthrough (as Christian raised)

Do we want to have dt=0.0 for early return of doStep()? Or do we want to play it save and not allow it for BCS. 
(To be clear: the master must call doStep() with dt > 0.0.)

andreas-junghanns:
Mini-design-conference: The master is not allowed to ask for dt=0. But the FMU might early-return with dt=0. How do we reflect both these cases in the mathematical model with respect to tc_{i} and tc_{i-1}?

andreas-junghanns:
@clagms : your PRs are getting more and more outdated - are you planning on giving them up, or will your finish them?

PR #1076, opened 2020-06-24
make independent variable mandatory in xml and add references to indepent 

### LINKED ISSUE #808: Clarify usage of independent variable
@klausschuch  will create a PR according to the discussion / poll in https://github.com/modelica/fmi-design/blob/master/Meetings/2020-02-25-FMI-Design-Berlin/minutes.adoc

chrbertsch:
Resolves #808 

Taking into account notes from the Berlin FMI Design meeting https://github.com/modelica/fmi-design/blob/master/Meetings/2020-02-25-FMI-Design-Berlin/minutes.adoc#detailed-discussions-and-polls 

Please review, comment and improve (maintainers are welcome to directly commit)

Do we need a change in an xsd making sure to have exactly one independent variable?

### DIFF
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -460,12 +460,12 @@ In case the FMU utilizes a numerical integrator with variable step size and erro
 An FMU for Co-Simulation might ignore this argument.
 
 * `startTime` and `stopTime` can be used to check whether the model is valid within the given boundaries, or to allocate the necessary memory for storing results.
-`startTime` is the <<fixed>> <<initial>> value of the <<independent>> variable.
+`startTime` is the <<fixed>> <<initial>> value of the <<independent>> variable end inherits its unit.
 +
 _[It is defined with <<causality>> = <<independent>> in the <<modelDescription.xml>>._
 _If the <<independent>> variable is `time`, `startTime` is the starting time of initialization.]_
 
-* If `stopTimeDefined == fmi3True`, then `stopTime` is the final value of the <<independent>> variable.
+* If `stopTimeDefined == fmi3True`, then `stopTime` is the final value of the <<independent>> variable end inherits its unit.
 If the environment tries to compute past `stopTime`, the FMU has to return <<fmi3Error,`fmi3Status == fmi3Error`>>.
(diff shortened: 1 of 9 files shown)


pmai:
So I was not present at the Berlin meeting, and the writeup doesn't make it clear to me either: What problem is solved exactly by making the independent variable mandatory? I think most of the cleanups make sense, but they would make as much sense if a non-specified independent variable would just be assumed to exist in default form as was done in 2.0.1 (simillarly as its unit is still defaulted to seconds).

The only thing that kind of seems to be implied is that this forces FMUs that support partial derivatives (through fmi3GetDirectionalDerivative and fmi3GetAdjointDerivative), which is an optional feature not often supported esp. for CS, to support calculating derivatives w.r.t. time.

Given that making this variable exist has consequences (e.g. must be supported in fmi3GetFloat64, needs a VR, ...) beyond mere XML templating, and the majority of FMUs likely have little use for this, it strikes me as odd, so I'm probably missing something.

What am I missing here?

chrbertsch:
> Given that making this variable exist has consequences (e.g. must be supported in fmi3GetFloat64, needs a VR, ...) beyond mere XML templating, and the majority of FMUs likely have little use for this, it strikes me as odd, so I'm probably missing something.

I think the main argument was to ensure that this variable has a value reference, so that it can be accessed just like any other variable and does not need special treatment or assumptions (e.g., having an implictely defined value reference)

andreas-junghanns:
Klaus: to avoid name clashes (time, Time, TIME,...)
Claudio: be able to set displayUnit (Pierre: was possible before)
Klaus: remove implicit definitions
Pierre: this adds work to support interface - but one does not get much benefit, they rarely have derivatives or partial derivatives...
Andreas: Should we also remove the `fmi3SetTime()`?
Pierre: That was voted down.
Pierre: we need to make clear, e.g. what value do you get for GetFloat64 of time.
Andreas: Let´s have an example referenced in the standard in the Reference FMUs that show how to time is used.

andreas-junghanns:
Pierre: Make sure we note that the independent-time variable cannot be an array.

pmai:
Issue #1078 created to remind us of the more detailed spec needed.

Issue #1077, opened 2020-06-25
Merge BCS and HCS


andreas-junghanns:
TorstenS: After a number of changes to the standard, the initial differences between BCS and HCS are disappearing - except for event handling. Is it to much to asked to have event handling even in BCS and if yes, what is the remaining difference?
Karl: To do nothing in event handling is problematic.
Pierre: If the master does not do the event handling, the FMU could do it internally at the beginning of the next doStep.
Pierre: Clocks are not allowed in BCS, but they are optional in HCS anyway. Clocks could be an optional feature.
TorstenS: Agreed - some master will just not support them.
Christian: BCS is similar to FMI 2.0 - how does that change? How does this work then?
Pierre: Event handling is not the issue, clocks are - are they optional?
TorstenS: If an environment does not support the new features, some will not.
Christian: Users will be confused about what FMU interface to use if this all turns to optional features, BCS was nice and simple and guaranteed what worked.
Pierre: The feature sets (profiles) is a better way to describe which use case an FMU supports.
TorstenS: Profiles are orthogonal to the interface types.
Christian: Currently clock support was not guaranteed, event handling is?
Karl: Yes.
TorstenB: We are plenty of application where event handling is not needed, e.g. HiL simulations, or FE-tools. That means we should keep BCS.
Pierre: This proposal would allow event handling to be optional.
TorstenB: This makes it harder for importers that do not deal with event handling: The implementers of importers (and exporters) have to read more and understand what to ignore.
Pierre: We should be able to write the standard such that we clarify what to ignore if you want to fall back to BCS.
TorstenB: It is harder to ignore a feature in a merged description than having the extra chapter.
Pierre: We have to address clearly at the start what to read when you start reading the standard.
TorstenB: What does optional event handling mean for an FMU? Is there a capability flag?
Karl: The FMU can notice the absence of the event-handling call and then internally do the "right thing".
Claudio: Will I understand can exchange events? Or if they have empty functions?
Karl: And the direct feedthrough will then be handled - the function must be optional.
TorstenS: But we are going through eventMode in HCS - so the functions must be present.
Pierre: we should decide today, if this is a worthwhile undertaking and not waste time with creating a proposal. The remaining differences are - as of today´s understanding -  clocks and event handling. I would agree that this is worthwhile trying.

Poll: Do you agree we should try to merge BCS into HCS?
In Favour: Andreas, Karl, Klaus, Pierre, TorstenS
Against: TorstenB
[...]

PR #1086, opened 2020-07-06
Merge BCS and HCS

andreas-junghanns:
a lot still to be done towards #1077


andreas-junghanns:
Torstens: Can you both have a look? Now that I see the change, I think merging HCS and BCS is a must: the text simplifies, the understanding is greatly improved, much of the strange differentiation between the different CS types simply goes away.

chrbertsch:
Now that BCS and HCS are called "Co-Simulation", I consider it strange that we talk about "Scheduled Co-Simulation" (abbreviations CS and SCS). Shouldn't we call it differently  (also with a two-letter abbreviation)?
The two-letter abbreviation of "Scheduled Simulation" is not good, so what about "partitioned simulation" (PS), as the "activate model partition" is the fundamental method of this kind of FMI?
Then we would have abbreviations ME, CS and PS.

t-sommer:
I would still prefer "Scheduled Execution" (SE), as it was call originally.

andreas-junghanns:
Web design meeting:
AndreasJ: I like the document improvement: it is simpler and easier to read, in general much better. But: Is something lost in functionality that we wanted to add for HCS?
Result: all agree this merge is better and doable. Improvement suggestion:
Clarify:
 - FMU must be able to signal about not caring about events (do not call enterEventMode)
 - event mode immediately after initialization (and after ReconfigurationMode, see next) with Early Return from doStep()
    Early Return after dt==0.0 is possible. The master should call with dt > 0.0.
 - ReconfigurationMode will be reachable directly from StepMode.
Masoud: now the ME and CS state machines are not symmetric anymore and wrapping a ME in CS will be hard.
Pierre: But the simplification is worth the change in CS, preferred over symmetry. But even if we have it asymetric, it can be handled by auto-transitions.
Andreas: Do we want to keep ME 3.0 similar to ME 2.0?
Pierre: We should have this discussion about state transitions in a separate issue.
All: Agree that this merge is a step forward and should be followed as outlined above.

andreas-junghanns:
I addressed all open point we discussed at the meeting today.

Please someone else check the changes and overall state.

Missing:
 - xsd images (Torsten wanted to fix a few other things before this makes sense)
 - We still pull a code snippet with s1_fmi3InstantiateBasicCoSimulation from ReferenceFMUs/examples/co-simulation.c

t-sommer:
These should not hold us back
 
> * xsd images (Torsten wanted to fix a few other things before this makes sense)

Is pure XSD cosmetics.

> * We still pull a code snippet with s1_fmi3InstantiateBasicCoSimulation from ReferenceFMUs/examples/co-simulation.c

Will be updated once this PR is merged.

andreas-junghanns:
> Small suggestion: shouldn't we in the overview table 1 of 1.2.4. mark clocks (and additionally events) in CS as optional?

Currently, "optional" is not communicated at all (just available or not). We could extend the cells to also allow for "optional" entries.

Another issue with this table is the "solver included" for Scheduled Execution: Strictly this is not true in two ways: "Solver" for controller code has no meaning and should be called "scheduler". And the scheduler is clearly NOT included, that is the feature.

Issue #713, opened 2020-01-14
Layered standard for automotive CAN


chrbertsch:
(I think this was discussed at several design meetings but I found no other ticket for it).

Before the release of FMI 3.0 we shall make sure that future layered standard can build on the terminals an features to define layered standards e.g. for 

- standardization of terminals of certain domains
- bus simulation for automotive buses, virtual ECUs
- ...

chrbertsch:
@pmai : could you please comment on this as you have ideas regarding layered standards

andreas-junghanns:
A mental exercise is certainly not going to give us the certainty we would like to have.
A better proposal appeared during the ProSTEP discussions w.r.t. vECUs with FMI 3.0:
If we can show in a layered standard how to describe a CAN bus interface for a vECU, such a certainty would largely be attained.

chrbertsch:
(Andreas J, Christian B)

Let's create a first a layered standard for automotive CAN.
@andreas-junghanns : I will create a draft and review with @pmai

andreas-junghanns:
Here is an initial sketch of what we need to define to get the ball rolling...

Assumptions:
 - we define standard/classic CAN (read: not CAN-FD)
 - description format is either DBC or ARXML
 - timing information comes from the definition files and we do not provide overwrite capabilities
 - message structure and scaling the same way
 - no definition file composition defined/allowed

Definition sketch:
 - in /extra/org.modelica.fmi.can_bus/ all CAN bus description files are placed, the root name of the file is the name of the bus
 - we will define a Terminal type CAN_Message that groups signals and matches by variable name
 - each CAN_Message has its own dedicated clock
 - all variables belonging to a CAN_Message are clocked variables (the output/input clock is active when a message is sent/received)
 - variable types are signed or unsigned integers with the minimal width required to accurately represent the raw signal inside the message
 - an annotation in the variable defines which bus (which description file), which message and which signal the variable represents via attributes of the new custom annotation

Open:
 ? link CAN_Messages to clock via name convention?
 ? link CAN_Messages to clock via annotation?
 ? do not link CAN_Messages to the variable clocks explicitly, but implicitly via the clock of the variables?
 ? above is a signal-view CAN; alternative, we could send a binary type with the payload of the message (I like the first better and I would not yet define the second option)
 ? how do we specify which node(s) to emulate inside the FMU?

pmai:
Some further thoughts based on or discussions:

[...]

PR #1091, opened 2020-07-28
Layered CAN

andreas-junghanns:
First start, most technical issues are in, format still unclear, example missing.
This is getting more pressing as an example for the eFMI people to write their standard layered on top of FMI 3.0.

Format: Pierre can you help?

Content: Pierre, anything important I am missing?

Example: ChristianS - I am waiting for a Terminal&Icon example to adapt manually to a simple CAN bus example. Please let me know when you have something available ASAP.

Build: TorstenS: I have added a standalone .adoc - improvement suggestions?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,9 @@ jobs:
           name: Generate HTML
           command: docker run -v ~/project/:/documents/ --name asciidoc-to-html asciidoctor/docker-asciidoctor asciidoctor --base-dir /documents/ --backend html5 --failure-level WARN --verbose docs/index.adoc
       - run:
+          name: Generate Layered CAN HTML
+          command: docker run -v ~/project/:/documents/ --name asciidoc-to-html asciidoctor/docker-asciidoctor asciidoctor --base-dir /documents/ --backend html5 --failure-level WARN --verbose docs/X_layered_CAN.adoc
+      - run:
           name: Validate HTML
           command: |
             if [ $CIRCLE_BRANCH = "master" ]; then
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
           command: docker run -v ~/project/:/documents/ --name asciidoc-to-html asciidoctor/docker-asciidoctor asciidoctor --base-dir /documents/ --backend html5 --failure-level WARN --verbose docs/index.adoc
       - run:
           name: Generate Layered CAN HTML
-          command: docker run -v ~/project/:/documents/ --name asciidoc-to-html asciidoctor/docker-asciidoctor asciidoctor --base-dir /documents/ --backend html5 --failure-level WARN --verbose docs/X_layered_CAN.adoc
+          command: docker run -v ~/project/:/documents/ asciidoctor/docker-asciidoctor asciidoctor --base-dir /documents/ --backend html5 --failure-level WARN --verbose docs/X_layered_CAN.adoc
       - run:
           name: Validate HTML
           command: |
@@ -51,7 +51,7 @@ jobs:
           name: Create archive
           command: |
             cd ~/project
-            zip fmi-standard.zip README.adoc CHANGELOG.adoc LICENSE.txt docs/index.html docs/fmi-spec.css docs/images/* docs/images/schema/* headers/*.h schema/*.xsd
+            zip fmi-standard.zip README.adoc CHANGELOG.adoc LICENSE.txt docs/index.html docs/X_layered_CAN.html docs/fmi-spec.css docs/images/* docs/images/schema/* headers/*.h schema/*.xsd
       - store_artifacts:
           path: ~/project/fmi-standard.zip
       - run:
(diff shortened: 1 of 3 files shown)


chrbertsch:
@andreas-junghanns : why not use "." as a separator instead of "::" as with structured variable names in FMUs? Could be helpful to keep tools consistent for "normal variables" and "CAN variables.

andreas-junghanns:
> @andreas-junghanns : why not use "." as a separator instead of "::" as with structured variable names in FMUs? Could be helpful to keep tools consistent for "normal variables" and "CAN variables.

Specifically for the reasons of indicating that CAN signals are NOT normal variables, a different separator might be a good idea.
Note: The DBC specification I have does not allow '.' inside a signal or message name, so in theory it would work as separator (not sure about ARXML, I did not check there).

The '.' separator indicates Modelica or C hierarchies.
Using "::" indicates better a different hierarchy: that of a communication bus.
We could define a new `variableNamingConvention = busHierarchy` that uses "::" as separator.
But unfortunately, we can only have one variable naming convention right now (attribute variableNamingConvention in <fmiModelDescription>) and we cannot mix then 'structured' and 'busHierarchy'.
We might have to use 'structuredAndBus' to indicate both are used and that both separators are in use.

Summary:
I think the strongest argument pro "::" is the simple differentiation between C/Modelica variables and bus signals that have completely different semantics (e.g. clocked variables).

chrbertsch:
We should perhaps point out explicitely that this describes an idealized bus, exposing signals as FMI Terminals/Variables and not considering real bus behaviour like

- Bus arbitration: CAN messages are sent according to priority (determined by Frame_ID).
- Potocol functions: i.e. CAN Request for retransmit is a specific protocol function.
- Buffer overflow: when an ECU receives more messages than its buffer can hold.
- Bus errors: electrical errors which cause failed message transmission.
- Bus congestion: too many CAN messages for the bandwidth of the bus.

andreas-junghanns:
> We should perhaps point out explicitely that this describes an idealized bus, exposing signals as FMI Terminals/Variables and not considering real bus behaviour like

Thanks, excellent and comprehensive list, added as proposed.

andreas-junghanns:
I have not received feedback for a while and that must mean all is well. I will merge this PR at the end of this week, unless you stop me.

chrbertsch:
@andreas-junghanns : please to not yet merge this. We just had a company-internal discussion on this and will provide additional feedback. Thanks!

chrbertsch (changes requested):
I want to provide feedback from an internal discussion with @tg9541 and @fajansen on this proposal:

We appreciate the use-case and proposal, but have two main change requests:

- we think that the name is misleading and could be misunderstood: For an "Automotive CAN Layered standard" many people would expect something different. The current approach is about "Automotive CAN to signals", and this should be reflected in the title and description. Otherwise some people would expect a "low cut", message based or binary realization of an "Automotive CAN" layered standard from the name. There could be another "Low Cut" Layered standard, and we should keep this possibility open by a suitable name of this one.
- Could the current approach be extended more generally from CAN to Automotive networks, i.e. "Automotive networks to signals"? This should be at least discussed. 

See also my comments on https://github.com/modelica/fmi-standard/issues/713

@andreas-junghanns  and other interested : we could also organize a discussion with my colleagues to further discuss this

andreas-junghanns:
>     * we think that the name is misleading and could be misunderstood: For an "Automotive CAN Layered standard" many people would expect something different. The current approach is about "Automotive CAN to signals", and this should be reflected in the title and description. Otherwise some people would expect a "low cut", message based or binary realization of an "Automotive CAN" layered standard from the name. There could be another "Low Cut" Layered standard, and we should keep this possibility open by a suitable name of this one.

I see your point, but no suggestion from you for a better name.
Please suggest better names, I am willing to consider, of course.
 
>     * Could the current approach be extended more generally from CAN to Automotive networks, i.e. "Automotive networks to signals"? This should be at least discussed.

Classic CAN shines with simplicity and this proposal was meant to be a useful and usable test in how to write a layered standard on top of FMI.
Extending to other network protocols will invite a number of difficult design decisions that I currently have no time to solve, given that the base standard text requires significant efforts to reach beta quality.
I will say more to the comments in #713.

> @andreas-junghanns and other interested : we could also organize a discussion with my colleagues to further discuss this

I will be available for such a meeting if an agenda and clear goals are presented.
A general session about what could possibly be done is currently not very helpful to move things forward - we are too busy with other parts of the standard (and my own wish list is a mile long already).

chrbertsch:
> I see your point, but no suggestion from you for a better name.
> Please suggest better names, I am willing to consider, of course.

Here are some proposals for names of this layered standard: 
* "CAN2Signals"
*  "Automotive CAN to signals mapping"
* "Automotive CAN (high cut)"
* "CAN Messages to Signals Mapping"

andreas-junghanns:
>     * "CAN2Signals"
>     * "Automotive CAN to signals mapping"
>     * "Automotive CAN (high cut)"
>     * "CAN Messages to Signals Mapping"

This is easy for me: CAN2Signals - thank you for the suggestion!
Short, precise, no special characters or spaces to confuse.
I will pretend this was the result of a vote and pick this for a title for now.

PR #1094, opened 2020-07-29
Continuity and smoothness of inputs

chrbertsch:
Adressing "Open issue: smoothness in Hybrid CS at communication points" of https://github.com/modelica/fmi-standard/issues/1068

Please comment and improve this, especially I am unsure about the smoothness recommendation.


andreas-junghanns:
Christian: Did you keep this in draft status on purpose?

chrbertsch:
> Christian: Did you keep this in draft status on purpose?

I changed it to "ready for review" now; we should get some more feedback e.g. by @CSchulzeTLK, @friedrichatgc , @TorstenBlochwitz , @antvl or others before merging

andreas-junghanns:
web design meeting:
Smoothness should be guaranteed for all Intermediate Update Mode and Communication Points. If the importer notices a non-continuous change of an input, the master must raise an event.
Markus/Masoud/Andreas: Only at events will be a discontinuous changes (only for continuous variables, changes for discrete variables indicate a discrete jump and need proper handling).
Andreas: The group discussion indicates the need of a good introductory explanation for the assumptions of change rates and discrete and continuous changes and how to deal with those.
Masoud: Discrete variables can only change at events.
Markus: In 2.0 discrete variables can change at communication points (there are no events).
Christian: In CS without events, then we should allow changes to discrete variables.
TorstenB/Andreas: If a change to a discrete variable is detected, the FMU can detect such a change and knows how to handle those - the importer does not need to communicate this as an event or does not need to be informed about this from the outside.
TorstenB: CS without events does not support super-dense time.
Christian: If you change continuous variables non-continuously without event ...  (this is undetectable!!!!)
TorstenB: ... is not necessarily an issue because it depends on the solver used inside the FMU.
Andreas: It is the importers (formerly known as master) responsibility to raise an event if a discrete change happens at a continuous input variable. If no event is raised, the FMU internal solver can (and must for lack of information) assume a continuous change...
Andreas: (note to future self) we should describe this as the smoothness assumption at the beginning in the "common concepts" section.
TorstenB: Some solvers might loose accuracy in this case. Qua

andreas-junghanns:
As discussed during the design meeting, I added a brief introduction in the common concepts section describing Smoothness, Continuity and Discontinuity assumptions.
@chrbertsch : If you can manage to approve before your vacation great, else I will find someone else to check my changes.

chrbertsch:
The text looks good, bu does the section "Smoothness, Continuity and Discontinuity" really belong to 2_1_common_API? You talk about CS only there, so I suggest to move this to the CS chapter.

andreas-junghanns:
> The text looks good, bu does the section "Smoothness, Continuity and Discontinuity" really belong to 2_1_common_API? You talk about CS only there, so I suggest to move this to the CS chapter.

Indeed, that is a better place. Since this is close to math, I added this as the last section to the Mathematical Description.

Unless you stop me in the next hours, I will merge this (premise, "clearly better than before").

docs/4_1_co-simulation_math.adoc
```diff
@@ -263,3 +263,14 @@ _In reality, every variable output argument is computed by one_ `fmi3Get{Variabl
 
 _In *Step Mode* the input arguments to_ latexmath:[\mathbf{f}_{\mathit{doStep}}] _are defined by calls to_ `fmi3Set{VariableType}` _functions._
 _The variables computed by_ latexmath:[\mathbf{f}_{\mathit{doStep}}] _can be inquired by_  `fmi3Get{VariableType}` _function calls.]_
+
+==== Smoothness, Continuity and Discontinuity
+
+Since values of variables will be set at specific communication points by the importing tool, the FMU must make assumptions about the values between these communication points.
```
chrbertsch, line 269: "Since values of variables will be set ..." this is true only for inputs, right? So shouldn't we talk about inputs instead of variables her?

docs/4_1_co-simulation_math.adoc
```diff
@@ -263,3 +263,14 @@ _In reality, every variable output argument is computed by one_ `fmi3Get{Variabl
...
+
+<<continuous, `Continuous`>> variables can be assumed to continue changing during these intervals according to smoothness and continuity assumptions (see <<recommendedIntermediateInputSmoothness>>).
```
chrbertsch, line 271: same: variables --> inputs?

chrbertsch: What does "during these intervals" mean? For intermediate update, it  has to be smooth anyway. So at communication points? And what does continuity and smoothness mean here? I think, it should also be w.r.t. intermediate updates inputs. For some kinds of co-simulation, like TLM, the master could really provide smooth inputs in that sense. For most others, not. E.g, if an extrapolation polynomial is built based on some output derivatives of a connected FMU, the intermediate value at the end of the communication interval will usully not match with the new input at the communication point. Do we expect the importer (FNAM :-) ) then always to raise an event?

<recommendedIntermediateInputSmoothness>> was originially intended for intermediate inputs only, as the name indicates

docs/4_1_co-simulation_math.adoc
andreas-junghanns: Agreed, fixed in next commit (I think).

docs/4_1_co-simulation_math.adoc
andreas-junghanns: I tried to be more clear and specific according to your comment - please check if this is ok now.

docs/4_1_co-simulation_math.adoc
```diff
@@ -266,11 +266,13 @@ _The variables computed by_ latexmath:[\mathbf{f}_{\mathit{doStep}}] _can be inq
...
+Since inputs will be set at specific communication points by the importing tool, the FMU must make assumptions about the values between these communication points, including points of intermediate updates.
 
-<<continuous, `Continuous`>> variables can be assumed to continue changing during these intervals according to smoothness and continuity assumptions (see <<recommendedIntermediateInputSmoothness>>).
-If a <<continuous>> variable changes `discontinuously` (the smoothness assumption fails at this communication point), the co-simulation algorithm must raise an event to indicate the discrete change to the FMU.
+Between communication points, even when intermediate updates are called, all changes must be assumed to be continuous.
 
-<<discrete, `Discrete`>> variables keep their values between communication points changing their values.
-Furthermore, changing <<discrete>> variables will likely require special handling within the FMU.
+<<continuous, `Continuous`>> inputs can be assumed to continue changing between communication points according to the extrapolation polynomial build with the output derivatives of the connected FMU (see <<recommendedIntermediateInputSmoothness>>).
```
chrbertsch, line 274: an extrapolation polynomial is only one possibility, the master can do other things.
I would replace 

> <<continuous, `Continuous`>> inputs can be assumed to continue changing between communication points according to the extrapolation polynomial build with the output derivatives of the connected FMU (see <<recommendedIntermediateInputSmoothness>>).
with 

<<continuous, `Continuous`>> inputs may change between communication in case of <<intermediateVariableSetRequested>>`== true`. These are provided by the the co-simulation algorithm for example by an extrapolation polynomial build with the output derivatives of the connected FMU (see <<recommendedIntermediateInputSmoothness>>).

andreas-junghanns: Sorry, I don´t understand the == true.
Could you please tell me how the first sentence ends?

chrbertsch: Oh, that disappeared at copy-paste. Was ment to be 
``
<<continuous, Continuous>> inputs may change between communication in case of <<intermediateVariableSetRequested >>`== true`. These are provided by the the co-simulation algorithm for example by an extrapolation polynomial build with the output derivatives of connecte FMUs (see <<recommendedIntermediateInputSmoothness>>).
``

andreas-junghanns: With minor corrections accepted in the text.

PR #1095, opened 2020-07-29
Layered XCP

andreas-junghanns:


pmai:
I think the general approach is a good one; prior to reviewing in detail, first some generic thoughts in the general direction of the PR:

- We might want to be more generic than XCP: Given that an FMU is a DLL/SO, direct memory access calibration could also be supported, so maybe we would want to support non-XCP based A2L files as well.

- For XCP-based A2L, especially TCP/UDP based XCP, it would be nice to be able to formulate a rule how the port number varies when multiple instances of the FMU are created: The A2L spec can only give one port, but usually you need to have separate XCP servers per instance.

andreas-junghanns:
Note to self: 
 - ???
 - Port increment in case of multiple instantiations

andreas-junghanns:
>     * We might want to be more generic than XCP: Given that an FMU is a DLL/SO, direct memory access calibration could also be supported, so maybe we would want to support non-XCP based A2L files as well.

This would be more complicated as this would require mandating the a2l to contain proper addresses (offsets to the load base, of course). It would preclude that the XCP server built in and the a2l generator contrive a virtual memory space that only custom XCP servers know how to handle. This is especially helpful when the XCP protocol is used to measure into simulation components that are not C-code/ECU based (plant models, simulation backbones etc.).
 
>     * For XCP-based A2L, especially TCP/UDP based XCP, it would be nice to be able to formulate a rule how the port number varies when multiple instances of the FMU are created: The A2L spec can only give one port, but usually you need to have separate XCP servers per instance.

That is indeed an interesting thought. I would suggest that we propose a simple auto-increment-until-port-open mechanism. Do you have something else/better in mind?

andreas-junghanns:
I have not received feedback for a while and that must mean all is well. I will merge this PR at the end of this week, unless you stop me.

PR #1097, opened 2020-07-29
disallow fmi3GetOutputDerivatives in Intermediate Update Mode

### LINKED ISSUE #1089: Shall we allow fmi3getOutputDerivatives in intermediateUpdateMode?
With https://github.com/modelica/fmi-standard/commit/b35ccb9a57cd3ad5ac175b8c1b14f1ff0630a15c  "Added fmi3GetOutputDerivatives for calling sequence requirement", which is already merged, calling fmi3getOutputDerivatives gets allowed in CS intermediateUpdateMode. 

I think this is not intended and is not reflected in our mathematical description https://fmi-standard.org/docs/3.0-dev/#math-co-simulation.
Output derivatives are typically called a communication points to construct an extrapolation polyonimial (in FMI 3.0 by the master), which should not change within a given communicaiton interval. So I do not see the use case the get intermediate outputs derivatives. And I am not sure, if every FMU could provide these, even if they can at communication points.  And I would like to avoide having an extra capability flag just for that. 

I searched quickly and could not find a referecen in the minutes, except for https://github.com/modelica/fmi-design/blob/master/Meetings/2020-04-21-22-Online-FMI-Design-Meeting/minutes.adoc: 
_"Karl: should we then rename also the outputDerivatives?
Torsten: The **derivatives refer to the communication point**. The master defines how to calculate the parameters. You could set the slope of to reach a certain end point."_

So I would forbidfmi3getOutputDerivatives in intermediate Update Mode.

@andreas-junghanns , @friedrichatgc , @CSchulzeTLK and others: could you please comment?

chrbertsch:
Resolves #1089

### DIFF
--- a/docs/4_2_co-simulation_api.adoc
+++ b/docs/4_2_co-simulation_api.adoc
@@ -558,7 +558,7 @@ If the function returns <<fmi3Fatal>> the FMU enters the terminal state.
 
 Allowed Function Calls::
 
-`fmi3Get{VariableType}`, <<fmi3GetOutputDerivatives>>::
+`fmi3Get{VariableType}`::
 If <<intermediateVariableGetAllowed,`intermediateVariableGetAllowed == fmi3True`>>, the value of intermediate variables can be retrieved.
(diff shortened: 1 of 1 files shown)


andreas-junghanns:
WEB Design Meeting: Is someone having a use case to where the OutputDerivatives are needed?

TorstenB: I will think about this and leave a note in the ticket....

PR #1099, opened 2020-07-30
Unify naming of intermediate update in  CS

chrbertsch:
... that were mentioned in the discussion today or that I have noticed to be mis-formatted
(I think there was another issue but do do not find it anymore.)


andreas-junghanns:
While checking your changes, I found that we have inconsistent naming of intermediate variable access.
It is sometimes called intermediate access, intermediate update, intermediate value access, intermediate variables and intermediate variable access.

I now think your initial IU is more appropriate and that we should change all other place to use intermediate update everywhere to keep the changes minimal (access and update seem to mean the same thing, at least to me).

chrbertsch:
> 
> 
> While checking your changes, I found that we have inconsistent naming of intermediate variable access.
> It is sometimes called intermediate access, intermediate update, intermediate value access, intermediate variables and intermediate variable access.
> 
> I now think your initial IU is more appropriate and that we should change all other place to use intermediate update everywhere to keep the changes minimal (access and update seem to mean the same thing, at least to me).

I agree with renaming! (cannot approve as I submitted this PR)

andreas-junghanns:
This was more extensive and confusing than I initially thought.
@t-sommer : please note that I renamed 2 or 3 XML attributes to make them name-consistent.
This will effect the Reference-FMUs - I partially messed with the repo already, sorry 👎 

Christian: Sorry that I undid you previous work, but I think consistency is one of the most precious things we must have in a standard to make it easy to understand. I can only hope you agree.

andreas-junghanns:
This breaks because the Reference-FMU repo must accept a PR I made.
@t-sommer : which do we merge first?

t-sommer:
> This breaks because the Reference-FMU repo must accept a PR I made.
> @t-sommer : which do we merge first?

https://github.com/modelica/Reference-FMUs/pull/51 is merged

andreas-junghanns:
@t-sommer: Please have a look at these changes: they rename XML attributes and I would like you to share the blame for this change ;)

PR #1104, opened 2020-08-01
Replace master

andreas-junghanns:


andreas-junghanns:
a few more rounds will be needed: slave, importer etc... I think I can make this a lot more consistent throughout

andreas-junghanns:
I think we are through this here. This was a lot of search-and-intelligent-replace. I am will merge this unless someone stops me within the next 24h.

PR #1106, opened 2020-08-03
Better explain and improve graphics quality of Early return graphics …

### LINKED ISSUE #1056: Better explain and improve graphics quality of Early return graphics
singled out from #1019 

grafik

-- Do the plots for the different FMUs all have the same time axis?
-- What do the green, red, dashed, non-dashed lines mean?

masoud-najafi:
Resolves #1056


andreas-junghanns:
@t-sommer : This needs to be merged in sync with the reference FMUs - please take care of this PR.

docs/2_2_common_schema.adoc
```diff
@@ -120,8 +120,6 @@ _[Note that the definition of an encoding scheme is a prerequisite in order for
 _Furthermore, note the FMI calling interface requires that strings are encoded in UTF-8._
 _Since the XML files are also required to be encoded in UTF-8, string variables need not to be transformed when reading from the XML files in to C string variables.]._
 
-The special values `NaN`, `+INF`, `-INF` for variables values are not allowed in the FMI XML files.
-
```
andreas-junghanns, line 124: This change was discussed several times and the only conclusion I can draw from the protocols is that we do not want to change this unless we have a good use case for allowing these special values for specific XML attributes, but clearly no wholesale without explaining what the semantic for each of these values for every allowed special value would be.

docs/2_2_common_schema.adoc
```diff
@@ -120,8 +120,6 @@ _[Note that the definition of an encoding scheme is a prerequisite in order for
 _Furthermore, note the FMI calling interface requires that strings are encoded in UTF-8._
 _Since the XML files are also required to be encoded in UTF-8, string variables need not to be transformed when reading from the XML files in to C string variables.]._
 
-The special values `NaN`, `+INF`, `-INF` for variables values are not allowed in the FMI XML files.
```
andreas-junghanns, line 123: ... and, I think this should be an extra PR, as it does not fit well to this PR´s title and could confuse reviewers.

t-sommer:
@masoud-najafi, please open separate PRs for the above (unreleated) changes

Issue #1000, opened 2020-05-23
Improve description of clocks


chrbertsch:
In 2.1.8.1.1 (https://fmi-standard.org/docs/3.0-dev/#clock-types-for-evaluation-of-clocked-model-partitions) the text needs improvement:

Synchronous clocks:

*  `active `and `subactive `are used in the definition of synchronous clocks, but are not defined.
see also https://github.com/modelica/fmi-standard/issues/890
Here a try:  
  -- A clock is called `active` (also denoted as the clock "ticking") when triggered from the outside for an `input clock`, or triggered from within the FMU for an `output clock`, the discrete time states associated to this clock are updated.
  -- at any other times the clock is called to be `subactive` 
* the formatting of local and inputs in 
![grafik](https://user-images.githubusercontent.com/16363372/82722156-4bf39880-9cc4-11ea-8e9b-3cce97240aa9.png) 
does not make sense. There are no `local inputs` according to the linked definitions of `local `and `input`. Shouldn't this be rewritten as u_j are either external `inputs` or any other `local `variable of the FMU that is not a state of the clock partition.
* [x] I do not understand the meaning of the `subactive` argument for `fmi3SetClock` in https://fmi-standard.org/docs/3.0-dev/#fmi-api-setting-getting-clock-activation-state 
EDIT: Addressed by: https://github.com/modelica/fmi-standard/pull/1060
* Which combinations with `values` are allowed for synchronous clocks? (Can `values[i]=fmi3ClockActive`  and `subactive[i]=fmi3True`??) Why not use the `values` array with a (not yet exsisting) value `fmi3ClockSubactive` for that purpose?
* I do not understand the non-normative text  _[...This mode can also be requested by the calling environment with fmi3SetClock, e.g. to form partial derivatives ∂y_j/∂x_(j−1).]_
* Do clock priorities make sense for synchronous clocks at all? I consider this a contradiction. We should clarify this in https://fmi-standard.org/docs/3.0-dev/#_clock_priority

Communication Point clocks:
![grafik](https://user-images.githubusercontent.com/16363372/82722913-e8b93480-9cca-11ea-83ba-562e81b94009.png)
* this is no definition. What is j and j-1? The last communication point? What is x, u, y?
* What does the sentence "Additional cases need to be handled if a clock tick is not associated to an event that requires the creation of a communication point for the model partition that generated the event." mean?
[...]

PR #1107, opened 2020-08-05
Moving Communication Point Clocks Descriptions to SE chapter

andreas-junghanns:


andreas-junghanns:
this should solve part of #1000

PR #1109, opened 2020-08-06
Remove colors from tables and simply in-table references

andreas-junghanns:


andreas-junghanns:
Note: I left (18) in to keep the apparent "symmetry" of the table - otherwise the empty cells are unique and might cause confusion. This way (18) is listed in the table of "allowed combinations".

I removed the colors and I merged the table with the default and other allowed values for initial.

PR #1111, opened 2020-08-06
Clarify restrictions on modelIdentifier

andreas-junghanns:


andreas-junghanns:
@klausschuch : unless you stop me, I will merge this in 24h

PR #1112, opened 2020-08-07
Fix mode descriptions

### LINKED ISSUE #989: Check if all API call restrictions in the different modes are contained in FMI 3.0
After merging #982 which removed the API function call restrictions in the different states (and which was fine as they were outdated), we should check that we have not missed porting still valid restrictions that were contained in FMI 2.0.1 to the new description format, see https://github.com/modelica/fmi-standard/pull/982#issuecomment-629201757 and https://github.com/modelica/fmi-standard/pull/982#issuecomment-629205060

andreas-junghanns:
fixes #989


andreas-junghanns:
@pmai , @t-sommer , @TorstenBlochwitz , @clagms : I have essentially moved text around and fixed the way the state descriptions are presented (format and removing the broken lists of forbidden functions). This should be easy to review if I introduced bugs. What is harder, is check if the descriptions themselves are correct. When giving feedback, please note to which level you checked.

PR #1113, opened 2020-08-07
Edit feature table and set anchors

andreas-junghanns:


andreas-junghanns:
I have now added explicit links to the clocks description chapters from the feature table. This should make it clear.
@clagms : Please have a look and see if this is sufficient. If not, please modify this PR.

PR #1117, opened 2020-08-10
Remove SetClock restriction of IU

andreas-junghanns:
@masoud-najafi : this is one of the fixes from the mixed PR 1106. Can you check if this was an intended removal and please explain what happens if an input clock is set in fmi3IntermediateUpdate?


andreas-junghanns:
So should I delete this pull request?
Or is removing this restriction ok, but needs a better place?
The "this function" should be changed into <<fmi3SetClock>>, right?

masoud-najafi:
Please delete this PR. The original text is correct, although a bit confusing.

Issue #1115, opened 2020-08-07
Move dependency information from <ModelStructure> to variables


t-sommer:
With the removal of alias variables in FMI 3 the dependency information can be stored directly in the model variables by moving/adding the following attributes to the model variable elements:

- `dependency`
- `dependencyKind`
- `initialUnknown`

The `<ModelStructure>` element then becomes obsolete.

TorstenBlochwitz:
Looks good to me. In this way, everything is on one place and you do not have to fiddle around with the lists in ModelStructure. We have to specify clearly to which variables which dependency is allowed to be defined. And we need to specify how unknown dependency is to be defined in contrast to "no dependency". But this is to be done anyway, independent from the way how the dependency information is stored.

TorstenBlochwitz:
Looks good to me. In this way, everything is on one place and you do not have to fiddle around with the lists in ModelStructure. We have to specify clearly to which variables which dependency is allowed to be defined. And we need to specify how unknown dependency is to be defined in contrast to "no dependency". But this is to be done anyway, independent from the way how the dependency information is stored.

t-sommer:
> We have to specify clearly to which variables which dependency is allowed to be defined. And we need to specify how unknown dependency is to be defined in contrast to "no dependency".

The way we specify the dependencies would stay the same:

- no attribute `dependencies`: variable depends on everything
- `dependencies=""`: variable depends on nothing
- `dependencies="2 3"`: variable depends on variables with VR 2 and 3

For Clocks we might to want to define a default of "no dependencies" if the attribute is missing.

andreas-junghanns:
> For Clocks we might to want to define a default of "no dependencies" if the attribute is missing.

I do not like these exceptions, unless it is clear why clocks are different. Can someone explain that, please?

Until then, I would assume:
An output clock ticking depends, just like other variables, on who-knows-what, if not specified - so you have to assume on "everything".

masoud-najafi:
Clocks are not really different. The only difference is that beside their dependencies on other KnownVariables such as regular-Inputs, we need also define their dependencies on input-Clocks. 
"everything" includes all "input clocks". That's make output clocks dependent, i.e. synchronous with input clocks.
Synchronous output clocks should remains exceptions for  few blocks, such as SubSample, SuperSample, and Conditional switchs.

TorstenBlochwitz:
I agree to Masoud: Clocks are not different to other variable. I don't see why the exception is needed.

[...]

PR #1120, opened 2020-08-10
Move dependency information to model variables

### LINKED ISSUE #1114: Add dedicated causalities for clocks, continuous states, derivatives, and event indicators
We decided to list continuous time states, derivatives and event indicators as model variables in FMI 3 in order to simplify the description of dependencies. However these variables have a special meaning and API functions to access them. In order to properly describe the way these special variables can be accessed, I propose to introduce four additional causalities (as suggested by @friedrichatgc):

- clock
- continuousState (**edit**: redundant; can be inferred from `derivative` attribute)
- derivative (**edit**: redundant; can be inferred from `derivative` attribute)
- eventIndicator

Alternatively we could add new variable types (similar to `Clock`):

- ContinuousState
- Derivative
- EventIndicator

t-sommer:
fixes #1114 and #1115


andreas-junghanns:
This looks promising, to say the least. Let´s hope someone that is using dependencies will verify that your proposal does not miss any use case.

masoud-najafi:
Currently the direct dependency of output-event(clock) wrt input-event (clocks) is not defined and is considered to be false by default. This may need to be updated and allow for some special cases a direct feed-through. I wonder if this change would allow adding output clock dependencies on input clocks (if decided)?

andreas-junghanns:
> Currently the direct dependency of output-event(clock) wrt input-event (clocks) is not defined and is considered to be false by default. This may need to be updated and allow for some special cases a direct feed-through. I wonder if this change would allow adding output clock dependencies on input clocks (if decided)?

Why do you think such a outpu-input-clock dependency cannot be defined? Where is that excluded?

masoud-najafi:
> 
> 
> > Currently the direct dependency of output-event(clock) wrt input-event (clocks) is not defined and is considered to be false by default. This may need to be updated and allow for some special cases a direct feed-through. I wonder if this change would allow adding output clock dependencies on input clocks (if decided)?
> 
> Why do you think such a outpu-input-clock dependency cannot be defined? Where is that excluded?
The input-clock and output clock variables should be treated in the same way as for regular input and output variables. I.e., the order of them may be respected in the XML (for display reasons). I did not found this in the text, may be I read it fast.
Unlike regular output variables, where dependencies on inputs and states are considered as TRUE, if "dependencies" attribute is not defined, for output clocks the default dependencies should be FALSE. 
In the next build, I will check if these can be done with recent changes in #1120

t-sommer:
Wouldn't it be sufficient to simply add `dependencies=""` to the `<Clock .../>` element?

t-sommer:
After discussing this PR with @KarlWernersson it turned out that the initial dependency information is only redundant for simple models and not allowing the re-ordering of states / derivatives and event indicators might be a burden for some tools (i.e. it's not redundant, as I previously thought). Also the continuous states cannot be derived from the `derivative` attribute, because these might point e.g. to inputs.

TorstenBlochwitz:
@t-sommer, @KarlWernerssen: I don't understand the first point of your last comment. Why is not possible to add the dependency information directly to the variables?

I understand closing #1114: some variables can have more than one meaning at the same time (an input can be a derivative, an output can be a continuous state). For that reason neither the introduction of a new causality nor a new variable type works.

PR #1122, opened 2020-08-10
Remove incomplete enumeration of legal xs:double values

### LINKED ISSUE #644: NAN, +INF, -INF, MaxDouble, MinDouble, MaxInteger, MinInteger, etc
In FMI 2.0, it is stated the following sentence in page 29:
 "The special values NAN, +INF, -INF for variables values are not allowed in the FMI xml files."
But there is no more mention about the use of these special values in the API that is generated in the FMU. These special values are in different libraries depending on different systems and from different languages (C++, Java, Python, ...) and the terminology of these symbols is not unified. For instance, NaN symbol is textually expressed in Java as NaN and in C++ as nan. Generally speaking, NaN can be textually expressed in different manners and differently codified between languages too.

So there should be a higher specification in this sense. Here are some questions that we thing should be answered:
1.	Are these special values to be used in the API?
2.	If so, what would be the convention? What is the convention in text? What is the convention in codification?
3.	If so, what happens with other special values like MaxDouble, MinDouble, MaxInteger, MinInteger?
4.	If so, why is not it possible to use them in XML? We think it should be possible to use them also in the XML IN CASE it is to be used in the API.
Our understanding according to the sentence in page 29 of the previous FMI 2.0 specification is that the spirit of the standard is to avoid the use of the special values. We are on the same line and we think that it should be also mentioned that these symbols should not be used in the API.

t-sommer:
fixes #644

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -120,8 +120,6 @@ _[Note that the definition of an encoding scheme is a prerequisite in order for 
 _Furthermore, note the FMI calling interface requires that strings are encoded in UTF-8._
 _Since the XML files are also required to be encoded in UTF-8, string variables need not to be transformed when reading from the XML files in to C string variables.]._
 
-The special values `NaN`, `+INF`, `-INF` for variables values are allowed in the FMI XML files.
-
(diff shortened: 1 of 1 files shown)


andreas-junghanns:
Masoud thought it better to explicitly say these values are allowed.
Since this is different to 1.0 and 2.0 I kind of agree that an explicit word about this is kind of in order - don´t you think?

t-sommer:
We should mention this in the release notes.

PR #1124, opened 2020-08-11
Remove TODOs

andreas-junghanns:
I removed a number of useless, outdated or paranoid TODOs from the text such that we can focus on the real TODOs.


andreas-junghanns:
@clagms : I went through the glossary. There are many TODOs and I wonder if we should have general math terms defined here that can easily be googled. The only reason I see that we have to have a term here is the case when we use a term in a special way or on one special way that requires explanation. We should not try to replace google here. Agreed? Unless you stop me in the next 24h, I will remove any term that brings up a reasonable, unique description via google.

clagms:
> @clagms : I went through the glossary. There are many TODOs and I wonder if we should have general math terms defined here that can easily be googled. The only reason I see that we have to have a term here is the case when we use a term in a special way or on one special way that requires explanation. We should not try to replace google here. Agreed? Unless you stop me in the next 24h, I will remove any term that brings up a reasonable, unique description via google.

Sounds good @andreas-junghanns! Thank you so muchg!

andreas-junghanns:
As promised, I:
 - removed TODOs in the text that are meanwhile addressed
 - removed TODOs in the appendix if we use the term in accordance with a definition easily available on google (we are not writing a simulation encyclopedia)
 - removed synonyms where it made sense to me, both from the text and from the glossary once removed from the text

I don´t think this needs a big discussion, unless you think I went beyond the reasonable (mad HIHIHI from the off).
One approval would suffice for me to merge this into master, I think.
Stop me otherwise.

clagms (commented):
All looks good. Just fixed a typo.

PR #1127, opened 2020-08-20
Move eventOccurred and clocksTicked to fmi3DoStep()

t-sommer:


andreas-junghanns (approved):
I believe this is a lot cleaner and simpler than before.
I am hoping that we will get some feedback from at least one member of the HCS group to acknowledge this change not breaking any of the intended use cases.

KarlWernersson (approved):
If i remember corectly so was one of the main reasons to have these flags in the callback to allow the master to tell the fmu to handle its events internally and continue with the simulation.
As I understand it this is anyway no longer possible so then its much more cleaner to have them as outputs of the doStep function.

MBlesken:
@andreas-junghanns @KarlWernersson As we observed in the last design meeting the changes on CallbackIntermediateUpdate do not comply with SE requirements. Since it is used at several places (e.g. the state diagrams) it might be the easiest solution to roll back your changes on the signature of  CallbackIntermediateUpdate. Would you do that?

andreas-junghanns:
@MBlesken : yes, complete rollback of this change would be a solution.

Before we do that, I would like to point out and argue for alternatives.

0. Complete rollback: re-introduce the complicated event-handling mechanism for CS.
1. We could introduce only the clocksTicked argument to fmi3IntermediateUpdateCallback and declare this only useful for SE. This would keep one callback and would leave all other changes of this ticket to keep CS event handling simple.
2. We could introduce a new callback only used for SE and only to communicate clocksTicked. Maybe even with arguments to show exactly which clock(s) ticked.

I think 1. is better if you think SE uses fmi3IntermediateUpdateCallback anyway. Otherwise 2. seems like the better choice.

Currently, the text says that fmi3IntermediateUpdateCallback is used in CS and SE.
Can you please verify that this is really a requirement and explain how this mechanism would be used for SE use cases?

MBlesken:
@andreas-junghanns I was aiming for your suggestion 1. I also thought about 2. which is tempting but seems to come with the need of changing state diagram including discussions and so on. So I think we can agree on 1. 

> Currently, the text says that fmi3IntermediateUpdateCallback is used in CS and SE.
Can you please verify that this is really a requirement and explain how this mechanism would be used for SE use cases?

At least for SE I think that the example of Section 5.2 explains how this mechanism would be used for SE use cases.

Issue #1126, opened 2020-08-19
Connection of variables with variability clock


masoud-najafi:
I wonder where in the FMU specification, it has been indicated that only output clocks of an FMU can be  connected to input clock of another FMU? I mean, we should underline somewhere that only output of variability Clock can be connected to input of variability Clock. It may be obvious, but should be mentioned somewhere.

andreas-junghanns:
Can you create a small PR with a suggestion, please?

t-sommer:
IMHO this is up to the co-simulation algorithm. As long as the calling sequence is correct it can connect whatever it wants.

chrbertsch:
Can we close this due to #1130?

masoud-najafi:
Yes.

PR #1130, opened 2020-08-21
Update 2_2_common_schema.adoc

masoud-najafi:
Connection of variables with variability clock #1126

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1195,12 +1195,16 @@ These parameters can be changed independently, unlike <<calculatedParameter,calc
 
 [[input,`input`]]
 `= input`: The variable value can be provided from another FMU.
-[[inputClock,`inputClock`]] If <<variability>> = <<clock>> the variable defines an <<inputClock, `input clock`>> that is controlled by the importer.
+[[inputClock,`inputClock`]] If <<variability>> = <<clock>> the variable defines an <<inputClock, `input clock`>> that is controlled by the importer. 
+An <<inputClock, `input clock`>>  can be left unconnected or only connected to an <<outputClock,`output clock`>>.
+
 
 [[output,`output`]]
 `= output`: The variable value can be used by another FMU.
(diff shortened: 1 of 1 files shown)


andreas-junghanns:
@masoud-najafi : Please check my change proposals.

masoud-najafi (commented):
I do not understand this "the importer generates the clock signal.".  
I understand things in this way: the importer does not generate itself any clock signal, a clock signal is generated by outputClock of an FMU and the importer just transmits it to the inputClock of anther FMU. This is very similar to the regular inputs: the importer does not generate a signal by itself.

andreas-junghanns:
> 
> 
> I do not understand this "the importer generates the clock signal.".
> I understand things in this way: the importer does not generate itself any clock signal, a clock signal is generated by outputClock of an FMU and the importer just transmits it to the inputClock of anther FMU. This is very similar to the regular inputs: the importer does not generate a signal by itself.

An importer that wants to run a single FMU must generate clocks to trigger functionality inside the FMU.
This is trivial for periodic clocks, right?
It could even do that for a-periodic clocks, asking the human operator about when to trigger such clocks.
In any case, it is not required, that another FMU is involved to trigger input clocks.

masoud-najafi:
So Actually, the importer creates a fictitious FMU with an outputClock that generates the clock ticks. We have had this kind of behavior with FMUChecker that creates input signals to the FMU out of an CSV file. 
IMHO, this is very specific to the importer, and may not bring into the standard. The FMU should stay agnostic of who is creating the clock ticks.

andreas-junghanns:
> 
> 
> So Actually, the importer creates a fictitious FMU with an outputClock that generates the clock ticks. We have had this kind of behavior with FMUChecker that creates input signals to the FMU out of an CSV file.
> IMHO, this is very specific to the importer, and may not bring into the standard. The FMU should stay agnostic of who is creating the clock ticks.

But you are proposing a text where it seems that an FMU is mandatory and write it as if no other way is allowed.
I think this must be clearly stated.

masoud-najafi:
OK. I got your point. What I am just proposing is to avoid connection between Clock and non-Clock ports. If the input-clock port is left unconnected (which is allowed in the text), then the importer can do what it wants with it, like sending periodic ticks.

PR #1152, opened 2020-08-30
Add use cases for IntermediateUpdate and links to those

andreas-junghanns:


andreas-junghanns:
@t-sommer : I added some text at the place I thought best. That was not obvious, so I might still have missed the best place to explain this. But I also added links to features in the feature table and hope we guide people quickly to the relevant topics this way. Suggestions welcome!

docs/2_1_common_api.adoc
```diff
@@ -617,16 +617,12 @@ Allowed Function Calls::
...
 
-The FMU enters this state when the FMU calls <<fmi3CallbackIntermediateUpdate>> in state *Step Mode* (CS) or in state *Clock Activation Mode* (SE).
-When the callback function returns, the FMU returns to *Step Mode* (CS) or to *Clock Activation Mode* (SE).
-
-In this state the co-simulation algorithm can retrieve information from the FMU between communication points.
 This state is only available in Co-Simulation and Scheduled Execution.
-Functions called in this state must not return <<fmi3Discard>>.
 
-The FMU enters this state by calling <<fmi3CallbackIntermediateUpdate>> within <<fmi3DoStep>> (CS) or <<fmi3ActivateModelPartition>> (SE) and leaves the state towards state *Step Mode* (CS) or *Clock Activation Mode* (SE) if the function returns <<fmi3OK>> or <<fmi3Warning>>.
+The FMU enters this state by calling <<fmi3CallbackIntermediateUpdate>> within *Step Mode* (CS) or *Clock Activation Mode* (SE) and leaves the state towards state *Step Mode* (CS) or *Clock Activation Mode* (SE) if the function returns <<fmi3OK>> or <<fmi3Warning>>.
```
t-sommer, line 622: Why do you want to change this?

docs/2_1_common_api.adoc
```diff
@@ -617,16 +617,12 @@ Allowed Function Calls::
...
 If the function returns <<fmi3Error>> the FMU enters state *Terminated*.
 If the function returns <<fmi3Fatal>> the FMU enters the terminal state.
+Functions called in this state must not return <<fmi3Discard>>.
```
t-sommer, line 625: Why not?

t-sommer (changes requested):
See my comments. The reasons for intermediate update look good.

docs/2_1_common_api.adoc
andreas-junghanns: > Why do you want to change this?

The second half of the sentence talks about returning to states. It seemed logical to also use the states as the starting point of the transition, not a "random" function call on the stack.

docs/2_1_common_api.adoc
andreas-junghanns: > Why not?

Well, it does not say - I just moved the sentence to another place and was wondering the same.
I remember we discussed this, but maybe with the simplifications of Intermediate Update, those reasons are gone?
Currently, we allow only Set/Get functions and returning fmi3Discard seems harmless to me as well.

I will remove it.

Issue #1151, opened 2020-08-30
Describe allowed functions in states


chrbertsch:
(singled out from #1016 )

Currently in  https://fmi-standard.org/docs/3.0-dev/#_state_event_mode and https://fmi-standard.org/docs/3.0-dev/#_state_continuous_time_mode a lot of allowed function calls are missing (e.g. fmi3GetXXX, fmi3SetXXX) ...

Additionally, for some states like Reconfiguration mode, the description of allowed and forbidden function calls is missing completeley

andreas-junghanns:
I have the feeling these are not the only modes, where we miss functions.

But: have you check the super states? I tried to make sure that functions callable in all sub-states of a super state are listed there.

andreas-junghanns:
BTW: We agreed to no list forbidden functions anymore - I removed all such lists a while back.

I just checked: both Configuration Mode and Reconfiguration Mode list the only call allowed (that I can think of): 

fmi3ExitConfigurationMode
    With this function call, the FMU is returned to state Instantiated.
fmi3Set{VariableType}
    Only for variables with causality = structuralParameter and variability = fixed or variability = tunable.


fmi3ExitConfigurationMode
    Returns back to Event Mode (ME), Step Mode (CS) or Clock Activation Mode (SE).
fmi3Set{VariableType}
    Only for variables with causality = structuralParameter and variability = tunable.

chrbertsch:
> BTW: We agreed to no list forbidden functions anymore - I removed all such lists a while back.

OK, you are right.

> I just checked: both Configuration Mode and Reconfiguration Mode list the only call allowed (that I can think of):

OK

But where I think the allowed functions for ContinuTimeMode and EventMode are incomplete

PR #1153, opened 2020-08-30
Fixes to 3.2 (ME API)

### LINKED ISSUE #1016: Review of 3.2 (ME API)

chrbertsch:
Resolves #1016 

except for #1151 

Strangely, in https://github.com/modelica/fmi-standard/commit/5e2b4f47a9ebabd4499962df5535a9f524489829 I had to introduce a dummy Todo line to get correct formatting. Could someone familier with adoc please check?

### DIFF
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -495,6 +495,8 @@ If `stopTimeDefined == fmi3False`, then no final value of the <<independent>> va
 +
 _[If the <<independent>> variable is `time`, `stopTime` is the stop time of the simulation.]_
 
+TODO: Remove this line. When I do this, I get warnings and a wrong anchor. What goes wrong?
+
 <<fmi3ExitInitializationMode>>::
 [[fmi3ExitInitializationMode,`fmi3ExitInitializationMode`]]
 [source, C]
--- a/docs/2_1_common_api.adoc
+++ b/docs/2_1_common_api.adoc
@@ -493,7 +493,8 @@ _If the <<independent>> variable is `time`, `startTime` is the starting time of 
 If the environment tries to compute past `stopTime`, the FMU has to return <<fmi3Error,`fmi3Status == fmi3Error`>>.
 If `stopTimeDefined == fmi3False`, then no final value of the <<independent>> variable is defined and argument `stopTime` is meaningless.
 
-<<fmi3ExitInitializationMode>>::[[fmi3ExitInitializationMode,`fmi3ExitInitializationMode`]]
+[[fmi3ExitInitializationMode,`fmi3ExitInitializationMode`]]
+<<fmi3ExitInitializationMode>>::
 [source, C]
 ----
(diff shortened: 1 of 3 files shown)


andreas-junghanns (approved):
I fixed your formatting TODO - if you agree, please merge or let me know and I will merge this.

chrbertsch:
@andreas-junghanns : 

So you had to remove the comment line 
_[If the <<independent>> variable is `time`, `stopTime` is the stop time of the simulation.]_
to get awas with the ToDo problem.

Do we no longer want this comment? What was the reason for the problem.

Could you please merge, if the rest is fine for you? Thanks!

andreas-junghanns:
I thought this was the line you wanted removed...
Your comment said "this" line - so I had to pick.
Removing the line below made less sense.
And the comment I removed adds no information, right?

I am available on the on the phone, we can discuss this quickly to resolve this.

The issue you had was that anchors have restrictions on where to place them. I looked at other anchor placements and this one seemed to have worked ok.

chrbertsch:
@andreas-junghanns : the comment can be removed indeed. I also fixed the formatting by moving the anchor above. This PR can b merged from my side now.

Issue #1129, opened 2020-08-21
Local clocks in Table-6


masoud-najafi:
In Table-6, the combination of Causality "Local" and Variability "Clock" is not allowed. Note that you may have local clocks that may not be exposed to outside. This happens especially when you export a model composed of FMUs with interconnected clock links.

andreas-junghanns:
How would such an internal clock be different from an output clock?
Why would we differentiate?

chrbertsch:
Ticket-viewing-meeting: 

Andreas: I do not understand the difference between a local clock and an output clock.
Masoud: a local clock is not exposed.
Torsten: ticking has to be treated internally. For local variables I understand the use case to get more internal information. What could one do with the information of an internal clock?
Andreas: Debugging.
Torsten: We should discourage connecting a local clock to another input clocked
Klaus: then we have local clocked variables
Pierre: Must they be local?
TorstenB: Yes
Klaus: For me the decision is too fast. What does it mean?
Torsten: I see debugging as a valid use case.
Klaus: no output variable may depend on a local clock.

@masoud-najafi I will create a PR, including the restrictions mentioned above.

masoud-najafi:
>no output variable may depend on a local clock.

@klausschuch: By "output variable", do you mean "output clock variable"? Because a regular output variable may depend on local-clock ticks, i.e., at a lock clock ticks the output may change.

PR #1162, opened 2020-09-01
Clarify usage of local clocks

masoud-najafi:
Local clocks in Table-6 #1129


andreas-junghanns:
I think the table did not reflect our latest discussion point: You cannot set output clocks. I fixed that.

I also added an xref anchor for localClock and added an explanation for that in the causality table.

@masoud-najafi : ok?

masoud-najafi:
Yes, you cannot set output-clocks and local-clocks.

andreas-junghanns (approved):

PR #1164, opened 2020-09-03
Fix Get/SetClock descriptions

andreas-junghanns:
This was not just a simple one-sentence fix...


masoud-najafi (changes requested):
At line 483:  "paeridic" should be "aperidic".
What do think about adding this sentence on Get/SetInterval(), it may be a little overkill: 
"These functions  must be called only at the first activation of periodic input/output clocks. For aperiodic clocks they must be called at least at the first activation; for subsequent activations, if not called, previous retrieved values will be used."

IZacharias (changes requested):
Why is it necessary to forbid calling fmi3GetIntervalDecimal and fmi3GetIntervalFraction for Input Clocks? Especially for Scheduled Execution we could use this for the following case: An Input Clock is linked to an Output Clock via TriggeredBy attribute, i.e. when the Output Clock ticks, fmi3ActivateModelPartition must be called for the Input Clock. With the fmi3GetIntervalXXX functions the master could ask the FMU if an immediate triggering is desired or if the Input Clock should be activated delayed.
We came across this use case in the course of discussions about packing V-ECUs into FMUs.

andreas-junghanns:
> Why is it necessary to forbid calling fmi3GetIntervalDecimal and fmi3GetIntervalFraction for Input Clocks? Especially for Scheduled Execution we could use this for the following case: An Input Clock is linked to an Output Clock via TriggeredBy attribute, i.e. when the Output Clock ticks, fmi3ActivateModelPartition must be called for the Input Clock. With the fmi3GetIntervalXXX functions the master could ask the FMU if an immediate triggering is desired or if the Input Clock should be activated delayed.
> We came across this use case in the course of discussions about packing V-ECUs into FMUs.

Then we need to fix the table in 2.1.8.5 as well.
It is clearly not necessary to forbid Get on input clocks, but I always assumed that input clocks are driven from the outside (even if via triggeredBy back-looping). Why would the FMU know better about the delay then a central scheduler? What information is inside the FMU that the scheduler does not have for such delay information?

andreas-junghanns:
> At line 483: "paeridic" should be "aperidic".
This was already fixed on my end - can you please verify?

> What do think about adding this sentence on Get/SetInterval(), it may be a little overkill:
> "These functions must be called only at the first activation of periodic input/output clocks. For aperiodic clocks they must be called at least at the first activation; for subsequent activations, if not called, previous retrieved values will be used."

Yes, I added this and clarified this for each function, please check.
But I think we cannot let people NOT call for aperiodic clocks - how else would you know when the next clock ticks?
At least for Get on output clocks... am I confused again? Sorry...

masoud-najafi:
> But I think we cannot let people NOT call for aperiodic clocks - how else would you know when the next clock ticks?
> At least for Get on output clocks... am I confused again? Sorry...

As far as there is no ambiguity, it is fine. Now it is completely clear. Thank you!

masoud-najafi (approved):

chrbertsch (approved):

andreas-junghanns:
(Note to self:) I will wait with this merge until Irina removes the last doubt about the Get on input clocks.

andreas-junghanns:
> Then we need to fix the table in 2.1.8.5 as well.
> It is clearly not necessary to forbid Get on input clocks, but I always assumed that input clocks are driven from the outside (even if via triggeredBy back-looping). Why would the FMU know better about the delay then a central scheduler? What information is inside the FMU that the scheduler does not have for such delay information?

@IZacharias : Can you please answer this so we can decide if this is something we need to support?

IZacharias:
@andreas-junghanns  We came across this use case in the course of discussions about packing V-ECUs into FMUs. I received this explanation from one of our V-ECU experts:
Some MCAL modules emulate certain timing behavior by calculating events. This calculation is based on parameters within the BSW module configurations, the current and possibly a cached time stamp and the inputs read at this time. 
Especially HW events could be calculated in advance by the V-ECU and then be made known to the simulator, so that it triggers this "event" at the calculated time by activating the V-ECU. At this time of activation the V-ECU will re-read the inputs that have to be determined at a certain time.

andreas-junghanns:
Discussion at the web-design meeting:
The use case described by @IZacharias can also be handled by an outputClock that is connected to an inputClock via triggeredBy attribute (exactly like for CS and ME) of the same FMU.
The importer needs to connect the inputClock of the FMU to the outputClock of the same FMU.
We need to clarify also that this makes only sense if both outputClock and inputClock are aperiodic clocks.
And we can then restrict the call to fmi3GetIntervalXXX() to outputClocks in general (all 3 interface types).

@IZacharias : Is this correct and complete?

andreas-junghanns:
@chrbertsch , @masoud-najafi , @IZacharias : I have added a number of other changes, moved the text to a central place and clarified what we worked on during the past weeks. Please (re-)check those changes - thank you!

masoud-najafi (approved):
Is it possible to see the rendered doc of a PR? It is hard to check the tables by looking the "files changed" tab.

Line:948: Does this sentence "It is possible to define multiple unions of  < < clock,`clocks` > >." mean:
--an output clock can be connected to multiple inputClocks. (we need this).
or
--several outputClock- InputClock independent unions may be present. (it is also needed).

Line 973: this information is also present in the dependencies attribute of the output-Clock. 
Using the Dependencies attribute is more general than using the attribute ClockReferenc. For example, it is possible to have dependency on several input clocks (clock unions). 
In any case, we should avoid to provide the same information in multiple places.

Line 1341: using the union of clocks (TriggeredBy attribute) would not contradicts this sentence. May be we can say it is an exception.

chrbertsch:
> Is it possible to see the rendered doc of a PR? It is hard to check the tables by looking the "files changed" tab.

@masoud-najafi :
Click on the "check mark" of the last commit 
grafik

Click on the Details of the CI documentation build
grafik

(after possibly login to CircleCI with your Github account), 
you can now download the generated docuemation for this commit as a zip file
grafik

andreas-junghanns:
> Line:948: Does this sentence "It is possible to define multiple unions of < < clock,`clocks` > >." mean:
> --an output clock can be connected to multiple inputClocks. (we need this).
> or
> --several outputClock- InputClock independent unions may be present. (it is also needed).

Both, because the restriction is currently only on inputClocks: they can only reference one outputClock (stated in the text).
I added a sentence to be absolutely clear.

> Line 973: this information is also present in the dependencies attribute of the output-Clock.
> Using the Dependencies attribute is more general than using the attribute ClockReferenc. For example, it is possible to have dependency on several input clocks (clock unions).

I added a sentence to indicate "other dependencies".

> In any case, we should avoid to provide the same information in multiple places.
In principle, this is true. But some of the information is hard to find and if we can have a second small hint somewhere to make sure restrictions are known, I consider this an important readability improvement.
 
> Line 1341: using the union of clocks (TriggeredBy attribute) would not contradicts this sentence. May be we can say it is an exception.

Sorry, what file? Can you quote here, please?

masoud-najafi:
> Sorry, what file? Can you quote here, please?

line 1341: 2_2_common_schema in   "It is not possible to associate more than one < < clock > > to a variable."
using the union of clocks (TriggeredBy attribute) would not contradicts this sentence. May be we can say it is an exception.

docs/2_1_common_api.adoc
```diff
@@ -885,6 +940,40 @@ _Example:_
 _A clocked PI controller can be defined as an FMU and the input and output signal of the controller can be defined to be on the same <<inputClock>>._
 _In such a case, the environment deduces the <<clock>> ticks from the components that are connected to the PI controller (e.g. a component connected to the input, or to the output may define the relevant <<clock>>).]_
 
+===== Clock Unions [[clock-relationships-for-communication-point-clocks]]
+
+It is possible to define a dependency from an <<outputClock>> to an <<inputClock>> of the same FMU using the <<triggeredBy>> attribute of <<inputClock>> variables.
+An <<outputClock,`output clock`>> and the associated <<inputClock,`input clocks`>> define a clocks union to indicate that the <<outputClock>> triggers the associated <<inputClock, `input clocks`>> at the same time instant.
+It is only meaningful to combine an <<outputClock,`output clock`>> with an <<periodic,aperiodic>> <<inputClock>>.
```
IZacharias, line 947: periodic --> aperiodic:
`It is only meaningful to combine an <<outputClock,`output clock`>> with an <<aperiodic,aperiodic>> <<inputClock>>`

docs/4_2_co-simulation_api.adoc
```diff
@@ -469,18 +469,20 @@ Once all events are handled and `newDiscreteStatesNeeded == fmi3False`, the FMU
...
 <<fmi3SetClock>>::
 For <<inputClock,`input clocks`>>, <<fmi3SetClock>> is called after entering *Event Mode* to set the activation status of <<clock,`clocks`>>.
-For both <<input>> and trigerred <<clock,`clocks`>>, this function can be called several times, only if recomputations of clock state are needed during *Event Mode*.
+This function can be called several times, only if recomputations of clock state are needed during *Event Mode*.
 
 <<fmi3GetIntervalDecimal>> & <<fmi3GetIntervalFraction>>::
-For both <<outputClock,`output clocks`>> and <<inputClock,`input clocks`>> it is allowed to call these functions during *Event Mode*.
+For <<outputClock,`output clocks`>> and <<localClock,`local clocks`>> it is allowed to call these functions during *Event Mode* and *Intermediate Update Mode*.
+These functions must be called only at the first activation of <<perodic>> <<outputClock,`output clocks`>>s.
+For <<perodic,`aperiodic`>> <<outputClock,`output clocks`>>, these functions must be called at every activation _[to inquire the next clock tick]_.
```
IZacharias, line 481: > For <<perodic,`aperiodic`>> <<outputClock,`output clocks`>>, these functions must be called at every activation _[to inquire the next clock tick]_.

Should it really be "must be called" and not "can be called"? 
We should make it clearer that the clock ticks of input clocks that are triggered by the output clock are meant here:

> _[to inquire when triggered <<inputClock,`input clocks`>> must tick ]_

masoud-najafi:
> (after possibly login to CircleCI with your Github account),
> you can now download the generated docuemation for this commit as a zip file

Thanks Christian!
Then, may be the figure  image::images/example_usecase.png at line970 of 2_1_common_api.adoc , is irrelevent of the section "2.1.8.5. Clock Unions".

Also the table in the section "2.1.8.5. Clock Unions" is not clear. As per my understanding (which may be wrong) for a periodic clock, it is the importer who schedules all periodic clock ticks at the beginning of the simulation based on periods of all periodic clocks in the model (composed of several FMUs).  In this sense, there is no difference between input-clock or output clock and  both can be aperiodic, periodic, and strictly-periodic, isn't it? So, do we need this table.

docs/4_2_co-simulation_api.adoc
andreas-junghanns: Agreed, fixed as suggested - thank you!

andreas-junghanns:
> Also the table in the section "2.1.8.5. Clock Unions" is not clear. As per my understanding (which may be wrong) for a periodic clock, it is the importer who schedules all periodic clock ticks at the beginning of the simulation based on periods of all periodic clocks in the model (composed of several FMUs). In this sense, there is no difference between input-clock or output clock and both can be aperiodic, periodic, and strictly-periodic, isn't it? So, do we need this table.

The table says that if the FMU as a (strictly) periodic input clock, it cannot be attached to an output clock.
But if you have an aperiodic input clock, you can attach that input clock to any kind of output clock.
I will add a sentence to explain the purpose of the table (as it took me a while also to understand it)

IZacharias:
@andreas-junghanns I am happy with all of your adaptions. 

But there are still inconsistencies:

In 5.2.3.3, fmi3GetIntervalDecimal and fmi3GetIntervalFraction must be added to the list of allowed function calls:

```
<<fmi3GetIntervalDecimal>> & <<fmi3GetIntervalFraction>>::
For <<outputClock,`output clocks`>> and <<localClock,`local clocks`>> it is allowed to call these functions during *Intermediate Update Mode*.
These functions can be called only at the first activation of <<perodic>> <<outputClock,`output clocks`>>.
For <<perodic,`aperiodic`>> <<outputClock,`output clocks`>>, these functions must be called at every activation _[to inquire when triggered <<inputClock,`input clocks`>> must tick]_.
```

(Same text as in Co-Simulation chapter 4.2.8 except for the mention of Event Mode. Event Mode. does not exist in Scheduled Execution.)

In the table in 2.1.8.5, calling fmi3GetIntervalDecimal and fmi3GetIntervalFraction for Output clocks in Intermediate Update Mode is still not allowed.

andreas-junghanns:
> These functions can be called only at the first activation of <<perodic>> <<outputClock,`output clocks`>>.
> For <<perodic,`aperiodic`>> <<outputClock,`output clocks`>>, these functions must be called at every activation _[to inquire when triggered <<inputClock,`input clocks`>> must tick]_.
> (Same text as in Co-Simulation chapter 4.2.8 except for the mention of Event Mode. Event Mode. does not exist in Scheduled Execution.)

Ah, yes, sorry and thank you!! Fixed.
 
> In the table in 2.1.8.5, calling fmi3GetIntervalDecimal and fmi3GetIntervalFraction for Output clocks in Intermediate Update Mode is still not allowed.

Yes, agreed as well, and here I also merged the output and local clock columns as they are identical. I think we need to be clear about the difference here for SE and the others... do you agree with the current wording?

IZacharias:
> Yes, agreed as well, and here I also merged the output and local clock columns as they are identical. I think we need to be clear about the difference here for SE and the others... do you agree with the current wording?

Yes, thank you!

IZacharias (approved):

PR #1166, opened 2020-09-08
Rename parameter "clocksTicked" to "clocksAboutToTick"

t-sommer:
and "ententOccurred" to "eventEncountered" and update documentation


andreas-junghanns (changes requested):
This reads now as if we have to call fmi3EnterEventMode() must be followed by a fmi3NewDiscreteStates() call. What happens between these two calls? fmi3SetXXX()? Can you please clarify?

t-sommer:
> This reads now as if we have to call fmi3EnterEventMode() must be followed by a fmi3NewDiscreteStates() call. What happens between these two calls? fmi3SetXXX()? Can you please clarify?

The documentation doesn't say that you cannot call functions in between.

andreas-junghanns (approved):

PR #1175, opened 2020-09-10
Reintroduce "clocksTicked" parameter to fmi3CallbackIntermediateUpdate()

andreas-junghanns:


andreas-junghanns:
@MBlesken : is my re-intro change sufficient? It is surprisingly little I had to do, which makes me wonder, what I missed.
@t-sommer : On the C-code side: ok?
All: Torsten argues for better names in #1166 - this is the time to rename clocksTicked here as well, if we have a better suggestion.

t-sommer:
Since there is no event iteration in SE,  I'd say `clocksTicked` is correct here.

t-sommer:
I'm just wondering, why we have to signal clock ticks, but not events in the callback?

andreas-junghanns:
> I'm just wondering, why we have to signal clock ticks, but not events in the callback?

In SE, all events are clocks.

andreas-junghanns:
> 
> 
> > I'm just wondering, why we have to signal clock ticks, but not events in the callback?
> 
> In SE, all events are clocks.

@IZacharias , @MBlesken : please correct me if I am wrong...

IZacharias (approved):

MBlesken (approved):

PR #1177, opened 2020-09-16
Layered network with binary message data

andreas-junghanns:


andreas-junghanns:
@pmai : is this already all that is required? Please give me feedback, because if this is already a solution, I would drop the intermediate step you can see in #1178.

pmai:
I think that it basically is as easy as this, i.e. instead of actual signals, a full frame binary variable would be used. However I think the current approach is not quite right:

- I think the use of a naming convention for the raw message seems a bit non-ideal; but if it is to be used, then I think the suffix should not be `_DATA`, which would suggest that it only contains the data portion of the frame, maybe something more like `_FRAME` seems preferable.
- I think if we allow both raw message and signal variables to be used at the same time, some guarantees must be written to ensure that at least the two overlapping parts are always filled and always have identical values, and/or we must have a mechanism to otherwise ensure that it is clear which is being used.
- It should be clarified that the full frame binary variable can also be used without the signals part, if only full frame data shall be exchanged.
- I think it must be clarified which part of the CAN frame is contained in the binary variable, was well as provide a mapping to bytes: I would suggest everything from SOF until EOF inclusive (i.e. without IFS and prior to bit-stuffing).

andreas-junghanns:
@pmai, @chrbertsch : I followed some suggestions I got and more:
 - I now dropped the restriction to CAN (adding LIN, CAN FD, CAN EX, Flexray, at least).
 - _DATA and _FRAME are now allowed to allow two different levels of abstraction.
 - PDU terminals are now added as intermediate to reflect Frame structures, including a rule for synthetic names in case the PDU layer is not part of the network technology represented (simple CAN and LIN).

Missing:
 - We need a better name for the standard now (if this stays).
 - We need text to declare inconsistency of binary and value representations (at least "undefined").
 - Examples.

andreas-junghanns:
General question: should we not merge this into master? These documents are completely parallel and isolated and could be dropped at any time, if we do not want to release them with 3.0.

SebastianPanek:
Hello,

I would like to join this discussion. I'm not an expert in FMI, but I have been using virtual networks in SiL simulations for two years now and have some very concrete ideas and requirements with respect to them. I think some important things have not been discussed yet but also need a lot of attention:

Assumptions:

- Let's assume, a FMI 3.0 model wants to use CAN/FlexRay/LIN/Automotive Ethernet ports and to communicate via virtual buses/networks with other models.

- It can open any number of such ports and use them simultaneously  - the number is not limited.

- Networks and bus ports are per definition bi-directional - unlike signal ports which are either in or out

- Networks and bus ports must be opened before they can be used and shall be closed when they are not used 

- Network ports need to be configured - Automotive Ethernet ports require a primary MAC address, possibly several multicast MAC addresses, 0 or many VLAN tags in case of 802.3Q extended frames

- Multicast MAC addresses can be assigned to a Ethernet port dynamically at runtime

Questions to be answered:

1. What is the right communication layer to communicate with a network/bus port? Lower cut would mean ISO/OSI layer 2. Layer 1 is inappropriate because simulating voltage levels on the wire would be too detailed and too slow. However, it might be desired in special cases. Layer 3 or higher would be too high because these layers are typically not present on CAN buses.

2. What to pass to a network / bus port when sending data? A raw frame including header, payload and trailer? Or only payload + frame identity and address information? In my opinion the latter answer is the right one: payload + frame identity and/or addresses.

3. How to configure a network/bus port? How to pass it the necessary information to make it operable?

4. How to query the state and errors of a network/bus port?

andreas-junghanns:
@chrbertsch : I have significantly changed (improved?) this proposal according to what we discussed last week. Please have a look - I would appriciate feedback both on technical issues as well as the didactic approach.

I will continue improving things incrementally but the main components and the main flow should be stable now.

t-sommer:
Please include your XML examples from separate files and add them to the validation script.

andreas-junghanns:
> 
> 
> Please include your XML examples from separate files and add them to the validation script.

But those "files" are only snippets and not complete - how would that work?

t-sommer:
> But those "files" are only snippets and not complete - how would that work?

You can still include them form a C file that can be compiled. It doesn't have to do anything useful, but the syntax will be validated.

andreas-junghanns:
> > But those "files" are only snippets and not complete - how would that work?
> You can still include them form a C file that can be compiled. It doesn't have to do anything useful, but the syntax will be validated.
Done: Now there are two .xml files - is this what we need for the validation? How does the machine know which XSD to use?

andreas-junghanns:
This was moved to github.com/modelica/fmi-ls-bus

PR #1178, opened 2020-09-16
CAN2Signals as layered standard

andreas-junghanns:
I somehow broke the previous PR #1091.
This is the same again, sorry.


andreas-junghanns:
This proposal might be superseded by #1177. Otherwise this waits for further feedback.

andreas-junghanns:
Discussion with @pmai : We will push for the version with binary variables. We can revert to this if we find hard issues there.

PR #1190, opened 2020-09-21
Fix triggeredBy

andreas-junghanns:
... to allow also in ME and CS by removing restriction to Communication Point Clocks.


andreas-junghanns:
@pmai : This #PR should be merged only after #1182 to simplify things (I did not reset to master when I started these changes, sorry). Other than that, it is trivial.

pmai (approved):

PR #1191, opened 2020-09-21
Explain local variables

andreas-junghanns:
This still misses:
 - addition of ClockedState to ModelStructure (to list dependencies)
 - renaming of Derivative to StateDerivative (to differentiate to normal variable derivatives)

This should be discussed tomorrow at the design meeting.

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1212,7 +1212,7 @@ They can be continuous-time or discrete <<state,`states`>>, their derivatives, c
 Some local variables are listed in the `<fmiModelDescription><ModelStructure>`, if they are a `Derivative`, `EventIndicator` or `ClockedState`.
 During initialization: you can ignore local variables only if they do not have start values and are not listed as `InitialUnknown`.
 After initialization: local variables are ignorable, if you do not want to use the structure of the model for connectivity or linearization of the FMU (or they are not listed in the `<fmiModelDescription><ModelStructure>` at all).
-You can set a local variable only if they are a state.
+You can set a local variable only if they are a state in Model Exchange.
 
 _[TODO: add ClockedState elements to ModelStructure.]_
 _[TODO: rename Derivative to StateDerivative to differentiate between derivatives of states and general derivatives of any other variable.]_
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -1205,9 +1205,20 @@ The algebraic relationship to the <<input,`inputs`>> is defined via the <<depend
 An <<outputClock,`output clock`>> can be left unconnected or can be connected to an <<inputClock, `input clock`>>, usually of another FMU (for exceptions see <<triggeredBy>>).
 
 [[local,`local`]]
-`= local`: Local variable that is calculated from other variables or is a continuous-time <<state>> or an event indicator (see <<ModelStructure>>).
+`= local`: Local variables are a subset of all internal variables.
+Local variables are listed in the `modelDescription.xml`.
+Local variables cannot be variables of any other causality (see this table cell).
+They can be continuous-time or discrete <<state,`states`>>, their derivatives, clocked states or event indicators.
+Some local variables are listed in the `<fmiModelDescription><ModelStructure>`, if they are a `Derivative`, `EventIndicator` or `ClockedState`.
+During initialization: you can ignore local variables only if they do not have start values and are not listed as `InitialUnknown`.
+After initialization: local variables are ignorable, if you do not want to use the structure of the model for connectivity or linearization of the FMU (or they are not listed in the `<fmiModelDescription><ModelStructure>` at all).
+You can set a local variable only if they are a state.
+
+_[TODO: add ClockedState elements to ModelStructure.]_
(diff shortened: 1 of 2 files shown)


docs/2_2_common_schema.adoc
```diff
@@ -1205,9 +1205,20 @@ The algebraic relationship to the <<input,`inputs`>> is defined via the <<depend
...
 [[local,`local`]]
-`= local`: Local variable that is calculated from other variables or is a continuous-time <<state>> or an event indicator (see <<ModelStructure>>).
+`= local`: Local variables are a subset of all internal variables.
+Local variables are listed in the `modelDescription.xml`.
+Local variables cannot be variables of any other causality (see this table cell).
+They can be continuous-time or discrete <<state,`states`>>, their derivatives, clocked states or event indicators.
+Some local variables are listed in the `<fmiModelDescription><ModelStructure>`, if they are a `Derivative`, `EventIndicator` or `ClockedState`.
+During initialization: you can ignore local variables only if they do not have start values and are not listed as `InitialUnknown`.
+After initialization: local variables are ignorable, if you do not want to use the structure of the model for connectivity or linearization of the FMU (or they are not listed in the `<fmiModelDescription><ModelStructure>` at all).
+You can set a local variable only if they are a state.
```
KarlWernersson, line 1215: ME only (i would also like it of CS but it was decided no in separate ticket, layered standard should be used

KarlWernersson (changes requested):
added small comment

andreas-junghanns:
web-design meeting:
Karl explained the proposals to add ClockedStates, and renaming of Derivative to StateDerivative.
Karl: a ClockedState is different from a discrete state. Discrete state can be anything in the model that changes discretely but it can be hidden.
Klaus: Setting state is unclear, other rules still apply. Needs to be clarified.
Masoud: What is the use case for ClockedStates.
Karl: not all clocked variables are ClockedStates. 
Klaus: What can we do with ClockedStates?
Karl: the entire sub-active part defines how to set these variables and what they mean.
Andreas: Do you have to list all ClockedStates?
Karl: No, you can reset only some of them. You can let the importer do more interesting stuff. You can get partial derivatives w.r.t. these states, if they are listed. You can also get the state based on the previous value of the state.
Karl: ClockedStates are discrete variables, changing when a clock ticks, like counters.
    The old version of the discretised states are different from states depending on clocks.
Klaus: Do we need to also define the "old kind" of discrete states? What if you do not have clocks?
Karl: This is only asking for states depending on clocks. But the old kind of discrete states are not generally listable...
   This list contains all discrete states that get updated when a clock ticks.
   We do not have a state causality or derivative causality - we just list them in the model structure.
    To publish this information is not requires, but it can be helpful.
Pierre: There is a difference between states and outputs. There is a difference between reconstructable and, well, state. Karl wants to expose this internal relationship at the model structure.
Karl: we do not need a complete list of ClockedStates for partial derivatives.
Pierre: this just adds model structure information.
Andreas: Are we all in agreement that this is a good extension?
 -- no opposition
Klaus: do we need to have this complete list?
Karl: no, not necessarily
Klaus: the subactive clock tick means: do not update the discrete state, just compute the non-state variables (locals, outputs...).
Karl: subactive is for iteration purposes, the states are not changed (like a counter does not increment in subactive mode).
Andreas: Are there better names for subactive? 
Pierre: we have to be clear about what can be done with these ClockedStates: Perturbations for derivatives etc.
Karl: If you do not want people to fiddle with your states, do not expose them?
Pierre: This state information itself is helpful, even if you do not change them.
Klaus: Continuous states are exposed so the solver changes them from outside.
Pierre: for the ClockedStates you can have a mix of who changes: internal and external.
Masoud: Can we have directional derivatives for clocked variables?
Klaus: yes, but it would be very messy.

Andreas: Do we agree we want this? Karl is surprised that this disappeared from the original proposal.
Karl: Even in CS we may expose continuous states to get derivatives ("do interesting stuff")
Pierre: the more information you expose the better, the more you hide, the trickier things get.
Andreas: Hiding things is important to FMI technology - the exporter should be in control and things should still work.
Pierre: the more exposed, the more you can do with this - even reverse engineering.

We all agree that state information for CS is optional, and this is the same here: ClockedStates are optional and even if listed, do not have to be complete and the importer can ignore them completely.

Masoud: subactive should work for all clocks.
Pierre: This is nice to have, but you do not need subactive.
Karl: you need to know if subactive is available for value propagation. If subactive is not supported, this needs to be known.

Andreas: Do we need a new capability flag? No: it is a clock attribute (e.g. combining FMUs requires individual control per clock).

andreas-junghanns:
Pierre: should we re-introduce "previous" again, now that we have ClockedStates?
Karl: If we do not make it required, it will be messy and less useful - I suggest to leave it out.
Klaus: What is the benefit for the importer to have "previous"?
Karl: It only looks similar to derivatives, but it is not the same.
Pierre: What guarantees do we need to have it useful compared to all the other guarantees (non-guarantees) we give for the rest of the model structure.
Karl: this only helps if you can rely on having it.
Klaus: what would be the use case of knowing the previous value?

Conclusion: no "previous". Ever!

docs/2_2_common_schema.adoc
andreas-junghanns: Added.

TorstenBlochwitz (approved):

PR #1195, opened 2020-09-22
Rename eventModeRequired to eventModeUsed

andreas-junghanns:


clagms:
@andreas-junghanns the doStep function description has to mention this parameter. It currently doesn't. It also implies that the importer has to enter event mode regardless of the value of this argument.

andreas-junghanns:
@t-sommer : Do you want to propose a better name?

t-sommer (approved):

Issue #1199, opened 2020-09-22
Multiple possible interpretations of clock API (and examples)


clagms:
- Remove clocksAboutToTick from doStep because it's redundant with eventEncountered
- If you want to know if a clock is active, use getclock functions.
- A clock is active in a superdense time instant. call to newDiscreteStates makes it tick, and become inactive.
  - same applies to subactive. If we set a clock subactive, it remains subactive until the next fmi3NewDiscreteStates
- Function fmi3EnterEventModeTYPE has extra information that is important for model exchange but seems useless in co-simulation.

Types of clocks:
- Periodic: we know when they will tick.
- APeriodic:
  - Predictable: we know when they will tick.
  - Unpredictable: we don't know when they will tick.

Masoud: A model exchange scenario with unpredictable output clocks:
- The importer finds a zero crossing event
- To be continued...

Difference between model exchange and co-simulation with synchronous clocks: in co-simulation, the importer does not know when an unpredictable clock will tick. Only the FMU knows this.

A co-simulation scenario that allows inputs to be set before a clock ticks and while it is active:
- We have an FMU with an unpredictable output clock that ticks when a crank shaft inside that FMU is on top.
- The importer calls DoStep on the FMU.
- The FMU returns early (signals an early return through eventEncountered argument of doStep function) because (internally to the FMU) crank shaft 1 (modelled inside the FMU) is on top (and that will cause output clock 1 to tick).
- Importer sends the FMU into event mode.
- The Importer check which clock is active (by using GetCLock on every output clock of the FMU).
- The importer finds out that clock 1 is active.
- Importer gives new inputs that affect the clocked partition of clock 1 
    (before calling new discrete state).
- the clock 1 has not yet ticked.
- the importer calls fmi3NewDiscreteStates() and the clock 1 ticks inside the call.

Scenario to justify need for triggeredBy.
Masoud: Output clocks must not be associated to any variables (i.e., their clocked partition must be empty). 
Piere: if you want to have a non empty clocked partition, then you must use an input clock that is triggeredBy that output clock.
Christian: if an clock is signalled as being active, there's yet no calculation been done.
Pierre: Output clocks no longer seem to fit the definition in that they are no longer attached to a clocked partition.
Andreas: where is this requirement coming from? Example: if a plant model raises a "cylinder top" event, this is something that can onyl be computed if a lot of variables changes weith this event. This is just one FMU. 
[...]

PR #1206, opened 2020-09-25
Improve synchronous clocks example

clagms:
New iteration on the example motivating and explaining Synchronous Clocks.


andreas-junghanns (approved):

andreas-junghanns:
@clagms : why is this still draft? Can we merge this?

clagms:
Because I need to clarify some of the semantics in https://github.com/modelica/fmi-standard/issues/1199 
Our "mathematical" description is not correct as @klausschuch pointed out, and I was planning to fix that.
But if you want to merge this now, we can do that later. 
It would make sense that @klausschuch has another look at how I addressed his comments...

andreas-junghanns:
> 
> 
> Because I need to clarify some of the semantics in #1199
> Our "mathematical" description is not correct as @klausschuch pointed out, and I was planning to fix that.
> But if you want to merge this now, we can do that later.
> It would make sense that @klausschuch has another look at how I addressed his comments...

Our current policy is to ask: is the standard better with after the merge?
If "yes", merge.
We need to move forward, we cannot wait until ALL related issues are fixed or we risk deadlocking all progress for fear of making incomplete improvements.

andreas-junghanns:
Did you mark all open TODOs in the text or in tickets? If yes, either of us should merge. Now.

PR #1208, opened 2020-09-25
Change freeze(d) to frozen

andreas-junghanns:


chrbertsch (approved):
We might put "frozen" in parenthesis. But also fine as it is.

chrbertsch:
> We might put "frozen" in parenthesis. But also fine as it is.

Corretion: I did noit mean paranthesis but quotation marks.

chrbertsch (approved):

PR #1217, opened 2020-10-09
Version independence

andreas-junghanns:


chrbertsch (changes requested):
"minor versions"are used but are not not defined in the current specification.
We should either define them in the specificiation or include a reference.

(The SSP spec (https://ssp-standard.org/publications/SSP10/SystemStructureAndParameterization10.pdf) contains a definition and reference to "semantic versioning in 2.2.)

docs/1___overview.adoc
```diff
@@ -222,6 +222,15 @@ To be simulator independent, the FMU must export its implementation in self-cont
 This requires that the target operating system and processor be known.
 Once exported with binaries, the FMU can be executed by any simulator running on the target platform (provided the necessary licenses are available, if required from the model or from the used run-time libraries).
 
+Version independence::
++
+--
+[horizontal]
+Produce once:: An exporter can always export an FMU only once with the most advanced FMI minor version supported (per FMI major version), because features of later minor versions are optional and ignorable by older importers.
```
chrbertsch, line 229: "can always export an FMU only once" is not clear to me.  
Do you want to express that an FMU once exported for a certain minor version will remain a valid FMU for all coming minor versions belonging to the same major version or FMI?

andreas-junghanns:
> 
> 
> "minor versions"are used but are not not defined in the current specification.
> We should either define them in the specificiation or include a reference.
> 
> (The SSP spec (https://ssp-standard.org/publications/SSP10/SystemStructureAndParameterization10.pdf) contains a definition and reference to "semantic versioning in 2.2.)

I shamelessly copied and edited the SSP text to FMI needs. Thanks for pointing this out!

docs/1___overview.adoc
andreas-junghanns: This is essentially what we are trying to say with the second point.
Is that not clearly expressed?

chrbertsch: I have difficulties understanding this sentence.

Proposal: Change
"Produce once:: An exporter can always export an FMU only once with the most advanced FMI minor version supported (per FMI major version), because features of later minor versions are optional and ignorable by older importers."
to
"Stay valid:: FMUs exported once for a given minor version of FMI (typically the latest one released a that point of time) stay valid FMUs for future minor versions of the same major version, because features of later minor versions are optional and ignorable by older importers."

andreas-junghanns: "Stay valid" is not quite covering it: we also guarantee that the FMI runs in earlier minor versions and for that direction "stay" just does not seem right.

chrbertsch: What a about "future-proof"? 
"future-proof:: FMUs exported once for a given minor version of FMI (typically the latest one released a that point of time) stay valid FMUs for future minor versions of the same major version, because features of later minor versions are optional and ignorable by older importers."

andreas-junghanns: What future? It is also "past-proof" because it runs also on earlier versions.
And for a supplier having customers ahead and behind in the version race, covering both ends is great: "run all".

(that is why I did not call them past/future, but earlier/later)

chrbertsch: What about:
"Produce once:: An exporter can export an FMU for any FMI minor version (preferably using the latest released minor or patch version); this FMU will stay a valid FMU in any previous or future minor version of corresponding major version,  because features of later minor versions are optional and ignorable by older importers."

TorstenBlochwitz: We reformulated and shortened the text.

TorstenBlochwitz (approved):

PR #1218, opened 2020-10-10
Define versioning and layered standards

### LINKED ISSUE #1216: Define "layered standard" in the FMI3 specification document.
In the current speficiation, we refer to "layered standards" in the context of the extra directory, but do not explain what they are. 
We should have description/definition in the specification text.

SSP https://ssp-standard.org/publications/SSP10/SystemStructureAndParameterization10.pdf  defines it this way
> 
> “A layered standard defines extensions to the base SSP standard by specifying either standardized annotations, standardized extra files in SSP archives, and/or support for additional MIME types/file formats, as defined in section 2.1. A layered standard can include a single or combined set of extension mechanisms from this set. The layered standard is thus considered to be layered on top of the definitions and extensions mechanisms provided by this base standard.Layered standards can fall into three categories:
> * Layered standards can be defined by third parties, making them third-party layered standards, without any representations by MAP SSP for their suitability or content, or even knowledge by MAP SSP about their existence.
> * Layered standards can be defined/adopted and published by MAP SSP itself, making them MAP SSP layered standards. 
> * Finally, layered standards that have achieved enough adoption or importance to be included into the base standard set can be incorporated into a new minor release version of the base standard as an optional or mandatory appendix, making support for this layered standard optional or required for conformance with the newly published minor release version of the base standard.”
> 

I will provide a PR until Oct 16th, shall be updated after we have defined the rules for layered standards in the steering committee.

chrbertsch:
Resolves #1216

Text was first copied from SSP https://ssp-standard.org/publications/SSP10/SystemStructureAndParameterization10.pdf (2.2.) and then adapted to FMI, you can follow the commits for the adaptations.

I think it is good to define or reference the versioning scheme within the standard.

This PR shall be perhaps combined  with #1217
And regarding layered standards adapted after definition of rules for them in the steering committee.


andreas-junghanns (approved):
Great proposal - I added an abridged version in the Properties and Guiding Ideas.
If we merge both PRs, we should forward reference your section in the Guiding Ideas.

TorstenBlochwitz (approved):

PR #1222, opened 2020-10-13
The big move

andreas-junghanns:


andreas-junghanns:
@chrbertsch : We moved quite some text around. Much of this had to be fixed to reflect the current proposal.

The idea is to accept this state, even if this is just a preliminary state of the reordering (next is to move the mathematical formulas into the State sections). But if we keep this massive change as PR, any other PR will conflict dramatically.

I can walk you through some of the changes to explain what we had in mind...

chrbertsch (approved):
This looks really good and is a significant improvement. 

Only some Minor findings, I propose to  merge fast, they could be fixed later:

- 2.2.9. Getting Number of Event Indicators: is this not speficific to ME only and thus does not belong to common concepts? Can be done later
- Getting and Setting the Complete FMU State in 2.3.1.1.:  Shouldn't this be moved out e.g. as 2.2.12; can be done later
- Why is "2.5.6.1. Supporting Multiple Interface Types in one FMU" part of 2.5.6 "extra directory"?

Issue #1180, opened 2020-09-18
Add (*clockTicked) argument to fmi3NewDiscreteStates?


masoud-najafi:
Wouldn't it be "convenient" if a new argument (fmi3Boolean *ClockTicked) is added to fmi3NewDiscreteStates? Otherwise the importer should check all output-clocks after leaving fmi3NewDiscreteStates. This is very similar to the callback fmi3CallbackIntermediateUpdate where the clocksTicked indicates that GetClock should be called by the importer.
It both cases, it is a convenient flag, not a necessary flag. Because polling all output-clocks with GetClock must provide the same information.

andreas-junghanns:
This is an output argument to inform the importer that the an FMU clock ticket.
This saves the importer a lot of GetClock() calls.

Klaus: The name should not use "tick".
Pierre: someClocksActive?

No opposition.

masoud-najafi:
With the recent changes in the way clocks are handled in FMI-3.0, I think this ticket becomes obsolete and we can close it without doing anything. Except the for the name NewDiscreteState , where fmi3DoDiscreteStatesUpdate becomes more meaningful.
@clagms , What do you think about this ticket?

clagms:
Agreed @masoud-najafi !

andreas-junghanns:
IF we change this, I suggest to use fmi3UpdateDescreteStates().

masoud-najafi:
OK, I will rename them to fmi3UpdateDiscreteStates.

clagms:
> @clagms
> I am useless in doing PR, after my previous PRs, Torsten Sommer got mad on me :)
> So could I ask you for making a PR for renaming fmi3NewDiscreteState to fmi3DoDiscreteStatesUpdate, or learn me how to do it?

Sorry the delay @masoud-najafi ! I see that you've already done quite well 😄

bernaloo:
@masoud-najafi I can do that :)

PR #1224, opened 2020-10-16
#1180 : rename fmi3NewDiscreteStates to fmi3UpdateDiscreteStates

masoud-najafi:
Following the decision on renaming fmi3NewDiscreteStates  to fmi3UpdateDiscreteStates (https://github.com/modelica/fmi-standard/issues/1180) , I renamed all occurrences of fmi3NewDiscreteStates.
We need also update the APIs in reference FMUs.
CC: @t-sommer


chrbertsch (approved):

andreas-junghanns (changes requested):
`newDiscreteStatesNeeded` is still around - should we not also change that argument accordingly?

masoud-najafi:
> 
> 
> `newDiscreteStatesNeeded` is still around - should we not also change that argument accordingly?

If "updateDiscreteStatesNeeded" is fine, I will replace them all.

chrbertsch:
Perhaps  add "of": updateOfDiscreteStatesNeeded"?

andreas-junghanns:
> 
> 
> Perhaps add "of": updateOfDiscreteStatesNeeded"?

I think without "of" is better, because updateDiscreteStates is the function call, updateDiscreteStatesNeeded requires another call to that function. And it would be shorter too.

headers/fmi3FunctionTypes.h
```diff
@@ -488,15 +488,15 @@ typedef fmi3Status fmi3SetIntervalFractionTYPE(fmi3Instance instance,
                                                size_t nValues);
 /* end::SetIntervalFraction[] */
 
-/* tag::NewDiscreteStates[] */
-typedef fmi3Status fmi3NewDiscreteStatesTYPE(fmi3Instance instance,
-                                             fmi3Boolean *newDiscreteStatesNeeded,
+/* tag::UpdateDiscreteStates[] */
+typedef fmi3Status fmi3UpdateDiscreteStatesTYPE(fmi3Instance instance,
```
t-sommer, line 492: Shouldn't this be `discreteStatesNeedUpdate` instead of `updateDiscreteStatesNeeded`?

t-sommer (changes requested):

t-sommer (approved):

chrbertsch (approved):

andreas-junghanns (approved):

Issue #1049, opened 2020-06-06
Reverse order of attributes in pictures generated from the xsd files


chrbertsch:
In the pictures generated from the xsd files, the attributes appear in the opposite order as they are defined in the xsd files and are listed in the tables of the scheme description of the individual FMI types. 
![image](https://user-images.githubusercontent.com/16363372/83940511-fbf8f380-a7e4-11ea-9fe2-01d1518c83b5.png)

It would be nice, if we could change this, so that, e.g., also the mandatory attribute `modelIdentifier`  would be listed as first attribute. Is there a possibility to do this?

andreas-junghanns:
Pierre: we need to do this just before the release. If we cannot find a way to have the tool behave as we expect, we can reverse the  XSD elements to make the result what we expect.

chrbertsch:
@clagms and @pmai : Do you see a chance to fix this in the tooling? Or would you recommend to do the reordering in the xsd-files? Or will this have side-effects? This is not the most urgend ticket, but I would like to get it out of the way.

clagms:
I've updated the pre-release to fix this:
https://github.com/clagms/xsddiagram/releases/tag/XSDDiagram_1.2_FMI

The following PR updates the pngs:
https://github.com/modelica/fmi-standard/pull/1225

@chrbertsch please confirm that the generated figures are in the right order. I've confirmed this for the fmiModelDescription png file.
The code is the same for all the other files, and I've not noticed any errors in the export.

clagms:
@chrbertsch if you have used the script `generate_figures.ps1` before, and you wish to regenerate the pngs on your machine, make sure to delete the folder `xsddiagram` where the old files downloaded from the release are. 
This will cause the script to download the new version.

PR #1225, opened 2020-10-18
Generate pngs using attributes in correct order.

clagms:
Fixes https://github.com/modelica/fmi-standard/issues/1049


chrbertsch (approved):
Looks very good!

PR #1234, opened 2020-10-20
Add inverse display units and improve text

### LINKED ISSUE #1220: Allow inverse display units
Example: `mpg` (miles per gallon) as display unit for `L/100km` (liters per 100 km)

```xml
<UnitDefinitions>
  <Unit name="L/100km">
    <BaseUnit m="2" factor="1e-8"/>
    <DisplayUnit name="mpg" invert="true" factor="235.214583"/>
  </Unit>
</UnitDefinitions>
```

The display value `v_display` would be converted from the base value `v_base` as

```
v_display = factor * (1/v_base)
```

if `invert == true`.

t-sommer:
and formulas in section "Definition of Units"

fixes #1220

### DIFF
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -323,40 +323,43 @@ In this section, the units of the variables are defined.
 
 _[Unit support is important for technical systems since otherwise it is very easy for errors to occur._
 _Unit handling is a difficult topic, and there seems to be no method available that is really satisfactory for all applications, such as unit check, unit conversion, unit propagation or dimensional analysis._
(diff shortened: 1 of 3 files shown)


andreas-junghanns (approved):
I am not totally sold on the `inverse` idea, but can see that this is a way to keep units that "belong together" can stay together this way (and would otherwise require a different base unit).

chrbertsch (approved):
(I still do not like this feature, but following the discussion https://github.com/modelica/fmi-standard/issues/1220#issuecomment-712850562 and as no one joined me in my fundamental opposion :-)  let's move on )

I think the proposed changes are OK.

PR #1240, opened 2020-10-26
Serialization of array variables

### LINKED ISSUE #1187: Describe serialization of array only once and link everywhere else
From web-design meeting 21.09.2020.

klausschuch:
This PR should fix #1187 and partly #845 (the handling of string array variables is missing).

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -708,9 +708,7 @@ include::../headers/fmi3FunctionTypes.h[tag=SetClock]
 
 Sets <<clock,`clocks`>> activation status by providing the value references of the corresponding <<clock>> variables and their values.
 Note that <<clock>> variables, like any other variables can be scalar or array variables.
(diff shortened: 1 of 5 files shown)


klausschuch:
While checking the missing part about start elements for arrays of type fmi3String, I stumbled upon the missing minOccurs attribute. The default value of minOccurs is 1. Hence, there must be at least one start element element for each variable of type fmi3String.
This is problematic because there might be output strings without a start start string.

andreas-junghanns:
> 
> 
> While checking the missing part about start elements for arrays of type fmi3String, I stumbled upon the missing minOccurs attribute. The default value of minOccurs is 1. Hence, there must be at least one start element element for each variable of type fmi3String.
> This is problematic because there might be output strings without a start start string.

For fmi3Strings, the Start value is an element, for other variable types, it is an attribute. Why?
For the minOccurs: Can´t we simply add it with 0?

t-sommer:
> For fmi3Strings, the Start value is an element, for other variable types, it is an attribute. Why?

I think this is because Strings can contain spaces (unlike the other literals), so we can't use them as a separator (hence the separate elements).

klausschuch:
Created PR #1241 to fix the xsd.

andreas-junghanns (approved):

PR #1241, opened 2020-10-27
Allow String variables without Start elements

klausschuch:
The `start `attribute for strings became an element (`Start`) in this PR #523. This is fine and was discussed several times.
However, in the current xsd the `Start `element is mandatory.
Currently xsd validation fails for a String variable like this:

```
<String name="StringOut" valueReference="13" causality="output" variability="tunable">
    <Dimension valueReference="0"/>
    <Dimension valueReference="1"/>
</String>
```

This PR fixes the xsd to allow such `String` variables without `Start` elements.


andreas-junghanns:
@pmai or @t-sommer : Can either of you please approve? I am not sure because of the sequence element...

t-sommer (approved):

PR #1244, opened 2020-10-29
Allow setting input start values before fmi2EnterInitializationMode (#369)

### LINKED ISSUE #369: Setting input start values before fmi2EnterInitializationMode
**Reported by karl&#x2e;wernersson on 16 Feb 2016 13:47 UTC**
Reference tickets:
https://trac.fmi-standard.org/ticket/258
https://trac.fmi-standard.org/ticket/271
Web meeting Minutes  2014-07-11

Hi
In ticket #271 input lost the initial attribute approx,
the reason for this was the ticket #258 and the fowling discussion about it during the web meeting 2014-07-11.

One effect of this change was that start values for input is no longer allowed to be changed from default value before call to fmi2EnterIntializtaionMode as of specification page 48

**fmi2SetXXX can be called on any variable with variability ≠ "constant" before initialization
(before calling fmi2EnterInitializationMode) if
• initial = "exact" or "approx" [in order to set the corresponding start value].**

This can also bee seen in the SEA and INI definition of ME and CS state charts

So before the removal of "approx" for input inputs could be set by the implicit definition stated above, but after the removal they are no longer allowed to be set.

By reading the tickets, the minutes and my memory I cant recall that this aspect was ever discussed and it is an unintentional restriction that should be corrected and clarified in 2.0.1

Up until a few days ago i have never encounter an FMU that had a problem with this so it seems that most vendors have overseen this restriction (or doesn't care about it).

The reason for supporting changing start values of input is to make it more user friendly and robustness.
if a user generates an FMU that has singularities, or other problems with certain values on inputs, and they have forgotten to set proper start values this can easily be corrected at import, instead of having to re generate the FMU.

Karl Wernersson

----

Migrated-From: https://trac.fmi-standard.org/ticket/369

KarlWernersson:
update text to allow seting initial guess vaules for inputs before enter initialization mode.

Checklist

### DIFF
--- a/docs/2_3_common_states.adoc
+++ b/docs/2_3_common_states.adoc
@@ -346,7 +346,7 @@ If `stopTimeDefined == fmi3False`, then no final value of the <<independent>> va
 
 Function `fmi3Set{VariableType}`::
 
-This function can be called for variables with <<variability>> latexmath:[\neq] <<constant>> and with <<initial>> = <<exact>> or <<approx>>.
+This function can be called for variables with <<variability>> latexmath:[\neq] <<constant>> and with <<initial>> = <<exact>> or <<approx>>, or variables with <<causality>> = <<input>>.
 The intention is to set <<start>> and guess values for these variables.
 
 ===== State: Initialization Mode [[InitializationMode]]


andreas-junghanns (approved):

chrbertsch (approved):

docs/2_3_common_states.adoc
```diff
@@ -346,7 +346,7 @@ If `stopTimeDefined == fmi3False`, then no final value of the <<independent>> va
 
 Function `fmi3Set{VariableType}`::
 
-This function can be called for variables with <<variability>> latexmath:[\neq] <<constant>> and with <<initial>> = <<exact>> or <<approx>>.
+This function can be called for variables with <<variability>> latexmath:[\neq] <<constant>> and with <<initial>> = <<exact>> or <<approx>>, or variables with <<causality>> = <<input>>.
```
t-sommer, line 349: Wouldn't it make sense to require `inputs` to be `initial = exact`?

KarlWernersson: in a way yes, but exact input implies that they are known before hand and this is not true in many cases, so approx. would make more sense as setting this before enter initialization mode is just a start value for eventual iteration unless the input is unconnected. However approx. also forbids you to change them during initialization wich does not apply to inputs.. 

In general i think they were exact, in some beta release of fmi 2 this was removed and this bug that you wasn't allow to set them was introduced.

I would be fine to return the exact on inputs as they behave in the same way as variables with initial exact during initialization.

chrbertsch: Currently we forbid variables of causality `input `to have an `initial` attribute.  
@t-sommer : do you propose to change this?

t-sommer: I think it would be cleaner to require `initial = exact` for inputs since we could get rid of two exceptions in the text. Also IMHO it does not matter if the value is a guess value. It will always be exactly this value.

andreas-junghanns:
If PR #1260 gets accepted, this PR here should NOT be merged - it would be made redundant, or even add confusion.

andreas-junghanns (changes requested):
#1260 is a better solution to this issue, I think.

KarlWernersson:
Closing due to #1260

PR #1249, opened 2020-11-03
Correct so that relativeQuantity affects base units as well as DisplayUnits

### LINKED ISSUE #1165: FMI 3.0: Unit specification error regarding relativeQuantity
In Section _2.2.2. Definition of Units_ the conversion of units based on `<BaseUnit>` is described.

One important pitfall is to incorporate the `offset` if `relativeQuantity="true"`; an example is given in subsection _Signal connection check_ which even refers to the definition of `relativeQuantity` below.

The problem now is, that the definition/description of `relativeQuantity` only talks about semantic means w.r.t. `displayUnit`:

> If this attribute is true, then the offset of displayUnit must be ignored. [For example, 10 degree Celsius = 10 Kelvin if relativeQuantity = true and not 283.15 Kelvin.]

I think, this is wrong. The `relativeQuantity` is much more important for the formal conversion of units based on `<BaseUnit>` definitions.

The definition of `relativeQuantity` should be respectively adapted. Also a more complex example how to use it for unit conversions has to be added.

Finally, the formula given for `<BaseUnit>`-based unit conversion should already incorporate `relativeQuantity`. At the moment it is just:

 > BaseUnit_value = factor * Unit_value + offset

but I guess it should be `BaseUnit_value = factor * Unit_value + (if relativeQuantity 0.0 else offset)`, although it now requires some linking to `relativeQuantity` which is not part of the `<Unit>`/`<BaseUnit>` definition itself but rather variable specific (it is an attribute of `<Float64Type>`, not `<BaseUnit>`).

KarlWernersson:
fix #1165

### DIFF
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -458,7 +458,7 @@ _When two variables v1 and v2 are connected and for both of them `<BaseUnit>` el
 _If `factor` and `offset` are also identical, again the connection equation `v2 = v1` holds._
 _If `factor` and `offset` are not identical, the tool may either trigger an error or, if supported, perform a conversion; in other words, use the connection equation (in this case the `relativeQuantity` of the `<TypeDefinition>`, see below, has to be taken into account in order to determine whether `offset` shall or shall not be utilized):_
 +
-`factor(v1) * v1 + (if relativeQuantity(v1) then 0 else offset}offset(v1)) = factor(v2) * v2 + (if relativeQuantity(v2) then 0 else offset}offset(v2))`
+`factor(v1) * v1 + (if relativeQuantity(v1) then 0 else offset(v1)) = factor(v2) * v2 + (if relativeQuantity(v2) then 0 else offset(v2))`
 +
 _As a result, wrong connections can be detected (for example, connecting a force with an angle-based variable would trigger an error) and conversions between, say, US and SI units can be either automatically performed or, if not supported, an error is triggered as well._
 +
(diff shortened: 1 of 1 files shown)


t-sommer (approved):

docs/2_4_common_schema.adoc
```diff
@@ -346,7 +346,7 @@ A value latexmath:[v_{unit}] in `Unit` is converted to the base unit latexmath:[
 
 [latexmath]
 ++++
-v_{base} = \texttt{factor} * v_{unit} + \texttt{offset}
+v_{base} = \texttt{factor} * v_{unit} + \texttt{if relativeQuantity then 0 else offset}
```
chrbertsch, line 349: Should a  similar adaptation not be explicily made for
_(in this case the relativeQuantity of the <TypeDefinition>, see below, has to be taken into account in order to determine whether offset shall or shall not be utilized):_
`factor(v1) * v1 + offset(v1) = factor(v2) * v2 + offset(v2)`
in the paragraph "Unit check when connecting variables of different FMUs"

Shouldn't we explain, that `factor `and `offset ` are properties of the BaseUnit, while `relativeQuantity` is an attribute of a type and link to its definition

KarlWernersson: Yes that would be better

chrbertsch: @t-sommer : in which order shall we proceed?

chrbertsch (approved):

chrbertsch:
Can we merge this? Please review. I would like to  port this to FMI 2.0.2

t-sommer (approved):

PR #1257, opened 2020-11-03
Fix constant structuralParameter and a bit more

andreas-junghanns:


andreas-junghanns:
@KarlWernersson : Please check if this fully addresses your issue.

KarlWernersson (approved):
Thanks it looks good,
unfortunately a bit hard to see if the table has been changed correctly from the diff, but I assume it has.

Issue #1256, opened 2020-11-03
Table 1 modelExchange features


KarlWernersson:
Feature  Early Return  & Intermediate Update is marked as an X for model exchange, this should rather be a -, as you can actually do both there as it is included in its definitions of ME, however you not need the cumbersome additional features of co-simulation to do it.

andreas-junghanns:
@KarlWernersson : are you sure? So far, I thought that there is no need to have Intermediate Update Mode in ME because the master has complete control over how to advance time and when to set which input and read output. This is completely different in CS and SE where the control during a doStep() or activateModelPartition() must be handed back actively to the importer.

If you think ME can benefit from this, we need more than the table change. We would have to explain the MS use case in 2.3.3 State: Intermediate Update Mode at the beginning AND also fix the text that currently implicitly and likely also explicitly excludes ME from using Intermediate Update Mode.

chrbertsch:
Coming back to Karls proposal, we should perhaps clarify  and add a legend explainging what "x" and "-" mean in that table: 

For me: 
"check": suppported
"x" not supported
"-" not relevant/does not make sense

For me that points that Karl mentions are "not relavant" for ME, as you do not need these features and can get the same or even better with ME-means.

So Karl does not propose to use the "check" as is realized in https://github.com/modelica/fmi-standard/pull/1258/files

KarlWernersson:
@chrbertsch @andreas-junghanns 
yes Christian understood me correctly 
Form my understanding there should be  "-" as it is not relevant. An ME master can "return" whenever it want's and inquire values whenever it wants so it can do already all of this whit out the extra API's. But it does not use the exact features that are added to co-simulation.
sorry if I was unclear.

andreas-junghanns:
> Form my understanding there should be "-" as it is not relevant. An ME master can "return" whenever it want's and inquire values whenever it wants so it can do already all of this whit out the extra API's. But it does not use the exact features that are added to co-simulation.

Current PR #1258 should reflect your concerns now. Please check again.

PR #1258, opened 2020-11-03
Fix ME feature entries

andreas-junghanns:
~~If this gets accepted, we need to fix the text to reflect that we allow Intermediate Update Mode also in ME.~~

### DIFF
--- a/docs/1___overview.adoc
+++ b/docs/1___overview.adoc
@@ -156,10 +156,10 @@ image::images/fmi-types-overview.svg[width=50%, align="center"]
 |Solver Included
 |icon:times[]
 |icon:check[]
-|--
+|icon:times[]
 
 |Scheduler Included
-|--
+|icon:times[]
 |icon:check[]
 |icon:times[]
 
--- a/docs/1___overview.adoc
+++ b/docs/1___overview.adoc
@@ -149,17 +149,17 @@ image::images/fmi-types-overview.svg[width=50%, align="center"]
 |Scheduled Execution
 
 |Advancing Time
-|call <<fmi3SetTime>>
-|call <<fmi3DoStep>> and monitor argument <<lastSuccessfulTime>>
-|call <<fmi3ActivateModelPartition>>
+|Call <<fmi3SetTime>>
+|Call <<fmi3DoStep>> and monitor argument <<lastSuccessfulTime>>
+|Call <<fmi3ActivateModelPartition>>
 
 |Solver Included
 |icon:times[]
 |icon:check[]
-|icon:times[]
+|Not applicable
 
 |Scheduler Included
-|icon:times[]
+|Not applicable
 |icon:check[]
 |icon:times[]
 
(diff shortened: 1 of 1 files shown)


chrbertsch (changes requested):
I Propose to change from "check" to "--" and to add a legend what "check", "x", "-" mean

See https://github.com/modelica/fmi-standard/issues/1256#issuecomment-721237903

andreas-junghanns:
> 
> 
> I Propose to change from "check" to "--" and to add a legend what "check", "x", "-" mean
> 
> See [#1256 (comment)](https://github.com/modelica/fmi-standard/issues/1256#issuecomment-721237903)

Why do we need 3 values?
I think the "--" values should be replace with "x" (meaning "No support", "no", "not included"...).
This would leave us with two, easily understood values and no legend is needed.

If you think we need three values, please describe here, what "--" means compared with "x" - and I will reintroduce the "--".

andreas-junghanns:
Sorry: 8c7b580 commit message should read `Replace "--" with "icon:times[]"`

chrbertsch (changes requested):
I do not like the "check" for "early return" and "intermediate update" for ME, as these terms are linked to a concrete realization and API calls for CS.

If we do not want to introduce  a "-" with legend entry "does not make sense for this FMU kind", we could just write "intrinsic ME feature"  or "not needed in ME" instead of the "check" in the cells.

KarlWernersson (changes requested):
I agree with Christian,
The more i think of it the harder i think it is to make the table, what features should be mentionsed? we don't explain the features there so it could be hard to understand for someone just reading it without larger prior knowledge. And for someone with larger prior knowledge its not really necessary.

but at least a Not relevant tag (the way I interpretended the "-") is needed if we dont wat to change to much

andreas-junghanns:
> 
> 
> I do not like the "check" for "early return" and "intermediate update" for ME, as these terms are linked to a concrete realization and API calls for CS.
> 
> If we do not want to introduce a "-" with legend entry "does not make sense for this FMU kind", we could just write "intrinsic ME feature" or "not needed in ME" instead of the "check" in the cells.

Agreed.
Why not have some brief text in the cells - we do that elsewhere in the table to make it more easily understandable?
I will work this into this PR as a proposal.

andreas-junghanns:
> The more i think of it the harder i think it is to make the table, what features should be mentionsed? we don't explain the features there so it could be hard to understand for someone just reading it without larger prior knowledge. And for someone with larger prior knowledge its not really necessary.

Yes, this is a problem in all technical texts with significant depth.
The idea here is to introduce high-level concepts and in breadth-first fashion traverse the concept graph.
This can never be perfect, but we have to start somewhere to untangle the differences between the different FMI interface types.
We have to find a better way or leave this as the top-most traversal of the concepts, I think.

andreas-junghanns:
I changed the cell entries to now "speak" a small number of words.
Maybe this is good enough?

chrbertsch (approved):

KarlWernersson (approved):

Issue #271, opened 2018-10-16
initial = approx  not clear for inputs


modelica-trac-importer:
**Modified by andreas&#46;junghanns on 11 Jul 2014 10:08 UTC**
At the FMI web meeting the setting of initial=approx for input variables was discussed:
- It is unsymmetrical that initial must be set to approx and that there is not a default value.
- It is confusing to set it to approx, because approx is meant to be used for iteration variables.
- It is inconsistent with "InitialUnknowns", see text in #258: variables with initial=approx can be optionally included in InitialUnknowns (which is wrong for inputs).

Conclusion of the meeting: Input variables must have a start value (start attribute) and do not set a value for the initial attribute.

Furthermore, it was proposed by Torsten to explicitly state that if fmi2SetXXX is not called on an input, then the FMU must use the start value of the input (and does not trigger an error).

----

**Reported by otter on 11 Jul 2014 09:53 UTC**
At the FMI web meeting the setting of initial=approx for input variables was discussed:
- It is unsymmetrical that initial must be set to approx and that there is not a default value.
- It is confusing to set it to approx, because approx is meant to be used for iteration variables.
- It is inconsistent with "InitialUnknowns", see text in #258: variables with initial=approx can be optionally included in InitialUnknowns (which is wrong for inputs).

Conclusion of the meeting: Input variables must have a start variable and do not set a value for initial 

Furthermore, it was proposed by Torsten to explicitly state that if fmi2SetXXX is not called on an input, then the FMU must use the start value of the input (and does not trigger an error).

----

Migrated-From: https://trac.fmi-standard.org/ticket/271

modelica-trac-importer:
**Modified by andreas&#x2e;junghanns on 11 Jul 2014 10:08 UTC**

modelica-trac-importer:
**Comment by otter on 11 Jul 2014 11:00 UTC**
Fixed according to the suggestion at the web meeting.

Note, the statement "if fmi2SetXXX is not called on an input, then the FMU must use the start value of the input (and does not trigger an error)." was already in the document.

PR #1260, opened 2020-11-03
Make input initial=exact

andreas-junghanns:


andreas-junghanns:
This is an alternative solution to #1244 - as @t-sommer suggested.

t-sommer (approved):

chrbertsch:
Setting initial=exact makes the start value for inputs mandatory. Is this intended and was this also the case before?

andreas-junghanns:
> Setting initial=exact makes the start value for inputs mandatory. Is this intended and was this also the case before?

Yes, in FMI 2.0 input variables had to provide start values.
This re-definition (requiring inputs to have "initial = exact") makes the text in many ways simpler, because the "causality = input" case is not special anymore, but subsumed under the "initial = exact".

andreas-junghanns:
@KarlWernersson & @chrbertsch : Can you look at this and approve, please?
If you do, I would close PR #1244.

I want these PRs accepted or rejected because we will go into a WriteFest tomorrow - your feedback today would be good - thanks!

chrbertsch (approved):

KarlWernersson (approved):
I like it.

The problem originate in this fix
https://github.com/modelica/fmi-standard/issues/271

jph-tavella:
This applies to FMI 3.0, though, so I'm not sure how should we consider it for 2.0.1?

PR #1264, opened 2020-11-03
Move two sections from Common Concepts to ME

andreas-junghanns:


andreas-junghanns:
This is just the move - rewrite will happen at the next WriteFest.

chrbertsch (approved):

PR #1265, opened 2020-11-04
Disallow offset for inverse DisplayUnit

### LINKED ISSUE #1220: Allow inverse display units

chrbertsch:
resolves #1220

### DIFF
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -521,7 +521,8 @@ image::images/schema/DisplayUnit.png[width=70%, align="center"]
 A `<DisplayUnit>` is defined by `name`, `factor`, `offset`, and `inverse`.
 The attribute `name` must be unique with respect to all other names of the `<DisplayUnit>` definitions of the same `Unit`.
 _[Different `Unit` elements may have the same `<DisplayUnit>` names]._ +
-`inverse == true` is only allowed if `offset == 0`
+`inverse == true` is only allowed if `offset == 0`. +
+_[Reason: no use case is known for the combination of inverse and offset, which would also be more complicated.]_
 
 A value latexmath:[v_{unit}] in `Unit` is converted to a value latexmath:[v_{display}] in `DisplayUnit` by the equation:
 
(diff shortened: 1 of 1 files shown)


andreas-junghanns (approved):
This still feels kind of wrong. In A2L, MDF,... everywhere, inverse formulas for variables are allowed, without such restrictions.

t-sommer (approved):

KarlWernersson (commented):
Opinions on how to do this seems to be quite diverse, 
Current formula is wrong so it need to be fixed.

I gave formula both for making it correct with offset and a proposal to fix the offset.
I am personally leaning to skipping it but I don't have a strong opinion in the matter

But since the diverging opinions should we step back and collect opinions?

alternative, I can approve the pull request and if someone wants offset for inverse units they can reopen it.

Should we have a statement why offset is not allowed for inverse units 
(could be its slightly more complicated and we don't see actual use case so we decided to ignore it)

andreas-junghanns:
> alternative, I can approve the pull request and if someone wants offset for inverse units they can reopen it.

Agreed.
 
> Should we have a statement why offset is not allowed for inverse units
> (could be its slightly more complicated and we don't see actual use case so we decided to ignore it)

This would be needed. Can you add this to the PR?

chrbertsch:
> >    Should we have a statement why offset is not allowed for inverse units
    (could be its slightly more complicated and we don't see actual use case so we decided to ignore it)

> This would be needed. Can you add this to the PR?

I added a sentence. Please review.

andreas-junghanns (approved):

andreas-junghanns:
In case we ever meet a use case that requires this, we can easily drop this restriction.

KarlWernersson (approved):

PR #1272, opened 2020-12-08
New clocks

andreas-junghanns:


docs/2_2_common_mechanisms.adoc
```diff
@@ -252,13 +242,36 @@ This function call is allowed always and in all interface types.
...
+
+In Model Exchange, time is under the sole control of the importer and its integration algorithm.
+The model itself receives the current time to be used in its computation with <<fmi3SetTime>>.
+In fact, time is not necessarily advancing linearly as solvers might need to jump back and forth in time to localize events using zero-crossings of <<eventIndicator, event indicators>>.
+
+In Co-Simulation, time advances in (possibly variable) steps negotiated between the co-simulation algorithm of the importer and the FMU.
+The importer calls <<fmi3DoStep>> with the <<currentCommunicationPoint>> and a target <<communicationStepSize>> (required to be larger than 0.0).
+During this <<fmi3DoStep>>, both importer and FMU might encounter events (or other situations) that require reduction of the <<communicationStepSize>> (potentially even down to 0.0).
+The FMU may use <<earlyReturn>> argument of the <<fmi3DoStep>> function to tell the import that the FMU needs to return earlier, and the importer may use the callback <<fmi3CallbackIntermediateUpdate>> to signal the FMU that the later should return earlier.
```
chrbertsch, line 256: later  --> latter

docs/2_2_common_mechanisms.adoc
```diff
@@ -252,13 +242,36 @@ This function call is allowed always and in all interface types.
...
+
+They are identified with a unique handle called <<valueReference, value reference>>.
 
-All variables of an FMU are identified with a handle called <<valueReference, value reference>>.
-The handle is defined in the <<modelDescription.xml>> file as attribute <<valueReference>> in variable elements.
-Each variable must have a unique <<valueReference>>.
+The attribute <<causality>> defines the direction of the information flow with respect to the FMU (e.g. <<input>>, <<output>>, <<parameter>>).
 
-A variable can be a scalar or an array.
+A variable can be a scalar or an multi-dimensional array.
```
chrbertsch, line 274: an --> a

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,428 +338,335 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
+<<fmi3GetInterval>> and <<fmi3GetShift>> in *Initialization Mode*
+|clocked PI-controller with interval defined by fixed parameter(s) of the FMU
+
+|`tunable`
+|[[tunableClock,`tunable`]]<<fmi3SetClock>> in *Event Mode*,
+<<fmi3GetShift>> in *Initialization Mode*,
+<<fmi3GetInterval>> in *Event Mode*
+|clocked PI-controller with interval defined by tunable parameter(s) of the FMU
+
+// TODO should we rename aperiodic clocks to variable clocks to align keyword and name?
```
chrbertsch, line 508: I like "variable clock"

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,428 +338,335 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
-_Remove the entire section: we do not want to say: don't be stupid.]_
+Triggered output clocks::
+are activated within the FMU and the importer must call <<fmi3GetClock>> in *Event Mode*.
+There are two mechanisms for an FMU to communicate to the importer that a triggered output clock is about to tick:
+ * <<early-return,early return>> from a <<fmi3DoStep>> with <<clocksAboutToTick, `clocksAboutToTick == fmi3True`>>, and
+ * calling <<fmi3CallbackIntermediateUpdate>> with <<clocksTicked, `clocksTicked == fmi3True`>>.
 
-The environment can evaluate required <<clock>> casts for connections using the information on the <<clockReference>> attribute of assigned variables _[i.e. the variables that are assigned to a <<clock>> based on <<modelDescription.xml>> information]_.
+Input clocks::
+[[inputClock, `input clock`]]All clocks with <<causality>> <<input>> are collectively called <<inputClock>>.
```
chrbertsch, line 635: Is this really true?  From the table above I understand that only _triggered_ clocks with causality input are called "input clocks"

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,428 +338,335 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
 
-_[Example: Elementary blocks like a PI controller will have all variables on an <<inputClock>> with identical properties._
-_Connecting such blocks together will therefore be possible without computationally expensive clock cast operations (and the environment can readily do the computation of assigned variables of the connected blocks based on the same <<clock>>).]_
+OutputClocks::
```
chrbertsch, line 637: add space --> Output clocks::

docs/2_4_common_schema.adoc
```diff
@@ -659,129 +664,76 @@ _This information can, for example, be used to increase numerical stability and
...
-`= synchronousTime` is used, if the properties of the <<clock>> adhere to synchronous clock theory.
 
-[[communicationPoint,`communicationPoint`]]
-`= communicationPoint` <<clock,`clocks`>> define sampling points (i.e. communication points) for the variables of model partitions.
-
-It is not allowed to include <<clock,`clocks`>> of different `clockType` in one FMU.
-`clockType` is a required attribute.
+2+|Attributes for `ClockType`
 
 |`triggeredBy`
```
chrbertsch, line 670: Isn't this information that belongs to the model structure?

chrbertsch (commented):
Some small comments

docs/2_2_common_mechanisms.adoc
andreas-junghanns: Done

docs/2_2_common_mechanisms.adoc
andreas-junghanns: Done

docs/2_2_common_mechanisms.adoc
andreas-junghanns: We will leave this as is, we did this on purpose to use input clocks in the text. If we end up not using it, we can remove this definition later.

docs/2_2_common_mechanisms.adoc
andreas-junghanns: Done

docs/2_4_common_schema.adoc
andreas-junghanns: In principle yes, maybe. But this is the dependency of an inputClock on the tick of an outputClock - the direction is opposite to "normal" dependencies - we discussed this with Irina and Matthias at length, we should leave as is.

docs/2_2_common_mechanisms.adoc
andreas-junghanns: ... and also: outputClocks may additionally depend on inputClocks (subclocks).

docs/2_2_common_mechanisms.adoc
andreas-junghanns: We regrouped all aperiodic clocks in the table and renamed the `variable` clock to `changing` clock to avoid the confusion between noun and adjective `variable`.

docs/2_4_common_schema.adoc
```diff
@@ -855,1491 +807,1469 @@ Element `<fmiModelDescription><DefaultExperiment>` is defined as:
...
+The algebraic relationship to the <<input,`inputs`>> can be defined via the <<dependencies>> attribute of <<Output,`<fmiModelDescription><ModelStructure><Output>`>>.
 
-====== Placement, Extent, and Painting Order of Graphical Items
+[[local,`local`]]
+`= local`: Local variables are: +
 
-image::images/GraphicalRepresentation.svg[width=70%, pdfwidth=70%, align="center"]
+* continuous <<state,`states`>> and their <<StateDerivative,`StateDerivative`pass:[s]>>, <<DiscreteTimeState,`DiscreteTimeState`pass:[s]>>, <<EventIndicator,`EventIndicator`pass:[s]>> or <<InitialUnknown,`InitialUnknown`pass:[s]>>.
+These variables are listed in the `<fmiModelDescription><ModelStructure>`. +
+* [[localClock,`local clock`]] internal, intermediate variables or local clocks which can be read for debugging purposes and are not listed in the `<fmiModelDescription><ModelStructure>`.
```
IZacharias, line 930: Do local clocks still exist?

docs/2_4_common_schema.adoc
```diff
@@ -855,1491 +807,1469 @@ Element `<fmiModelDescription><DefaultExperiment>` is defined as:
...
-The actual dimension sizes of arrays are also constrained by the FMU platform, due to memory and addressing constraints:
-Since the API functions to access variables and their values are constrained to `size_t` individual elements, platforms with addresses of less than 64-bit width will not be able to access elements beyond their addressing limits, neither will they be able to allocate enough memory or address space to represent such arrays.
-For these reasons implementations must take platform-specific constraints into account when changing dimension sizes, and must be prepared to handle the inability of the FMU to adjust to the desired sizes during *Configuration Mode* or *Reconfiguration Mode*.
+>|_(18)_
+|_<<clock>>_
+|_Variable that defines a <<clock>>._
 
-Changing any dimension of a variable in *Configuration Mode* or *Reconfiguration Mode* invalidates the variable's current value (including its <<start>> value).
-It should be noted that changing a <<structuralParameter,`structural parameter`>> might affect dimension sizes of several variables.
+>|_(19)_
```
IZacharias, line 1333: In the table there are two line for clocks. Delete line 19?

docs/2_4_common_schema.adoc
andreas-junghanns: We have tried to keep clocks a "normal" variable.
So, yes, local clocks are not forbidden in the current document.

docs/2_4_common_schema.adoc
```diff
@@ -855,1491 +807,1469 @@ Element `<fmiModelDescription><DefaultExperiment>` is defined as:
...
+>|_(2)_
+|_<<tunable>> <<parameter>>_
+|_<<tunable,`Tunable`>> <<parameter>> (changing such a <<parameter>> triggers event handling (ME) or takes effect at the next communication point (CS and SE), and <<tunable>> <<calculatedParameter>>/<<output>>/<<local>> variables might change their values)._
 
-_[Flow variables have a direction and must fulfill a zero sum constraint i.e. the sum of all flow variables connected must be zero (Kirchhoff's current law)._
-_In addition because different tools might have different direction definitions both, `inflow` and `outflow` are available as `variableKind`._
-_For variables with `inflow` a positive value means that the flow is inwards, and for `outflow` a positive value means that the flow is outwards._
-_For the sake of simplicity in the following latexmath:[\dot{m}_i] denotes an inflowing quantity:_
+>|_(3)_
+|_<<fixed>> <<dependenciesKind,`dependent`>> <<parameter>>_
```
IZacharias, line 1263: Shouldn't calculatedParameter be written here instead of dependent parameter?

andreas-junghanns: yes, I think so

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,522 +343,711 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
+|latexmath:[\mathbf{T}_{interval, i}]
+|The time interval until the next <<clock>> tick, defined differently for the different clock types.
+latexmath:[\mathbf{T}_{interval, i}] can be set with <<fmi3SetIntervalDecimal>>) or get with <<fmi3GetIntervalDecimal>> as floating point value, or as rational number using <<fmi3SetIntervalFraction>> or <<fmi3GetIntervalFraction>>: latexmath:[\mathit{intervalCounter} / \mathit{resolution}] (see <<resolution>>).
 
-An <<inputClock>> is a <<periodic>> or <<periodic,aperiodic>> <<clock>> that is defined by a <<clock>> outside of the FMU.
-I.e., only when the <<clock>> is activated by the environment of the FMU, then the FMU is informed that the <<clock>> ticks at this time instant.
-An <<inputClock,input clocks>> must be independent to all other input clocks.
-_[TODO: what does that mean?]
-If dependent clocks exist, they can be exposed as <<local>> or <<output>> clocks.
+|latexmath:[\mathbf{t}_\mathit{event}]
```
MBlesken, line 713: Why not \mathbf{t}_\mathit{event, i} since this event occurs several times. It looks strange without a substcript i in equations with t_i and T_interval,i, don't you think?

MBlesken (commented):
I will continue the review...

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,522 +343,711 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
-Claudio: rewrite this to make it clearer: the distinction between these two are just local vars and outputs.
-Claudio: Also relate to the previous example.
-Claudio: Also change the modelica sentence.
-Claudio: Make it clear that clocks are just a way to implement numerically sound event having across different FMUs.
-]_
-[latexmath]
-++++
-y_j=g_j(x_{j},u_j,t_j);
-++++
+==== Selective Computation of Variables [[selectiv-computation]]
```
chrbertsch, line 477: Is the selective calculation of variables as described in this subsection really a common concept or >=2 FMI interface types?
To my understanding, this is only the case in ME. 
In CS, calculations are triggerd in fmi3Dostep, fmi3NewDiscreteStates, and in ME in activateModelPartition

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,522 +343,711 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
+Model partitions are initialized at their first clock activation, not in <<InitializationMode>>.
 
-_[TODO: What is a clock cast? If a clock cast can be done by the environment anyway, why mention which inputs/outputs can be connected?_
-_Remove the entire section: we do not want to say: don't be stupid.]_
+In some clock semantics (e.g. <<MLS12>>), a clocked variable has a value only when its associated clock is active.
+However, `fmi3Get{VariableType}` will return the value computed at the last clock activation time.
+Before the first clock tick, clocked variables have their initial value.
 
-The environment can evaluate required <<clock>> casts for connections using the information on the <<clockReference>> attribute of assigned variables _[i.e. the variables that are assigned to a <<clock>> based on <<modelDescription.xml>> information]_.
+The association between <<clocked-variable,clocked variables>> and their <<clock,clocks>> is defined by the <<model-dependencies,model dependencies>>.
```
chrbertsch, line 851: Could we explain this more explicitely, please?
My understanding when reading this: 
We have Ouputs or DiscreteTimeStates listed in the ModelStructure, which have in their dependendency list at least one clock, so they are clocked. But what about other (local) variables, that are not listed in the modelStructure and shall be clocked?

chrbertsch (commented):
(Comments and questions on chapters 1+2)

docs/3_4_model_exchange_schema.adoc
```diff
@@ -18,16 +18,16 @@ The attributes in the following table are defined on top of the <<common-capabil
...
-|`modelIdentifier`
+|<<modelIdentifier>>
 |Short class name according to C syntax, for example, `A_B_C`.
 Used as prefix for FMI functions if the functions are provided in C source code or in static libraries, but not if the functions are provided by a DLL/SharedObject.
-`modelIdentifier` is also used as name of the static library or DLL/SharedObject.
+<<modelIdentifier>> is also used as name of the static library or DLL/SharedObject.
 See also <<header-files-and-naming-of-functions>>.
 
 |`completedIntegratorStepNotNeeded`
 |If `true`, function <<fmi3CompletedIntegratorStep>> need not be called (this gives a slightly more efficient integration).
```
chrbertsch, line 28: "(this gives a slightly more efficient integration)" : I do not understand this text. Integartion in the sens of solving the ODE? Could someone please improvethe eplanation. Should be non-normative anyway.

docs/3___model_exchange.adoc
```diff
@@ -1,18 +1,44 @@
...
+
+* dedicated <<InitializationMode>>, which allows computation of consistent initial conditions over <<algebraic-loops,algebraic loops>>,
+
+* handling of <<clock,clocks>> and <<EventMode,events>>,
+
+* solution of algebraic loops involving inputs and output variables of this FMU (see <<algebraic-loops,algebraic loops>>),
+
+* selective computation of FMU variables required by the solver algorithm, e.g. outputs, event indicators, derivatives (see <<selectiv-computation>>),
+
+//TODO: Christian: is this the correct formulation?
```
chrbertsch, line 41: fine for me.

docs/4___co-simulation.adoc
```diff
@@ -1,93 +1,48 @@
 == FMI for Co-Simulation [[fmi-for-co-simulation]]
 
-This chapter defines the Functional Mock-up Interface (FMI) for the coupling of two or more simulation models in a co-simulation environment (FMI for Co-Simulation).
-It is designed both for coupling with subsystem models, which have been exported by their simulator together with its solver as runnable code, and for coupling of simulation tools on a single machine or as part of a distributed co-simulation.
+FMI for Co-Simulation provides a standardized interface for execution of simulation models or tools in a co-simulation environment.
+In contrast to FMI for Model Exchange, a Co-Simulation FMU implements not only the model algorithm, but also the required solution method.
+The data exchange between FMUs is restricted to discrete communication points latexmath:[\mathbf{t}_i].
```
chrbertsch, line 5: "The data exchange between FMUs is restricted to discrete communication points": this is only true for the "direct" data exchange, but via the importer in intermediateUpdateMode, data can be exchanged somehow.

docs/4_2_co-simulation_api.adoc
```diff
@@ -1,387 +1,134 @@
...
+|`fmi3Set{VariableType}`
 
-_[Usually the co-simulation algorithm should be able to derive (but is not forced to do so) the correct communication point times for <<inputClock,`input clocks`>> in advance and thus it should be able to set the proper <<communicationStepSize>> for <<fmi3DoStep>>._
-_This might not be possible if an <<periodic,aperiodic>> <<inputClock>> of an FMU depends on the ticking of an <<periodic,aperiodic>> <<outputClock>> of another FMU or other <<periodic,aperiodic>> tick sources.]_
+|Set continuous-time and discrete-time <<input,`inputs`>> latexmath:[\mathbf{u}_{c+d}(\mathbf{t})].
+|`fmi3Set{VariableType}`
 
-==== State Machine for Co-Simulation [[state-machine-co-simulation]]
+|Get values of variables latexmath:[\mathbf{v}(\mathbf{t})].
+|`fmi3Get{VariableType}`
```
chrbertsch, line 32: This violates our calling sequence of fmi3SetXX fmi3DoStep fmi3GetXX, order should be changed!

docs/5_2_scheduled_execution_api.adoc
```diff
@@ -8,379 +8,125 @@ The simulation algorithm signals to the FMU that it supports and has recognized
...
-- `fmi3ValueReference` <<clockReference>>: <<valueReference>> of an <<inputClock>> variable defined in the <<modelDescription.xml>> which shall be activated
+|Set discrete-time <<input,`inputs`>> latexmath:[\mathbf{u}_{d}(\mathbf{t})].
+|`fmi3Set{VariableType}`
 
-- `size_t` <<clockElementIndex>>:: The element index of the <<inputClock>> variable which shall be activated.
-For a scalar <<clock>> variable this must be 0; for array <<clock>> variables, the element clock to activate is specified using the 1-based element index.
-Using the element index 0 means all elements of the <<clock>> variable.
-(Note: If an array has more than one dimension the indices are serialized in the same order as defined for values in <<serialization-of_variables>>).
+|Get values of variables latexmath:[\mathbf{v}(\mathbf{t})].
+|`fmi3Get{VariableType}`
```
chrbertsch, line 50: This order violates the calling sequence, should be moved after fmi3activateModelPartition!

docs/5_2_scheduled_execution_api.adoc
```diff
@@ -8,379 +8,125 @@ The simulation algorithm signals to the FMU that it supports and has recognized
...
 
 Function <<get-and-set-variable-values,`fmi3Set{VariableType}`>>::
 This function can be called before scheduling a model partition for variables assigned to that model partition via its associated <<clock>> and all variables not associated to a <<clock>> (global variables).
 
-Function <<get-and-set-variable-values,`fmi3Get{VariableType}>>`, `fmi3GetOutputDerivatives`, <<fmi3GetDirectionalDerivative>>::
+Function <<get-and-set-variable-values,`fmi3Get{VariableType}>>`::
 These functions can be called after the computation of a model partition for variables assigned to that model partition via its associated <<clock>> and all variables not associated to a <<clock>> (global variables).
 
++
 Set/get operations must be atomic for a single variable.
```
chrbertsch, line 72: What does "atomic" mean here?

docs/5_2_scheduled_execution_api.adoc
```diff
@@ -8,379 +8,125 @@ The simulation algorithm signals to the FMU that it supports and has recognized
...
-         activateModelPartitionAperiodic(instance, activationTime);
-      case 8:
-         // Input clock AperiodicClock
-         activateModelPartition50ms(instance, activationTime);
-      ...
-   }
-}
-----
++
+This is a different timing concept compared to <<fmi3DoStep>> calls for Co-Simulation FMUs or the <<fmi3SetTime>> for Model Exchange FMUs.
```
chrbertsch, line 117: Is this and the following non-normative?

chrbertsch (commented):
(Comments and questions on chapters 3,4,5)

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
-// TODO: Configuration Mode
-// TODO: In CS is an iteration variable i
+===== State: Initialization Mode [[InitializationMode,*Initialization Mode*]]
+
+The <<InitializationMode>> is used by the simulation algorithm to compute consistent initial conditions for the overall system.
+Equations are active to determine the initial FMU state, as well as all <<output,`outputs`>> (and optionally other variables exposed by the exporting tool).
+Artificial or real algebraic loops over connected FMUs in <<InitializationMode>> may be handled by using appropriate numerical algorithms.
 
-Mathematical Description::
+In <<InitializationMode>>, the FMU computes initial values at the start time latexmath:[\mathbf{t}_\mathit{start}] using function latexmath:[\mathbf{f}_\mathit{init}], not present in the other modes, for example, equations to define the <<start>> value for a <<state>> or for the derivative of a <<state>>.
```
clagms, line 330: I have trouble understanding this sentence. Are we defining what f_init is? Why is it defined as a function? During initialization mode, the FMU activates a set of equations. We can call that set f_init, but I wouldn't call it a function...

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
+|Set <<tunable>> <<parameter,`parameters`>> latexmath:[\mathbf{p}_{\mathit{variability == tunable}}]
+|`fmi3Set{VariableType}`
+
+|Set continuous-time and discrete-time <<input,`inputs`>> latexmath:[\mathbf{u}_{c+d}(\mathbf{t})]
+|`fmi3Set{VariableType}`
 
-In this mode all continuous-time and discrete-time equations are active and the unknowns at an event can be computed and retrieved.
-The event time of a <<state event>> may be determined if a domain change of at least one event indicator is detected at the end of a completed integrator step.
+a|
+* Activate clocks and respective model partitions in latexmath:[\mathbf{f}_{\mathit{event}}]
```
clagms, line 534: I cannot find a formal definition of f_event. Where is it defined?

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
-This also enables cooperative multitasking of FMUs in CS.
+_[The following use cases are enabled:_
+
+* _Access to intermediate variables enables advanced Co-Simulation with interpolation/extrapolation techniques (such as polynomial extrapolation, Transmission Line Modeling (TLM) co-simulation, anti-alias filtering, smoothing of input among others)._
+* _<<IntermediateUpdateMode>> enables the same input approximation that was possible in FMI 2.0 with `fmi2SetInputDerivatives`, by evaluating the approximation polynomial and not within the FMU as in FMI 2.0._
+* _FMUs can inform the importer about an event, which occurs during an <<fmi3DoStep>> in case of CS (see <<fmi-for-co-simulation>>)._
+* _The Co-Simulation algorithm can request an <<early-return,early return>> from <<fmi3DoStep>>, because of an event between communication points (see <<early-return>>)._
+* _A model partition of an Scheduled Execution FMU calls <<fmi3CallbackIntermediateUpdate>> to signal clock activations with `clocksTicked == fmi3True`._
```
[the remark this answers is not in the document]
clagms: Fixed this.

docs/3_4_model_exchange_schema.adoc
clagms: I think it refers to the integration process. I also don't understand what the improvement in the performance is, but I rewrote it into non-normative text.

docs/3___model_exchange.adoc
clagms: Can I remove the TODO?

docs/5_1_scheduled_execution_math.adoc
```diff
@@ -1,19 +1,109 @@
...
 
-A model partition can only be activated once per activation time point latexmath:[t_i].
+A model partition can only be activated once per activation time point latexmath:[\mathbf{t}_i].
 
-Model partitions that are associated to <<outputClock,`output clocks`>> will accordingly provide the result values of the model partition's variables for the current <<outputClock>> tick time latexmath:[t_i] of the active <<outputClock>>.
+Model partitions that are associated to <<outputClock,`output clocks`>> will accordingly provide the result values of the model partition's variables for the current <<outputClock>> tick time latexmath:[\mathbf{t}_i] of the active <<outputClock>>.
 The activation of such an <<outputClock>> is not controlled by the simulation algorithm but internally by the FMU.
 
-More details can be found in <<operation-on-clocks>>, specifically <<CommunicationPointClocks>>.
+More details can be found in <<clocks>>, specifically <<clocks>>.
```
clagms, line 19: This looks like a typo? Two references to the same place?

clagms (changes requested):
Overall comments:
1. Motivation and explanation of clocked partitions are missing. This was removed. Now we expect the reader to know what a model partition means. I guess the plan is to have these added later? 
2. Lots of math was removed from ME and CS. I guess the plan is to have these added later? 
3. Some changes in this PR have little to do with Clocks.

This PR is definitely an improvement! Thank you for all the work!

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
+This argument is only valid in Model Exchange.
 
-TODO only in CS
+[[nextEventTimeDefined,`nextEventTimeDefined`]]
+* When <<nextEventTimeDefined,`nextEventTimeDefined == fmi3True`>>, the next time event is known and <<nextEventTime>> has a valid value.
 
-Once all events are handled and `discreteStatesNeedUpdate == fmi3False`, the FMU should be pushed to *Step Mode* by calling <<fmi3EnterStepMode>>, unless it requests to terminate the Co-Simulation by setting  <<terminateSimulation,`terminateSimulation`> to `fmi3True`.
-In this case, a new step can be started from the current communication point time.
+[[nextEventTime,`nextEventTime`]]
+* The return argument <<nextEventTime>> contains the absolute time of the next time event latexmath:[\mathbf{T}_\mathit{next}] if <<nextEventTimeDefined,`nextEventTimeDefined == fmi3True`>>.
```
chrbertsch, line 635: For time-based aperiodic input clocks with changing interval: How is the next interval defined. with the argement Nexst EventTime of fmi3UpdateDiscrete states? Or by a fmi3GetInterval call?

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
+|`fmi3Set{VariableType}`
+
+|Set continuous-time and discrete-time <<input,`inputs`>> latexmath:[\mathbf{u}_{c+d}(\mathbf{t})]
+|`fmi3Set{VariableType}`
 
-In this mode all continuous-time and discrete-time equations are active and the unknowns at an event can be computed and retrieved.
-The event time of a <<state event>> may be determined if a domain change of at least one event indicator is detected at the end of a completed integrator step.
+a|
+* Activate clocks and respective model partitions in latexmath:[\mathbf{f}_{\mathit{event}}]
+* Deactivate clocks and
```
chrbertsch, line 535: I do not get it. Why deactivate clocks here?

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
+`fmi3Get{VariableType}`, +
+<<fmi3GetDerivatives>>, +
+<<fmi3GetContinuousStates>>, +
+<<fmi3GetEventIndicators>> +
+
+a|
+* Evaluate latexmath:[\mathbf{f}_{\mathit{event}}], if no `fmi3GetXXX` function was called in this instant of super-dense time
+* Update previous values of discrete states: latexmath:[{}^\bullet\mathbf{x}_d:=\mathbf{x}_d]
+* Update previous values of <<buffers>>: latexmath:[{}^\bullet\mathbf{b}:=\mathbf{b}]
+* Deactivate active clocks and respective model partitions in latexmath:[\mathbf{f}_{\mathit{event}}]
```
chrbertsch, line 554: Is this done automatically by calling fmi3UpdateDiscreteStates, or is it necessary to call something?

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
+  - Reset respective clocked variables to their previous values: latexmath:[\mathbf{v}_{\mathit{d}} := {}^\bullet\mathbf{v}_{\mathit{d}}]
+  - Deactivate respective model partitions in latexmath:[\mathbf{f}_{\mathit{event}}]
+|<<fmi3SetClock>>
+
+a|
+* Activate output clocks and respective model partitions in latexmath:[\mathbf{f}_{\mathit{event}}] +
+* latexmath:[(\mathbf{y}_{c+d}, \mathbf{\dot{x}}_c, \mathbf{x}_{c+d}, \mathbf{z}, \mathbf{r}, \mathbf{w}_{c+d}, \mathbf{b}, \mathbf{T}_{\mathit{next}}, \mathbf{T}_\mathit{interval}) := \mathbf{f}_{\mathit{event}}({}^\bullet\mathbf{x}_{c+d}, \mathbf{u}_{c+d}, \mathbf{p}, {}^\bullet\mathbf{b}, \mathbf{t})]
+|<<fmi3GetClock>>, +
+<<fmi3GetInterval>>, +
+`fmi3Get{VariableType}`, +
```
chrbertsch, line 545: Is the integer part of super dense time incremented by such an fmi3Get{VariablesType} call?

chrbertsch (commented):
(Further questions and comments)

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,494 +338,574 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
+.2+|`input`
+|`changing`
+|[[changing,`changing`]]<<fmi3SetClock>> in <<EventMode>>,
+<<fmi3GetInterval>> in <<InitializationMode>> and in <<EventMode>> if and only if this clock ticked
+|simulation of the behavior of a control algorithm with variable execution time, generation of pulse sequences
+
+|`countdown`
+|[[countdown,`countdown`]]<<fmi3SetClock>> in <<EventMode>>,
+<<fmi3GetInterval>> in <<InitializationMode>> and in every <<EventMode>>
+|time-delayed actions after and event, for example, ignition signal some time after specific crank shaft angle
```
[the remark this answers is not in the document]
andreas-junghanns: Fixed.

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,494 +338,574 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
-FMI Design meeting: merge this sentence with the previous one.
-Missing the relationship to strict clocks.]_
+===== Scheduled Execution
 
+An external scheduler needs to know when a <<triggered>> <<inputClock>> depends on a <<triggered>> <<outputClock>> of the same FMU.
+When the <<outputClock>> ticks, the scheduler will trigger the corresponding <<inputClock>> according to all other scheduling constraints.
+The <<triggeredBy>> attribute of an <<inputClock>> defines such a relationship.
 
-Mathematically, the next <<clock>> tick at time instant latexmath:[t_i] is defined as:
+In case more than one clock have to be activated at the same time instant, the scheduler needs a priority to define the activation sequence.
```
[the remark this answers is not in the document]
andreas-junghanns: Fixed.

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,494 +338,574 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
+_It is not possible for an FMU exporting tool to know in advance the priorities of other FMUs that will be connected to an FMU in a simulation setup._
+_It is the task of the importer to derive a computational order for the computation of two or more distinct FMUs based on the local FMU clock priorities and input-output relationships of connected FMUs.]_
 
-.Notation description.
-[#table-notation-description]
-[cols="1,2"]
-|===
-|latexmath:[t_0]
-|The time instant in seconds at which the <<clock>> ticks the first time (= the base <<clock>> starts at the start of the simulation latexmath:[t_{\mathit{start}}] or when the controller is switched on plus an offset time latexmath:[t_{\mathit{offset}}] (xml-attributes `shiftCounter` and `resolution` in <<clock-type-definition>>).
+_[For periodic <<clock,`clocks`>> it is recommended to derive the priorities based on a rate monotonic scheduling scheme (smallest period leads to highest priority, that is, has the smallest priority value.]_
```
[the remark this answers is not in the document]
andreas-junghanns: I think this is ok as is.

docs/2_2_common_mechanisms.adoc
andreas-junghanns: I think this also applies to Step Mode in CS: it is a quality of implementation issue. If the FMU can separate state computation from output computation, why not produce outputs only when the importer requires them?
Please reopen if you do not agree.

docs/2_4_common_schema.adoc
andreas-junghanns: I don´t understand: which file? which line 19?

docs/2_4_common_schema.adoc
andreas-junghanns: @KarlWernersson , @TorstenBlochwitz , Modelicans: the causality `calculatedParameter` and the dependencyKind `dependent` - are they the same? If so, should we then not rename `calculatedParameter` to `dependentParameter` to be more consistent?

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,494 +338,574 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
+The association between <<clocked-variable,clocked variables>> and their <<clock,clocks>> is defined by the <<model-dependencies,model dependencies>>.
+<<clocked-variable,Clocked variables>> can only depend on one clock.
 
-_[TODO: A <<periodic>> <<output>> <<clock>> ticks at a priori equidistant time points._
-_The ticks of a <<periodic>> <<input>> <<clock>> are given by the importer.
-FMI Design meeting: merge this sentence with the previous one.
-Missing the relationship to strict clocks.]_
+===== Scheduled Execution
 
+An external scheduler needs to know when a <<triggered>> <<inputClock>> depends on a <<triggered>> <<outputClock>> of the same FMU.
```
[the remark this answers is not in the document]
andreas-junghanns: Yes, I like your formulation better. Fixed.

docs/2_2_common_mechanisms.adoc
andreas-junghanns: It would have to be a different subscript j because there is no synchronization of event numbers and ticks of that clock.
Therefore, this is likely ok. However, t_event does not fit when we talk about SE - and SE is listed... here, this needs clarification for the definition of t_event. I made a proposal in the PR.
Please <<resolve conversation>> if you are happy enough with that proposal.

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,522 +343,711 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
+
+Dependencies between variables of an FMU:
+
+* are helpful to classify <<algebraic-loops,algebraic loops>> across inputs and outputs when connecting FMUs.
+* associate continuous states and derivatives with each other and variables.
+* associate event indicators with variables.
+* define the sparseness when computing <<partial-derivatives,partial derivatives>>, avoiding unnecessary calls to <<fmi3GetDirectionalDerivative>> and <<fmi3GetAdjointDerivative>>.
+* assign <<clocked-variable,clocked variables>> to <<clock,clocks>>.
+
+The dependencies are encoded in the <<modelDescription.xml>> with the element <<ModelStructure,`<ModelStructure>`>> up to <<valueReference>> resolution.
```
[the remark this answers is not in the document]
andreas-junghanns: I added ", not individual elements of array variables".
I hope that clarifies this issue sufficiently.
Non-normative text could be added that if such element-wize resolution is requiered, the exporter must flatten the array.
Please reopen if this is not sufficient...

docs/2_3_common_states.adoc
```diff
@@ -1,63 +1,68 @@
...
-This state is entered when any of the following functions is called: <<fmi3InstantiateModelExchange>>, <<fmi3InstantiateCoSimulation>> and <<fmi3InstantiateScheduledExecution>>.
-The state is left by either calling <<fmi3FreeInstance>> or when any of the functions called during *FMU State Setable* returns <<fmi3Fatal>>.
-If any function called in super state *FMU State Setable* returns <<fmi3Error>>, the FMU enters state *Terminated*.
+The state <<FMUStateSetable>> is entered when any of the following functions is called: <<fmi3InstantiateModelExchange>>, <<fmi3InstantiateCoSimulation>> and <<fmi3InstantiateScheduledExecution>>.
+The state is left by either calling <<fmi3FreeInstance>> or when any of the functions called during <<FMUStateSetable>> returns <<fmi3Fatal>>.
+If any function called in super state <<FMUStateSetable>> returns <<fmi3Error>>, the FMU enters state <<Terminated>>.
 
 Allowed Function Calls::
 
 [[fmi3InstantiateModelExchange,`fmi3InstantiateModelExchange`]]
```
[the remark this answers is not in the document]
andreas-junghanns: We debated this back and forth - no matter how you do it, it breaks the pattern.
We decided that have it this way so it could be found more easily.
The anonymous start state cannot really be listed: it has no name and if you read the state chart, you can hardly see it.
Unless you are a comp-sci professional, start state is an unknown concept.

Simply moving it before the state "FMU state setable" hides it when jumping to "FMU state setable".
To list it outside the "Allowed Function Calls" might also be confusing when we convert the Function XXX:: to headings (still discussed now and then).

Conclusion: this is not an oversight, this is a compromise and we have not been able to find a better solution.

docs/2_4_common_schema.adoc
KarlWernersson: @andreas-junghanns , No they refer to different things, calculated parameter is a causality, dependencyKind is for the dependency in model structure, where you have ("constant", "fixed", "tunable") (these are all different types of linear dependency) and dependent which is all other types of dependencies nonlinear, unknown etc. 

However what @IZacharias  is refereeing to is table 19. where we list allowed combinations of casualty and variability, , both (3) **and** (4) refers to dependent parameters, which doesn't exits, it should just be,  fixed, calucatedparameter and tunalbe, calculatedparmaeter .
as it refers to those combinations in table 17.

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
+This function must be called if <<fmi3UpdateDiscreteStates>> returned with <<valuesOfContinuousStatesChanged, `valuesOfContinuousStatesChanged == fmi3True`>>.
+Can only be called in Model Exchange.
+
+Function <<fmi3GetDerivatives>>::
+Function <<fmi3GetEventIndicators>>::
+Function <<fmi3GetNumberOfContinuousStates>>::
+Function <<fmi3GetNumberOfEventIndicators>>::
+Can only be called in Model Exchange.
+
+Function <<fmi3UpdateDiscreteStates>>::
```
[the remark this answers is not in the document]
andreas-junghanns: Do you mean
1) repeat what is in the table for fmi3UpdateDiscreteStates for the function description in the text again?
2) Or define it for fmi3GetNumberOfEventIndicators?

For 1) we tried to keep the redundancy low and not repeat where not required to make the rest understandable
For 2) f_event might not be needed to compute the number of event indicators.

Can you please clarify your questions?

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
+<<fmi3GetContinuousStates>>, +
+<<fmi3GetEventIndicators>> +
+
+a|
+* Evaluate latexmath:[\mathbf{f}_{\mathit{event}}], if no `fmi3GetXXX` function was called in this instant of super-dense time
+* Update previous values of discrete states: latexmath:[{}^\bullet\mathbf{x}_d:=\mathbf{x}_d]
+* Update previous values of <<buffers>>: latexmath:[{}^\bullet\mathbf{b}:=\mathbf{b}]
+* Deactivate active clocks and respective model partitions in latexmath:[\mathbf{f}_{\mathit{event}}]
+* Increment super-dense time: latexmath:[\mathbf{t}:=(\mathbf{t}_\mathit{R}, \mathbf{t}_\mathit{I} + 1)]
+|<<fmi3UpdateDiscreteStates>>
```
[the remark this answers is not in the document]
andreas-junghanns: Ah!! We did not intend a calling order in the mathematical tables.
But this is not clear.
Can you propose how/where to clarify this?

docs/2_3_common_states.adoc
```diff
@@ -82,485 +87,644 @@ If `loggingOn == fmi3True`, the FMU enables a vendor defined set of `<LogCategor
...
-* allow FMUs to inform the importer of an event, which occurs during an <<fmi3DoStep>> in case of CS or <<fmi3ActivateModelPartition>> in case of SE (see <<co-simulation-api>>, <<Communication-during-update>> and <<IntermediateUpdateModeSE>>),
-* allow the importer in CS to request an early return from FMUs calling <<fmi3CallbackIntermediateUpdate>> with <<canReturnEarly, `canReturnEarly == fmi3True`>>, because of any event between communication points (see <<early-return>> and <<Computation-in-Co-Simulation>>).
-This also enables cooperative multitasking of FMUs in CS.
+_[The following use cases are enabled:_
+
+* _Access to intermediate variables enables advanced Co-Simulation with interpolation/extrapolation techniques (such as polynomial extrapolation, Transmission Line Modeling (TLM) co-simulation, anti-alias filtering, smoothing of input among others)._
+* _<<IntermediateUpdateMode>> enables the same input approximation that was possible in FMI 2.0 with `fmi2SetInputDerivatives`, by evaluating the approximation polynomial and not within the FMU as in FMI 2.0._
```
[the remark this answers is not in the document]
andreas-junghanns: Fixed.

docs/2_2_common_mechanisms.adoc
andreas-junghanns: TODOs in the document and another ticket (#1287) address this.

docs/3_4_model_exchange_schema.adoc
andreas-junghanns: more efficient because one call does not need to be made - that is written just before the comment.
Since this is confusing, I will remove this comment.

docs/4___co-simulation.adoc
andreas-junghanns: Indeed. I tried to improve the text there, please check and close this comment if you are happy or propose further improvements otherwise.

docs/4_2_co-simulation_api.adoc
andreas-junghanns: See above: no order intended. Solution above should fix this comment here as well, closing.

docs/5_2_scheduled_execution_api.adoc
andreas-junghanns: see above

docs/5_2_scheduled_execution_api.adoc
andreas-junghanns: This is comp-sci lingo for "uninteruptable" and needed here because SE also allows for preemtion.
I will close this because this should be clear to anyone implementing this (and `operation` and `atomic` are used in the same sentence and google produces a good wiki article).
Please propose improvement if you reopen.

docs/5_2_scheduled_execution_api.adoc
andreas-junghanns: It must be explained, that timing in all the interface types have their own definitions.
This is a good point and I think it must be normative to clarify the difference.

docs/2_1_common_math.adoc
```diff
@@ -152,30 +150,36 @@ latexmath:[\mathbf{y^{(j)}}]
...
-Every event removes automatically a previous definition of latexmath:[T_{\mathit{next}}], and it must be explicitly defined again, even if a previously defined latexmath:[T_{\mathit{next}}] was not yet reached (see <<fmi3UpdateDiscreteStates>>).
+|latexmath:[\mathbf{T}_{\mathit{next}}]
+|At an event instant, an FMU can define the next time instant latexmath:[\mathbf{T}_{\mathit{next}}], at which the next time event occurs (see also the definition of <<EventMode,events>>).
+Every event removes automatically a previous definition of latexmath:[\mathbf{T}_{\mathit{next}}], and it must be explicitly defined again, even if a previously defined latexmath:[\mathbf{T}_{\mathit{next}}] was not yet reached (see <<fmi3UpdateDiscreteStates>>).
+
+|[[relations,relations]] latexmath:[\mathbf{r}]
+|A vector of Boolean variables representing relations: latexmath:[\mathbf{r}_j := \mathbf{z}_j > 0].
+When entering <<ContinuousTimeMode>> all relations reported via the event indicators latexmath:[\mathbf{z}] are fixed and during this mode these relations are replaced by latexmath:[^{\bullet}\mathbf{r}].
```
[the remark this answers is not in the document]
andreas-junghanns: Please have someone from the Modelica community explain this to you (maybe @KarlWernersson ?).
Torsten explained this to me when we worked on the text and he is sure this is correct.
If this remains unclear, or you think we should improve, lets open a new issue and work on a PR after the big merge.

docs/2_2_common_mechanisms.adoc
```diff
@@ -252,13 +245,36 @@ This function call is allowed always and in all interface types.
...
 The standard header file as documented in this specification has version `"3.0"`, so this function returns `"3.0"`.
 
-==== Variables of a Functional Mockup Unit [[fmu-variables]]
+==== Advancing Time [[advancing-time]]
+
+This section highlights the differences of the concept of time (in general the independent variable) for the three different FMI types, ME, CS and SE.
+
+In Model Exchange, time is under the sole control of the importer and its integration algorithm.
+The model itself receives the current time to be used in its computation with <<fmi3SetTime>>.
+In fact, time is not necessarily advancing linearly as solvers might need to jump back and forth in time to localize events using zero-crossings of <<eventIndicator, event indicators>>.
```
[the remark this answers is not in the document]
andreas-junghanns: Now: "...time is not necessarily always advancing as solvers might...".

docs/2_2_common_mechanisms.adoc
```diff
@@ -252,13 +245,36 @@ This function call is allowed always and in all interface types.
...
+The importer calls <<fmi3DoStep>> with the <<currentCommunicationPoint>> and a target <<communicationStepSize>> (required to be larger than 0.0).
+During this <<fmi3DoStep>>, both importer and FMU might encounter events (or other situations) that require reduction of the <<communicationStepSize>> (potentially even down to 0.0).
+The FMU may use <<earlyReturn>> argument of the <<fmi3DoStep>> function to tell the import that the FMU needs to return earlier, and the importer may use the callback <<fmi3CallbackIntermediateUpdate>> to signal the FMU that the latter should return earlier.
+The output argument <<lastSuccessfulTime>> of <<fmi3DoStep>> allows the FMU to signal the importer its current internal time.
+
+In Scheduled Execution, time has a more discrete form.
+The scheduler of the importer activates specific tasks according to the time of the importer.
+The time itself is communicated to the FMU as <<activationTime>> argument of <<fmi3ActivateModelPartition>>.
+
+Depending on the instantiated FMI type, the importer is restricted in what functions it is allowed to call in order to drive the simulation.
```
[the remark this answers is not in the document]
andreas-junghanns: Fixed as proposed.

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,522 +343,696 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
-During this <<fmi3DoStep>>, both importer and FMU might encounter events that will reduce the <<communicationStepSize>> (potentially even down to 0.0).
-The FMU may use <<earlyReturn>> argument of the <<fmi3DoStep>> function to tell the import that the FMU needs to return earlier, and the importer may use the callback <<fmi3CallbackIntermediateUpdate>> to signal the FMU that the later should return earlier.
-This way both can determine if such a step-size reduction is required.
-The output argument <<lastSuccessfulTime>> of <<fmi3DoStep>> allows the FMU to signal the importer its current internal time.
+<<min-max,Attributes `min` and `max`>> can be defined for variables of float, integer or enumeration types.
+The question is how `fmi3Set{VariableType}`, `fmi3Get{VariableType}` shall utilize this definition.
+There are several conflicting requirements: +
+Avoiding forbidden regions (for example, if `u` is an <<input>> and "sqrt(u)" is computed in the FMU, `min = 0` on `u` shall guarantee that only values of `u` in the allowed regions are provided).
```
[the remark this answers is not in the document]
andreas-junghanns: These conflicts aren´t really itemizable - as the text suggests.
I would not like to mess with this part either - it took us a while to make it readable - even if it isn´t pretty.
If you reopen, please make a concrete proposal how to change it.

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,522 +343,696 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
-* Partial evaluation for a subactive <<clock>>: <<state,`states`>> are left unmodified, only outputs are computed.
-Current input values are taken into account.
-+
-_[This mode can be used to compute partial derivatives_ latexmath:[\partial y_j / \partial x_{j-1}] _.]_
+In all tables describing the mathematical model of a state (e.g. <<table-math-model-exchange,<<ContinuousTimeMode>>>>), for notational convenience one function (e.g. latexmath:[\mathbf{f}_\mathit{cont}]) is defined to compute all output arguments from all inputs arguments.
+In an FMU, every output argument can be computed by its specific get function.
+Additionally, the output argument need not be a function of all input arguments, but of only a subset from it, as defined in the <<modelDescription.xml>> under `<ModelStructure>`.
 
-===== Communication Point Clocks [[CommunicationPointClocks]]
+_[The functions above have the slight drawback that values must always be copied._
```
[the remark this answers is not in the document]
andreas-junghanns: I think this is where it belongs: the caching and copy-work incurred by get and set functions is in the cache section of the text.

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,522 +343,711 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
+Output clocks::
+[[outputClock, `output clock`]]There is only one <<clock>> type that is an <<outputClock>>: The <<triggered>> <<outputClock>>.
+We still refer to it as <<outputClock>> for simplicity.
+The importer calls <<fmi3GetClock>> to inquire the clock activation state.
 
-Since <<outputClock,`output clocks`>> are ticking based on model internal information, it is required in case of FMI for Co-Simulation to signal the ticking of an <<outputClock>> to the simulation environment during a <<fmi3DoStep>> or <<fmi3ActivateModelPartition>> computation.
-In most cases _[TODO: not all cases? which cases? if the time of the clock tick is a communication point already?]_ the signaling of an <<outputClock>> tick requires creating a communication point for exchanging additional information (i.e. variable values) with the simulation environment.
-Additional cases need to be handled if a <<clock>> tick is not associated to an event that requires the creation of a communication point for the model partition that generated the event. _[TODO: what?]_
++
+.Overview of Clock Types.
```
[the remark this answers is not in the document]
andreas-junghanns: This was done on purpose: input and output are not really defining features of clocks (especially because there is only one output clock). Separating these IO properties from the important properties that are then explained in the table was intentional. Maybe you can make a concrete proposal how we can make this clear after the merge?

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,522 +343,711 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
 
-|aperiodic
-|aperiodic, periodic, strictly periodic
+Calculated periodic clocks::
+are time-base periodic clocks where the interval and <<shift>> depend on <<fixed>> <<parameter, parameters>>.
+The importer must use <<fmi3GetInterval>> and <<fmi3GetShift>> to retrieve the clock interval and <<shift>> in <<InitializationMode>>.
 
-|periodic
-|-
+Tunable periodic clocks::
```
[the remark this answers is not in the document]
andreas-junghanns: We sorted the equations to apply to sub-types of clocks. If we list equations for periodic clocks, they apply to all its sub-types. Right now I still think we did this correctly (except for the GetInterval on triggered clocks, we are still working with Matthias and Irina on that).

docs/2_2_common_mechanisms.adoc
```diff
@@ -330,522 +343,711 @@ Set <<parameter,`parameters`>>, <<input,`inputs`>>, and <<start>> values, and re
...
-_Does the importer have to call fmi3SetInterval for aperiodic clocks at each clock tick if the clock should tick with the same interval? Or will there be only one clock tick per fmi3SetInterval call?
-FMI Design meeting: remark that set interval functions are informative to design model based controllers.
-The environment is the one that will tick.
-]_
+===== Model Partitions and Clocked Variables [[clocked-variable]]
 
-A <<clock>> is <<strict,strictly>> <<periodic>> if latexmath:[\Delta T] is fixed for an FMU and cannot be changed.
-The <<clock>> is periodic, if latexmath:[\Delta T] can be defined before calling `fmi3EnterInitializationMode` via <<fmi3SetIntervalDecimal>> or <<fmi3SetIntervalFraction>>.
-The <<clock>> is <<periodic,aperiodic>> if latexmath:[\Delta T] is not constant during a simulation run (<<fmi3SetIntervalDecimal>> or <<fmi3SetIntervalFraction>> can be called for that <<clock>> during *Event Mode*).
+A <<clock>> activates its specific model partition.
```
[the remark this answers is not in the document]
andreas-junghanns: It is in the glossary. I added a link to there. But not throughout the document - I rarely did for glossary entries.
I removed from the glossary where we use a term that can be googled easily and we do not use in a different form of that definition.
Terms in the glossary are general concepts that require a bit of explanation as used in the context of FMI.
Technical FMI terms defined in the document are generally linked.

If you have an idea how to improve this, please open a new issue.

docs/2_3_common_states.adoc
```diff
@@ -1,63 +1,68 @@
...
 
+To define allowed calling sequences of FMI functions, state machines are used.
 All FMI interface types share a number of states in their respective state machines.
-This chapter describes these common modes.
-FMI specific state-machine modes will be described in their respective chapters.
+This chapter describes these common states used in at least two of the interface types.
+FMI specific state-machine states will be described in their respective chapters.
+Each state of the state machine corresponds to a certain phase of a simulation.
 
 .Common calling sequence for C functions of common states for all three FMI types.
```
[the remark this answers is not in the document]
andreas-junghanns: The title says: common states, the state diagram is obviously not complete.
The way it is shown here makes it simple to export from the diagram editor and since the arrows do not match up perfectly with the grey state for all interface types, the grey state could cause confusion (see how to enter Event Mode).

docs/2_3_common_states.adoc
andreas-junghanns: We support iteration for solving the initial value problem. For that, those equations have to be computed.

docs/2_3_common_states.adoc
andreas-junghanns: The cell below has the definition.

docs/5_1_scheduled_execution_math.adoc
andreas-junghanns: Fixed.

docs/2_3_common_states.adoc
andreas-junghanns: We tried to make this clear in the table listing the clocks: for aperiodic clocks, it is GetInterval and depending on the kind the importer has to call it in every Event Mode or only in those where the clock ticked.

nextEventTime argument of UpdateDiscreteStates is used for time events only. I think this is what we have written there.

docs/2_3_common_states.adoc
andreas-junghanns: Because clocks might be conditional on some inputs. During numeric iteration in one super-dense time instant, this might activate and deactivate a clock. If not clear, please suggest concrete change.

docs/2_3_common_states.adoc
andreas-junghanns: This is the definition of what the function does that is on the right column.
This is the mathematical description of the function´s behaviour/action.

docs/2_3_common_states.adoc
andreas-junghanns: No, the time is only updated in fmi3UpdateDiscreteStates - as you see in its line of the table.

docs/2_3_common_states.adoc
chrbertsch: > nextEventTime argument of UpdateDiscreteStates is used for time events only.

So we use nextEventTime only for time events that are not associated with a time-based clock?

docs/2_2_common_mechanisms.adoc
andreas-junghanns: Do we have to state this specifically if we refer to caching? Caching by definition guarantees this. If we make an explicit comment here, this might be read as different from the standard comp-sci use. Please re-read and if you still think this is needed, please suggest exactly where to please what kind of comment.

docs/2_3_common_states.adoc
andreas-junghanns: Yes. @TorstenBlochwitz - please confirm.

docs/2_4_common_schema.adoc
andreas-junghanns: Fixed in the table (now 23).

docs/2_4_common_schema.adoc
andreas-junghanns: I now get it (after fixing (3) and (4) I found the reference and understood (19)): I dropped (19) and turned the (19) in the table into an (18). Restrictions to local variables apply still, no need to explain again for clocks.

docs/2_2_common_mechanisms.adoc
andreas-junghanns: Changed tactics: I close, you reopen if unhappy (I am drowning in change requests, sorry.)

docs/3___model_exchange.adoc
andreas-junghanns: @chrbertsch : There are two TODOs for you in the text (search for "TODO: Christian:"), can you please look at both and either remove or fix?

docs/4___co-simulation.adoc
andreas-junghanns: change of tactics: I am closing, if you are unhappy, please reopen.

docs/2_3_common_states.adoc
chrbertsch: Then I would explicitely state this, something like: 
For time events associated to time-based clocks, fmi3GetInterval shall be used to to to communicate the next event time instead of  nextEventTimeDefined and  nextEventTime.

docs/2_2_common_mechanisms.adoc
chrbertsch: As a "get" can trigger calculations: I think the question is, if the FMU _may_ apply caching or _must_ apply caching.
Is it guaranteed, that you get the same result if you call a get function two times (w/o a set, DoStep, UpdateDiscrete States or activateModelPartition call inbetween).

docs/2_3_common_states.adoc
chrbertsch: (My question was related to interpreting this table as describing a calling order. Then it would not have made sense to deactivate clocks at this point of time)

docs/2_3_common_states.adoc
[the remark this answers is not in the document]
chrbertsch: We could write something like this before the table: 
_In the following table the relations between the governing equations and assignments to the FMI API functions are given (but not a calling sequence)_
But we would have to write this before every of these tables for the states/superstates.

oliverkotte (commented):
In section 2.2.9.4., it should read:
qualifier describes how to treat the interval and intervalCounter argument
not
qualifier describes how to treat the shift and shiftCounter argument

docs/2_3_common_states.adoc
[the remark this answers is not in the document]
chrbertsch: I meant 1). Closing

docs/2_3_common_states.adoc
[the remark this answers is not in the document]
chrbertsch: I suggest  to create here a special section:
Allowed Function Calls to enter this Super State.

chrbertsch:
> 
> 
> In section 2.2.9.4., it should read:
> qualifier describes how to treat the interval and intervalCounter argument
> not
> qualifier describes how to treat the shift and shiftCounter argument

fixed by https://github.com/modelica/fmi-standard/pull/1272/commits/2fe49d4ea2aa4eadc62f867ad2dc276e67abfc98

docs/2_3_common_states.adoc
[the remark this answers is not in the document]
andreas-junghanns: Fixed as proposed.

docs/2_3_common_states.adoc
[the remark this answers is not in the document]
andreas-junghanns: Let´s open an extra issue for this, I want to merge this PR now.

docs/2_2_common_mechanisms.adoc
andreas-junghanns: Changed:
This enables computation on demand.
To:
This enables computation on demand with subsequent calls returning the same value until some set operation requires re-computation.

docs/2_3_common_states.adoc
andreas-junghanns: No, sorry I did not make this clear: Time events don´t have to be associated with a clock. Time events can be purely internal and still trigger event handling. We will have to find a central place where to describe this. Please open another issue, I want to finally merge this PR.

chrbertsch: Reading 2.3.2.1, I now understand it; no PR needed from my side.

PR #1273, opened 2020-12-22
Add Name to sourcefileset

MBlesken:
Hello. The motivation to add a name attribute to a soucefileset is the following use case: A consumer wants to compile a source file set and create a library for which he needs to choose a name. If provided by the source file set the name attribute can be used to name the library. I would like to invite everyone to express their opinion on this small extension and hopefully approve the idea. Thank you.

### DIFF
--- a/docs/2_6_source_code.adoc
+++ b/docs/2_6_source_code.adoc
@@ -40,6 +40,9 @@ An importer of the FMU has to regard every `<SourceFileSet>` of the matching `<B
 |Attribute
 |Description
 
+|`name`
+|Name of the `<SourceFileSet>`
+
 |`language`
 |Language of the source files (e.g. `C99`, `C++11`)
 
--- a/schema/fmi3BuildDescription.xsd
+++ b/schema/fmi3BuildDescription.xsd
@@ -79,6 +79,7 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 										</xs:element>
 										<xs:element ref="Annotations" minOccurs="0"/>
 									</xs:sequence>
+									<xs:attribute name="name" type="xs:normalizedString"/>
 									<xs:attribute name="language" type="xs:normalizedString"/>
 									<xs:attribute name="compiler" type="xs:normalizedString"/>
 									<xs:attribute name="compilerOptions" type="xs:string"/>


andreas-junghanns (changes requested):
Matthias: I support this idea. However, should we also request the name to be a C-identifier? This way, the use case you sketch, would actually be guaranteed to work.

MBlesken:
> Matthias: I support this idea. However, should we also request the name to be a C-identifier? This way, the use case you sketch, would actually be guaranteed to work.

@andreas-junghanns Choosing a C-identifier would help regarding this use case but the consumer could also create a C-conform name for a library from a string input. I copied the name attribute from the name attribute of the library where it is defined as xs:normalizedString. I also observe that other attributes like modelIdentifier are xs:normalizedString. So I decided to stay consistent to those.

andreas-junghanns:
> 
> 
> > Matthias: I support this idea. However, should we also request the name to be a C-identifier? This way, the use case you sketch, would actually be guaranteed to work.
> 
> @andreas-junghanns Choosing a C-identifier would help regarding this use case but the consumer could also create a C-conform name for a library from a string input. I copied the name attribute from the name attribute of the library where it is defined as xs:normalizedString. I also observe that other attributes like modelIdentifier are xs:normalizedString. So I decided to stay consistent to those.

@MBlesken : I check https://www.oreilly.com/library/view/xml-schema/0596002521/re88.html to see what xs:normalizedString means and it just means that white-spaces are all turned to space (0x20). This is quite liberal and might fail to work in many tool chains. I also agree that consistency is important - but maybe we should be more strict elsewhere as well?

andreas-junghanns:
@t-sommer : You build the first version of the sourcefileset mechanism. Can you please comment if you see stricter naming rules as helpful? Or do we need to be liberal here for some good reasons?

t-sommer (changes requested):
I'm in favor of the idea, but we should add some normative text that explains the purpose of this attribute and how to use it. If it's supposed to be used as the filename of a static library, `normalizedString` should be okay.

andreas-junghanns:
> 
> 
> I'm in favor of the idea, but we should add some normative text that explains the purpose of this attribute and how to use it. If it's supposed to be used as the filename of a static library, `normalizedString` should be okay.

Definition` of normalizedString: `The lexical space of xs:normalizedString is unconstrained (any valid XML character may be used), and its value space is the set of strings after whitespace replacement (i.e., after any occurrence of #x9 (tab), #xA (linefeed), and #xD (carriage return) have been replaced by an occurrence of #x20 (space) without any whitespace collapsing).`

And yes, file names may contain spaces. But have you tried linking library files with spaces in "your tool"? I would be surprised if that works out-of-the-box...

Would `xs:Name` not be better (also in other, similar, places)?

MBlesken:
@andreas-junghanns I wouldn't mind changing the type to xs:Name if @IZacharias and @t-sommer agree. I would rather not want to change "other, similar, places" in this PR though. What do you think?

@andreas-junghanns @IZacharias @t-sommer: Concerning @t-sommer 's comment "we should add some normative text that explains the purpose of this attribute and how to use it": I intentionally did not mention the use case in the specification. Do you think the use case "creating a library form the source file set" should be (briefly) mentioned in the specification? I tough such things should be excluded. ... I' afraid that this might trigger subsequent discussions. What do you think?

t-sommer:
> I would rather not want to change "other, similar, places" in this PR though. What do you think?

I suggest we do this in a separate PR where we change all XSD types consistently.

andreas-junghanns:
If we restrict the type to xs:Name, a non-normative explanation with example use case might be in order.
But: I am not a XML expert and if other places could conceivably benefit from similar restrictions, then I agree with @MBlesken to do this together in one commit (both XSD and text, including explanation as non-normative text).

IZacharias (approved):

PR #1282, opened 2021-01-16
Add ABI version for VS 2019

beutlich:
This also generalizes the MSVC version for `abi_ver` = 141.


andreas-junghanns (commented):
Why are you using .X for VS 2017 and VS 2019 and not all the other versions?

beutlich:
According to https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering there always is X=0 for all other versions.

t-sommer (approved):

PR #1291, opened 2021-02-07
Fix compression method name `deflate` -> `deflated`

beutlich:
This PR fixes two tiny issues on the FMU compression:
* The compression property (as for example returned by `unzip -Zsvh HybridVehicle.fmu`) is called `deflated`.
* The compression method is not a per-archive property, but a per-file property.

Additionally, it is not clear which zlib tool is meant, since usually zlib is a library (see https://en.wikipedia.org/wiki/Zlib).

### DIFF
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -10,7 +10,7 @@ The behavior of those functions is unspecified, so while calling environments ca
 The extension of the ZIP file must be `.fmu` _[, for example, `HybridVehicle.fmu`]_.
 The compression method for all files stored in the ZIP archive must be either 8 (`deflate`), or 0 (`store`).
 Only files stored using compression method 8 (`deflate`) may be stored with general purpose bit 3 set.
-The version needed to extract field of the archive must not be higher than 2.0, and encryption must not be employed.
+The field `version needed to extract` of the archive must not be higher than 2.0, and encryption must not be employed.
 The archive may not be a split or spanned ZIP archive.
  _[These restrictions ensure broad compatibility of the archive with common ZIP processing tools and libraries.)]_
 
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -8,7 +8,7 @@ The FMU must implement all common API functions according to <<general-mechanism
 Especially it is required that all functions that are part of the specified FMI interface type are present, even if they are only needed for optional capabilities that the FMU does not support.
 The behavior of those functions is unspecified, so while calling environments can rely on the functions being present, they cannot rely on any particular behavior for functions only needed for capabilities the FMU does not support.
 The extension of the ZIP file must be `.fmu` _[, for example, `HybridVehicle.fmu`]_.
-The compression method used for the ZIP file must be `deflate` _[(most free tools, such as zlib, offer only the common compression method `deflate`)]_.
+The compression method for all files stored in the ZIP file must be `deflated` _[(most free tools, such as zlib, offer only the common compression method `deflated`)]_.
 
 
 _[Note: especially section 4.4.17 of https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT[the ZIP format specification] states that backslashes "\" are forbidden as path separator, only forward slashes "/" are allowed._ +
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -8,8 +8,11 @@ The FMU must implement all common API functions according to <<general-mechanism
 Especially it is required that all functions that are part of the specified FMI interface type are present, even if they are only needed for optional capabilities that the FMU does not support.
 The behavior of those functions is unspecified, so while calling environments can rely on the functions being present, they cannot rely on any particular behavior for functions only needed for capabilities the FMU does not support.
 The extension of the ZIP file must be `.fmu` _[, for example, `HybridVehicle.fmu`]_.
-The compression method for all files stored in the ZIP file must be `deflate` _[(most free tools, such as zlib, offer only the common compression method `deflate`)]_.
-
+The compression method for all files stored in the ZIP archive must be either 8 (`deflate`), or 0 (`store`).
+Only files stored using compression method 8 (`deflate`) may be stored with general purpose bit 3 set.
+The version needed to extract field of the archive must not be higher than 2.0, and encryption must not be employed.
+The archive may not be a split or spanned ZIP archive.
+ _[These restrictions ensure broad compatibility of the archive with common ZIP processing tools and libraries.)]_
 
 _[Note: especially section 4.4.17 of https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT[the ZIP format specification] states that backslashes "\" are forbidden as path separator, only forward slashes "/" are allowed._ +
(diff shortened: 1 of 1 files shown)


docs/2_5_fmu_distribution.adoc
```diff
@@ -8,7 +8,7 @@ The FMU must implement all common API functions according to <<general-mechanism
 Especially it is required that all functions that are part of the specified FMI interface type are present, even if they are only needed for optional capabilities that the FMU does not support.
 The behavior of those functions is unspecified, so while calling environments can rely on the functions being present, they cannot rely on any particular behavior for functions only needed for capabilities the FMU does not support.
 The extension of the ZIP file must be `.fmu` _[, for example, `HybridVehicle.fmu`]_.
-The compression method used for the ZIP file must be `deflate` _[(most free tools, such as zlib, offer only the common compression method `deflate`)]_.
+The compression method for all files stored in the ZIP file must be `deflated` _[(most free tools, such as zlib, offer only the common compression method `deflated`)]_.
```
pmai, line 11: I would suggest the following wording, since very small files/directory entries will use method 0 (stored), and this specifies the other requirements, like no ZIP64, no encryption, no splitting, etc., thus achieving the broard compatibility that was intended:

```suggestion
The compression method for all files stored in the ZIP archive must be either 8 (`deflated`), or 0 (`stored`).
The version needed to extract filed of the archive must not be higher than 2.0, and encryption must not be employed.
The archive may not be a split or spanned ZIP archive.
 _[These restrictions ensure broad compatibility of the archive with common ZIP processing tools and libraries.)]_
```

beutlich: I am fine with compression and splitting, but I even thought ZIP encrpytion is legal and the simplest way to achieve some kind of FMU encryption (between trusted partners).

pmai: The problem with allowing encryption is that all FMI implementations would then have to support encryption, which currently is definitely not what is implemented, nor what many common libraries and tools support out of the box.

Furthermore legacy ZIP encryption (which is what is in ZIP 2.0) is basically useless due to e.g. Biham and Kocher's known plaintext attack, which just requires 12 bytes of known plaintext: This is trivial for FMUs that must contain the modelDescription.xml with its XML preamble, or many other sources of known plaintext. See https://github.com/kimci86/bkcrack for working code.

And all advanced encryption support in ZIP requires newer versions of the ZIP specification, thereby more or less placing you outside the widely supported subset anyways (they employ whole directory/archive encryption concepts, which are way more secure, but not widely supported).

Nothing prevents partners/tools that want to support more than the basic spec to allow encryption (and/or more compression algos, ZIP64, etc.), but requiring this across the board seems not useful. Furthermore there are better mechanisms inside FMUs to encrypt and/or protect content in other ways, which do not require the support by the importer.

Just my two centieuros...

KarlWernersson: Would it be of interest to update 1.0 and 2.0 to support store option?
I recently had to change our zip routine as it used store on certain files (zip files)  where the deflate was pointless. However an importer pointed out to me that that was illegal and their tool broke since their unzip routine wasn't  able to handle the split.
So if we will support this for 3.0 should we update 1.0,x and 2.0.x in the same way?

t-sommer (changes requested):
Java's `ZipInputStream` can't handle combinations of `stored` and `deflated`, so I suggest not to allow it. See https://github.com/CATIA-Systems/FMIKit-Simulink/issues/233 and https://bugs.openjdk.java.net/browse/JDK-8143613.

t-sommer:
> * The compression property (as for example returned by `unzip -Zsvh HybridVehicle.fmu`) is called `deflated`.

The [.ZIP File Format Specification](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) calls the method `deflate`.

beutlich:
OK, never mind. I guess the important message this that **all files** need to be `deflated` while directories are `stored`.

pmai:
> Java's `ZipInputStream` can't handle combinations of `stored` and `deflated`, so I suggest not to allow it. See [CATIA-Systems/FMIKit-Simulink#233](https://github.com/CATIA-Systems/FMIKit-Simulink/issues/233) and https://bugs.openjdk.java.net/browse/JDK-8143613.

Actually it does support stored and deflated, it just does not support the use of GP bit 3 (used for example in streaming compression to avoid having to write length and CRC into the local header) with stored. Since nearly all compression utilities create directory (and other very small entries, like 1-byte files, etc.) using stored as the compression method, unless you hand-code stuff, this is the way nearly all real FMUs have been distributed. E.g. all of the ReferenceFMUs in https://github.com/modelica/Reference-FMUs/releases/download/v0.0.4/Reference-FMUs-0.0.4.zip would be illegal under the strict reading of FMI 1.0/2.0, which is just stupid.

So if anything this is already the state of the art I'd say, and we should fix the FMI 3.0 definition to match actual practice (and the ZIP specification). If we want to we can point out the ZipInputStream and PKZIP 2.04g incompatibility, and suggest not to use that combination, but other than that I see no reason to disallow FMUs created using normal ZIP tools to work.

pmai:
> > * The compression property (as for example returned by `unzip -Zsvh HybridVehicle.fmu`) is called `deflated`.
> 
> The [.ZIP File Format Specification](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) calls the method `deflate`.

It actually calls it both ways: While the algorithm itself is called deflate nowadays more or less universally (also due to its retro-standardization as RFC1951), it has the following entries in its compression method table:

4.4.5 compression method: (2 bytes)

        0 - The file is stored (no compression)
        1 - The file is Shrunk
        2 - The file is Reduced with compression factor 1
        3 - The file is Reduced with compression factor 2
        4 - The file is Reduced with compression factor 3
        5 - The file is Reduced with compression factor 4
        6 - The file is Imploded
        7 - Reserved for Tokenizing compression algorithm
        8 - The file is Deflated
        9 - Enhanced Deflating using Deflate64(tm)
        ...

docs/2_5_fmu_distribution.adoc
pmai: > So if we will support this for 3.0 should we update 1.0,x and 2.0.x in the same way?

I think we should so clarify, since otherwise nearly no compliant FMUs are out there: Directory entries are usually stored as stored (sic) in nearly all zip utilities (dito for very small files, BTW), so this would make all manually built FMUs and most other FMUs illegal...

andreas-junghanns:
@t-sommer : Please comment on @pmai ´s last change proposal - if you are ok, please merge.

andreas-junghanns (approved):

beutlich:
Please bear in mind, that the reference to "zlib tool" is still not resolved.

pmai:
Since the currently proposed fix does not actually fix the underspecification problem, I've taken the liberty of pushing a more inclusive commit. An alternative way would be reference ISO 21320 (https://standards.iso.org/ittf/PubliclyAvailableStandards/c060101_ISO_IEC_21320-1_2015.zip), which provides a corresponding profile, however it is not restricted to 2.0.

andreas-junghanns (changes requested):
Please check wording: "extract field"? fields? a field?

pmai:
@andreas-junghanns it's the "version needed to extract" field, however I'm unsure how to phrase this more nicely... ;)

t-sommer (approved):

andreas-junghanns (approved):

PR #1294, opened 2021-02-10
Clarify that in "equations and actions" tables there is no order

chrbertsch:
a proposol to fix  the topic of this comment: https://github.com/modelica/fmi-standard/pull/1272#discussion_r573206441 
(I falsely read the "equations and actions" tables to suggest a specfic calling order of the associated API functions.)

### DIFF
--- a/docs/2_3_common_states.adoc
+++ b/docs/2_3_common_states.adoc
@@ -1,10 +1,10 @@
 === State Machine and Semantics [[common-state-machine]]
 
-To define allowed calling sequences of FMI functions, state machines are used.
+To define allowed calling sequences of FMI functions, state machines are used. Each state of the state machine corresponds to a certain phase of a simulation.
 All FMI interface types share a number of states in their respective state machines.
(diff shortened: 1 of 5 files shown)


andreas-junghanns:
Maybe the confusion can be reduced by switching the table columns.
This way it is clear that the C function (then left side), trigger the math/actions given on the right side.
@TorstenBlochwitz : Do you agree?

TorstenBlochwitz:
I agree in principle, but I am not sure that this fixes the problem. In Co-Simulation an order is to be considered.

However, we explain this at several place (at the beginning of the Model Exchange and Co-Simulation sections, in the partial evaluation section...). So I think at one point, this mus be sufficient!

andreas-junghanns (approved):

TorstenBlochwitz (approved):

andreas-junghanns:
We made the table heading a link, rather than the extra (ugly) text.

PR #1299, opened 2021-02-16
Clarify v_rest and v_known

andreas-junghanns:
We have improved the text and suggest this to be sufficient, please re-read and comment.

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -1046,8 +1046,7 @@ Details about which variables are in latexmath:[\mathbf{v}_{\mathit{known}}] are
 * latexmath:[{\mathbf{v}_{\mathit{rest}}}] is the set of <<input>> variables and discrete-time states of function *f* which cannot be set by the importer in the current state, but in other states _[for example, discrete-time <<input,`inputs`>> in <<ContinuousTimeMode>>]_.
 
 _[The variable relationships are different in different states._
-_For example, during <<ContinuousTimeMode>>, the partial derivate of a continuous-time output *y* with respect to discrete-time <<input,`inputs`>> is undefined, because discrete-time <<input, `inputs`>> cannot be set between events._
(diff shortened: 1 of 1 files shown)


docs/2_2_common_mechanisms.adoc
```diff
@@ -1040,15 +1040,14 @@ where
...
+* latexmath:[{\mathbf{v}_{\mathit{rest}}}] is the set of <<input>> variables and discrete-time states of function *f* which cannot be set by the importer in the current state, but in other states _[for example, discrete-time <<input,`inputs`>> in <<ContinuousTimeMode>>]_.
 
 _[The variable relationships are different in different states._
-_For example, during <<ContinuousTimeMode>>, a continuous-time output y does not depend on discrete-time <<input,`inputs`>> (because they are held constant between events)._
-_However, at <<EventMode>>, y depends on discrete-time <<input,`inputs`>>._
-_The function may compute the directional derivatives by numerical differentiation taking into account the sparseness of the equation system, or (preferred) by analytic derivatives.]_
+_For example, during <<ContinuousTimeMode>>, the partial derivate of a continuous-time output *y* with respect to discrete-time <<input,`inputs`>> is zero, because discrete-time <<input, `inputs`>> are constant between events._
```
chrbertsch, line 1049: Wouldn't a discrte time input in ContinuouTimeModel belong to v_rest, and thus a partial derivative w.r.t. to this variable would not be defined (a call to get it not allowed) instead of being zero?

chrbertsch (approved):

chrbertsch:
As I did the last commit, could someone els please have a look before merging this?

andreas-junghanns:
I removed your last sentence: it is not about y depending on a discrete input - that is always the case (or not), it is about if the variable output depends on can be changed, or not (in the current state). I think the example given is enough. If a second sentence is needed, it should be rewritten.

Back to you @chrbertsch : I cannot approve with my last change.

chrbertsch (approved):

PR #1306, opened 2021-02-17
Turn up into Up for Functional Mock-Up

andreas-junghanns:
Out logo uses "Functional Mock-Up Interface".
The web page almost always uses "Functional Mock-up Interface" in the text.

Do we want to use the logo way in 3.0?
This PR has the required changes.


chrbertsch (approved):
no strong opinion, but this makes it consistent with the logo.

Issue #1102, opened 2020-07-31
Change definition of fmi3Boolean from "int" to "char"


t-sommer:
since the arguments brought up in #48 don't hold any more for FMI 3 with the introduction of contiguous arrays and new data types.

andreas-junghanns:
char or unsigned char?

andreas-junghanns:
WEB design meeting: we should use int32, for packing int8 is the better choice. (and not flip-flop the definition)

Also, we should fix:
 - fmi3Clock, fmi3Boolean and fmi3Byte to use uint8_t

andreas-junghanns:
Pierre: fmi3Boolean should change to _Bool.

pmai:
See #1311, which should cover this; fmi3Clock is _Bool as well, since I could not for now find a good reason for it to be something else. If we move to ternary clocks, we would have to change in non-backward compatible ways anyways, so ...

andreas-junghanns:
> 
> 
> See #1311, which should cover this; fmi3Clock is _Bool as well, since I could not for now find a good reason for it to be something else. If we move to ternary clocks, we would have to change in non-backward compatible ways anyways, so ...

Should we then remove the macro definitions
#define fmi3ClockActive   1
#define fmi3ClockInactive 0
?
Which would also mean that we have to change the text removing references and explaining that an "true" clock is an "active" clock...

PR #1311, opened 2021-02-19
Use C99 int/bool types for all types

pmai:
Given that we already depend on the C99 integer types for our integer
types, we might as well switch to use the C99 _Bool type and integer
types for all other typedefs to make the interface ABI more stable.

This reflects the discussions in the design meeting on 2021-02-19

Checklist

* [ ] Used a [personal fork](https://docs.github.com/en/free-pro-team@latest/github/getting-started-with-github/fork-a-repo) of the repository to propose changes.
* [ ] [Built the specification](https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#building-the-specification-document).
* [ ] [Re-generated schema figures](https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#changing-the-xsd-schemas).
* [ ] [Linted the documents](https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#building-the-specification-document).

See also PR #1102 which started the discussion.

### DIFF
--- a/headers/fmi3PlatformTypes.h
+++ b/headers/fmi3PlatformTypes.h
@@ -54,7 +54,7 @@ typedef void*           fmi3FMUState;             /* Pointer to internal FMU sta
 /* end::FMUState[] */
 
 /* tag::ValueReference[] */
-typedef unsigned int    fmi3ValueReference;       /* Handle to the value of a variable */
+typedef        uint32_t fmi3ValueReference;       /* Handle to the value of a variable */
 /* end::ValueReference[] */
 
 /* tag::VariableTypes[] */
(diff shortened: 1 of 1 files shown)


andreas-junghanns (changes requested):
I think fmi3Clock should be uint8_t as well, because we define two macros with values to use (as opposed to using true and false).

t-sommer:
> I think fmi3Clock should be uint8_t as well, because we define two macros with values to use (as opposed to using true and false).

Did you mean `uint32_t`?

andreas-junghanns:
> > I think fmi3Clock should be uint8_t as well, because we define two macros with values to use (as opposed to using true and false).
> 
> Did you mean `uint32_t`?

I am old, I usually try to save bytes, but maybe uint32_t is the faster choice in todays platforms (seeing also that _Bool is now 32 bit). So, uint32_t would be ok with me.

t-sommer (approved):

andreas-junghanns:
This PR was merged with fmi3Clock still being of type _Bool.
What about fmi3ClockActive and fmi3ClockInactive macros?
These should not be assigned as values of a now _Bool clock.

@t-sommer: Can you at least comment here why I should not care about this inconsistency?

pmai:
> This PR was merged with fmi3Clock still being of type _Bool.
> What about fmi3ClockActive and fmi3ClockInactive macros?
> These should not be assigned as values of a now _Bool clock.

Why? _Bool is expected to take 0 and 1 as literal values, it only differs from int in that it will normalize all non-0 values to 1 on assignment.

We could include stdbool.h and use true and false but those are just macros and would needlessly pollute the namespace...

PR #1315, opened 2021-02-19
Move Config and Reconfig Modes out; fix state descriptions

andreas-junghanns:


TorstenBlochwitz (approved):
Close to perfect: the graphics of the state machines could be slightly improved.

PR #1319, opened 2021-02-26
Clarify clock interval computation in Event Mode

andreas-junghanns:


TorstenBlochwitz (approved):
A small step for a man...

PR #1320, opened 2021-02-26
remove clocksAboutToTick from doStep()

### LINKED ISSUE #1310: Mechanism for FMU in Event Mode to communicate output clock tick to importer
Neither of the two mechanisms for an FMU to communicate to the importer that a triggered output clock is about to tick
- early return from a fmi3DoStep with clocksAboutToTick == fmi3True, and

- calling fmi3CallbackIntermediateUpdate with clocksTicked == fmi3True

applies in EventMode. How to communicate to the importer that a triggered output clock ticks in Event Mode, so that the importer calls getClocks() on the FMU and gets to know about the clock tick?

In the motif below, FMU2...
1. gets pushed in Event Mode by the importer
2. gets its triggered input clock activated by the importer
3. runs updateDiscreteStates
4. wants to activate its output clock and communicate this to the importer - but how?

```
FMU1                                FMU2
===================                 ==================
aperiodic changing input clock
triggered output clock         ->   triggered input clock
triggered input clock          <-   triggered output clock
```

oliverkotte:
Fixes #1310

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -671,14 +671,14 @@ The importer calls <<fmi3GetClock>> to inquire the clock activation state.
 |`triggered`
 |[[triggered,`triggered`]]<<fmi3SetClock>> in <<EventMode>>,
 <<fmi3ActivateModelPartition>> in <<ClockActivationMode>>,
-arguments <<clocksAboutToTick>> (<<fmi3DoStep>>) and <<clocksTicked>> (<<fmi3CallbackIntermediateUpdate>>)
+argument <<clocksTicked>> of <<fmi3CallbackIntermediateUpdate>>
 |triggered by a hardware interrupt of an embedded system, e.g. a control algorithm, triggered by a crankshaft angle
 
 |output clock
 |`output`
 |`triggered`
 |<<fmi3GetClock>> in both <<EventMode>> and <<IntermediateUpdateMode>>,
-arguments <<clocksAboutToTick>> (<<fmi3DoStep>>) and <<clocksTicked>> (<<fmi3CallbackIntermediateUpdate>>),
+and argument <<clocksTicked>> of <<fmi3CallbackIntermediateUpdate>>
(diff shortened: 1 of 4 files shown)


andreas-junghanns (approved):
Looks good. We are now a bit brief on getClock in Event Mode. Is it clear now from the text that we have to call getClock in all instants of super-dense time?

docs/2_2_common_mechanisms.adoc
```diff
@@ -671,14 +671,14 @@ The importer calls <<fmi3GetClock>> to inquire the clock activation state.
 |`triggered`
 |[[triggered,`triggered`]]<<fmi3SetClock>> in <<EventMode>>,
 <<fmi3ActivateModelPartition>> in <<ClockActivationMode>>,
-arguments <<clocksAboutToTick>> (<<fmi3DoStep>>) and <<clocksTicked>> (<<fmi3CallbackIntermediateUpdate>>)
+argument <<clocksTicked>> of <<fmi3CallbackIntermediateUpdate>>
```
clagms, line 674: @oliverkotte did you update the c files accordingly?

clagms (changes requested):

docs/2_2_common_mechanisms.adoc
andreas-junghanns: Oliver removed the clocksAboutToTick from doStep - you can see that in the header file.
Are you referring to another change that might be missing still?

clagms: Nevermind @andreas-junghanns  and @oliverkotte . I've double checked that this is correct! Closing this.

clagms (approved):

andreas-junghanns:
@TorstenBlochwitz : Can you approve of this simplification? I know you keep pointing out that checking clock status is not expensive and all other variables have to be retrieved anyway - so this change should be consistent with this view.

@t-sommer : You introduced this return argument a while back, if I remember correctly. Are you ok with the change and the arguments here?

t-sommer (approved):

PR #1321, opened 2021-02-26
Improve xrefs for get/set clocks

andreas-junghanns:
The xrefs for SetClock and GetClock pointed to a strange place - I found a slightly better place. If you can suggest an even better one, please do!

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -314,6 +314,8 @@ The current values of the variables can be inquired with the following functions
 [source, C]
 ----
 include::../headers/fmi3FunctionTypes.h[tags=Getters]
+
+include::../headers/fmi3FunctionTypes.h[tags=GetClock]
 ----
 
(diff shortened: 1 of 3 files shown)


chrbertsch (commented):
Hm. The API subsection https://fmi-standard.org/docs/3.0-dev/#_api, where the xrefs for setClock and getClock link to, does not even mention these functions. 
Perhaps we should add explanation the useage of these functions there and link to their definition?
The subsection https://fmi-standard.org/docs/3.0-dev/#get-and-set-variable-values, where setClock and getClock are defined, does not explain it.

andreas-junghanns:
Yes, I only now thought of a way to improve the situation: I now repeat the GetClock and SetClock prototypes in the Clock-API section. I did not explain the arguments of the functions again, the link to the general get/set section has that sufficiently explained, I think. Please have a look and propose further shortcommings or improvements.

chrbertsch (approved):
Looks good!

PR #1332, opened 2021-03-03
Unify heading depth

### LINKED ISSUE #1327: Unify granularity of headings and index
While for some subsection we have very detailed  (numbered) headings and index entries (e.g. https://fmi-standard.org/docs/3.0-dev/#fmiTerminalsAndIcons: 
grafik
, for other (long and important) subsections like the definition of model variables we have almost none: 
grafik

We should unify the granularity.

chrbertsch:
resolves #1327 

Please review - yo are welcome to commit directly


andreas-junghanns:
Indeed, much better. Do we need more, or can we review and approve?

andreas-junghanns (approved):

Issue #1333, opened 2021-03-04
Description of input clocks neglects connected clocks.


clagms:
Where it says:
> Contrary to the importer being the source of the clock activation, the definition of the clock timing comes from the FMU, either through the modelDescription.xml or calling fmi3GetInterval.

It should mention the same that clocks are connected:
> Contrary to the importer being the source of the clock activation, the definition of the clock timing comes from an FMU, either through the modelDescription.xml, calling fmi3GetInterval, or from another output clock connected to the input clock.

PR #1334, opened 2021-03-04
Fix https://github.com/modelica/fmi-standard/issues/1333

clagms:
This is a small PR to fix https://github.com/modelica/fmi-standard/issues/1333


andreas-junghanns (approved):
I narrowed down your (needed) addition to the clock type that would allow such definition transision.

Explanation: Time-based input clocks can be connected on the system level - but only for certain clock types that allow adjusting clock intervals - the importer will have to enforce such dependencies.

Issue #1337, opened 2021-03-05
Unify capitalization of "Mock-up" vs. "Mock-Up"


t-sommer:
IMHO, it should be "Mock-up" (lowercase "u") since

- "mock-up" is one word
- the capitalization should correspond to the acronym "FMU"

Can the native speakers please give their opinion on this?

andreas-junghanns:
(Warning, non-native speaker speaking)
For a long time, your reading was the consensus of the group.
However, the Logo uses -Up.
And eFMI also preferred the -Up version and after a quick check, we decided to go consistently with -Up.
I made those changes in the document for 3.0, I think.

Unless we want to open another round of (IMHO fruitless) discussion (read: both spellings are strange on their own ways), I suggest to make stick with the newly homogenised Functional Mock-Up Interface.

@TorstenBlochwitz : Please correct my historical memory, if you see fit.

t-sommer:
Uppercase "Up" is also inconsistent with FMI 1.0 and FMI 2.0. Since FMI 3.0 has not been release yet, this is the right time to fix it.

t-sommer:
> However, the Logo uses -Up.

No problem, I can fix that.

andreas-junghanns:
> > However, the Logo uses -Up.
> 
> No problem, I can fix that.

This is the least of the problems: We agreed to use the same spelling with eFMI. We are not free to just do whatever anymore.
If we want to change that (again), then we have to find another agreement with the eFMI project.

And please: if we can´t even harmonise on something as simple as naming (note: not spelling), how hard would it be to harmonise on hard technical details. Can we please leave things as currently agreed on?

chrbertsch:
Regular Design Webmeeting: 
Only "FMI" and the "FMI"-logo without text is trademarked (as descriptive names cannot be trademarked), so the trademark topic is not touched by this. 
not super-important to most of us, will be discussed and decide in the Steering Committee.

Issue #1317, opened 2021-02-23
Mention intermediate alpha version in generated html docu


chrbertsch:
Currently we have in the generatd html document that it is "alpha.5". However we have made many changes and people get confused. I suggest  that we mention, that this is an intermediate version "alpha.5 + changes" and print the generation date and/or latest git commit hash

andreas-junghanns:
@t-sommer : can you explain how this is supposed to work without us changing the text mutliple times (alpha.x -> alpha.x´ ->alpha.x+y)?

t-sommer:
The only clean solution would be to use a separate branch for development e.g. [GitFlow](https://nvie.com/posts/a-successful-git-branching-model/).

t-sommer:
We could make that transition any time, e.g. when tag `v3.0-alpha.6`.

andreas-junghanns:
but the result would be even more moving parts for the development people... to simply clarify when a release is finished.
Is the extra parts worth that? I vote no, simply because the machine already scares contributors off. Adding complexity will not help that more important issue of getting more people to get involved.

chrbertsch:
Either we go for a very leightweight change, just changing 
![grafik](https://user-images.githubusercontent.com/16363372/108966038-fc406400-767d-11eb-9956-853dc3c9c6f8.png)
to "FMI 3.0alpha.5 + changes, generation date DD-MM--YYYY hh:mm:ss" or something like this (how could one include the generation date with an adoc command?)

Or we change to somthing like gitflow (which I would prefer): what would be the consequences?
E.g., do tickets automatically get closed when the are resolved by a merged PR to develop or only when later  merged from develop to master? @t-sommer : e.g. in the fmpy development tickets get closed when taken care of in develop, but there is no link to the commit

chrbertsch:
Thinking about it, I agree with @t-sommer that we should follow the GitFlow approach: 
merge PRs into develop, prepare release branches and use master only for tagged (pre)releases (e.g., alpha.6)
Only few people create PRs and should be able to adopt the small change of merging into develop instead of master, and now that we go towards a release and get more stable, we get more consistent with the branches and can benefit from the Gitflow tooling implemented in git clients.

We could set the develop branch to be the Github default branch, so that we still could use the feature of automaticilly closing linked tickets when merging a PR into develop, see https://stackoverflow.com/questions/14858075/set-the-develop-branch-as-the-default-for-a-pull-request

t-sommer:
[...]

PR #1338, opened 2021-03-05
Update docs for new GitFlow branching scheme

t-sommer:


chrbertsch:
Looking at the picture:
is 3.x really a support branch?  Should be 3.x.y in my view

chrbertsch (approved):

andreas-junghanns (approved):

pmai:
Did I miss the rationale for switching to gitflow again?

chrbertsch:
> 
> 
> Did I miss the rationale for switching to gitflow again?

See discussion in https://github.com/modelica/fmi-standard/issues/1317 and Steering Committe meeting: https://github.com/modelica/fmi-design/tree/master/Meetings/2021/2021-03-02-FMI-Steering#2-using-gitflow-branching-scheme

pmai:
Sorry that I missed this, but #1337 (and the steering meeting minutes) does not provide any rationale for switching development workflows, because the auto-publishing problem could just as easily have been fixed by not using master/main for the auto-publishing, which would only affect the CI scripts and release publishing, whereas the new development workflow now affects all developers (including down-stream developers) just to fix, it seems, a simple CI issue. This seems not good to put it mildly, especially in between releases. Probably has sailed, but let's maybe not have impactful process decisions to fix simple technical problems at very short notice in the future?

chrbertsch:
@pmai : What are your arguments against using the Gitflow branching scheme?
Can we discuss this in the regular webmeeting on Wednesday?
There are (unfortuneately) few people creating and working with PRs, and even fewer digging as deep into the git subtleties like @t-sommer  and you, so we should come to an acceptable solution for them

t-sommer:
I guess the whole discussion boils down to two questions that caused confusion in the past:

- should the development version be "customer facing"?
- should we do the release engineering on the main development branch (currently `master`)?

If the commit / tag added to the HTML in 369ec990b08392bd4773fef9080bfe97d363df99 solves this problem we can close this PR.

chrbertsch:
> should the development version be "customer facing"?

We should then link on https://fmi-standard.org/  and https://github.com/modelica/fmi-standard link not only the latest development version but more prominently the latest tagged version


> should we do the release engineering on the main development branch (currently master)?

Even with using Gitflow, we could the release preperation on an extra branch, right?

t-sommer:
> Even with using Gitflow, we could the release preperation on an extra branch, right

Either way all contributions should be made via PRs.

PR #1342, opened 2021-03-08
intermediateUpdateTime in SE

### LINKED ISSUE #1308: Intermediate update time in SE = modelPartition Activation time?
According to https://fmi-standard.org/docs/3.0-dev/#advancing-time, "time in SE has a more discrete form". 
To my understanding, the activation of a certain modelPartition corresponds to a discrete point of time. 
And at least for non-realtime simulation, during the evaluation of the model partition it stays at this simulation time.

Now what happens, if this model partition invokes fmi3CallbackIntermediateUpdate? Which time can the argument intermediateUpdateTime have? must it be the time of the modelPartition activation? Or can it be a later point of time?
If time is discrete for a modelPartitionActivation to my understanding the intermediateUpdateTime must be equal to the model activation time and cannot be a later point of time. Is this correct? Then we should explicitly state this

MBlesken:
As discussed concerning in last design meeting the intermediateUpdateTime is not relevant in SE. @andreas-junghanns, @chrbertsch please have a look. FYI: @IZacharias Closes #1308


andreas-junghanns (approved):

chrbertsch:
The "can be ignored" is quite vague. What is its meaning, if it is not ignored?

MBlesken:
> The "can be ignored" is quite vague. What is its meaning, if it is not ignored?
@chrbertsch so "should be ignored since it has no meaning in SE"...?

MBlesken:
> The "can be ignored" is quite vague. What is its meaning, if it is not ignored?

@chrbertsch I tried to be more specific in my adaptions and will commit close the PR now. Anyway, I will regard further comments or questions if you have any. Thank you for your feedback.

PR #1345, opened 2021-03-11
Set revdate and revnumber in CI

### LINKED ISSUE #1317: Mention intermediate alpha version in generated html docu

t-sommer:
fixes #1317

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,9 @@ jobs:
             git submodule update --init --recursive
       - run:
           name: Generate HTML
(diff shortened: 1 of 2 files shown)


chrbertsch (approved):
Looks good. 
We would use this for the development branch, and the tag name on main, right?

chrbertsch:
Should we mention, that the generated intermediate version is based on alpha.6 this this case?

t-sommer:
I've put the new branching scheme on hold because of @pmai's concerns. There is no pressing need to introduce it and this PR works either way.

PR #1353, opened 2021-03-16
Use bool from stdbool.h instead of _Bool

### LINKED ISSUE #1350: _Bool is not defined
Hi

I tried to compile my FMUs with alpha6. I was not able to compile with g++ on ubuntu (version 9.3) because _Bool is not defined.
It is used for fmi3Boolean and fmi3Clock in fmi3PlatformTypes.h

t-sommer:
fixes #1350

### DIFF
--- a/headers/fmi3PlatformTypes.h
+++ b/headers/fmi3PlatformTypes.h
@@ -81,12 +81,12 @@ typedef            bool fmi3Clock;    /* Data type to be used with fmi3ClockActi
                                          fmi3ClockInactive */
 
 /* Values for fmi3Boolean */
-#define fmi3True  1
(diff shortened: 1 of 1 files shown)


andreas-junghanns (approved):
I am sure this is much better.
I still do not like that clocks are also boolean - just saying, does not belong here.

pmai (approved):
Probably simplest approach, even if it pollutes the global namespace some more.

andreas-junghanns:
@t-sommer : can we still treat this as Alpha.6?

t-sommer:
v3.0-alpha.6 has already been tagged, so I would say no. However existing implementations should be API compatible.

PR #1354, opened 2021-03-16
Move and improve data flow images

andreas-junghanns:


chrbertsch:
- The clocks are missing in the pictures especially in the SE picture (we should liste them in the variables and name the red arrows)
- Do we really ned the solver in the tasks of the SE picture?
- Should we not use the term "model partition" instead of "tasks" in the SE picture?
- Could we visualize in the SE picture, that some outputs are clocked, i.e. come out of the model partitions?

andreas-junghanns:
We tried to keep the images simple: if we add too many details, the important differences are getting swamped.
E.g. clocks are not a difference between the different interface types.
The solver is a detail TorstenB wanted, maybe we could drop one.
Model partitions are not specific to SE, tasks are.

chrbertsch (approved):
Looks good.

PR #1355, opened 2021-03-17
Unify naming of "continuous states" and "derivatives" in API and XML

### LINKED ISSUE #1343: Unify naming of "continues states" and "derivatives" in API, XML, and docs
Currently we have 

- `<StateDerivative>` in the Model Structure
- `fmi3GetContinuousStates()` and `fmi3GetDerivatives()` in the API
- a mix of "continuous-time states", "states" and "continuous states" in the docs

I suggest to call them "continuous state" and "continuous state derivative" to clearly distinguish them from state diagram states, discrete states and output derivatives.

see also #1336

t-sommer:
to clearly distinguish the different types of derivatives in 

- fmi3GetContinuousStateDerivatives()
- fmi3GetOutputDerivatives()
- fmi3GetDirectionalDerivative()
- fmi3GetAdjointDerivative()

and

<ContinuousStateDerivative />

fixes #1343

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -1040,12 +1040,12 @@ where
 
 ** <<InitializationMode>>: The exposed unknowns listed as elements <<InitialUnknown>> in `<ModelStructure>` that have a floating point type.
 
-** <<ContinuousTimeMode>> (Model Exchange): The continuous-time outputs and state derivatives (= the variables listed as elements <<Output>> of `<ModelStructure>` with a floating point type and <<variability>> = <<continuous>> and the variables listed as elements <<StateDerivative>> in `<ModelStructure>`).
+** <<ContinuousTimeMode>> (Model Exchange): The continuous-time outputs and state derivatives (= the variables listed as elements <<Output>> of `<ModelStructure>` with a floating point type and <<variability>> = <<continuous>> and the variables listed as elements <<ContinuousStateDerivative>> in `<ModelStructure>`).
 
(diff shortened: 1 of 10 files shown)


andreas-junghanns (changes requested):
If you want to be correct, you have to call "states" "continuous-time states" and their derivatives "continuous-time state derivatives".
So your suggestion is now a compromise that uses longer names and still is not correct.
You even removed at least one "-time" where it makes perfect sense.

If we want to keep it short (and I would prefer short) we need to ask: is there a way to reasonably misunderstand what is ment. I see little danger of that.

t-sommer:
> If you want to be correct, you have to call "states" "continuous-time states" and their derivatives "continuous-time state derivatives".

The correct term would be "continuous w.r.t. the independent variable". But is continuous ever referring to something else in the document? If not we could just state that and use "continuous".

> So your suggestion is now a compromise that uses longer names and still is not correct.

See above.

> You even removed at least one "-time" where it makes perfect sense.

By accident. Fixed it.

andreas-junghanns:
We do have discrete states. Well, discrete w.r.t. value and discrete w.r.t. time.
"Continuous states" could mean a discrete-time state with a continuous value (float/double).
To be clear, we tried to consistently name them "continuous-time states".

t-sommer:
> "Continuous states" could mean a discrete-time state with a continuous value (float/double).

The question is: do we currently or will we have to refer to these in the docs?

andreas-junghanns:
> 
> 
> > "Continuous states" could mean a discrete-time state with a continuous value (float/double).
> 
> The question is: do we currently or will we ever have to refer to these in the docs?

We refer to discrete states, yes - there it is always clear what we mean.

andreas-junghanns:
WEB design meeting: agreed with Torsten´s proposal.

PR #1361, opened 2021-03-19
Clean up Clocks API and schema

t-sommer:
TODOs:

- [x] change "Clock with causality == input" to "input Clock"
- [x] remove excessive references to Clock within same chapter
- [x] write "Clock" w/o reference inside "Clocks" section
- [x] state that Clocks must have causality == input or output

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -341,7 +341,7 @@ typedef fmi3Status fmi3SetIntervalDecimalTYPE(fmi3Instance instance,
                                               const fmi3Float64 interval[]);
 ----
 
-A clock interval is set by the environment for the current time instant by the function <<fmi3SetIntervalDecimal>>.
+A Clock interval is set by the environment for the current time instant by the function <<fmi3SetIntervalDecimal>>.
 ....
 
 === Spelling and Capitalization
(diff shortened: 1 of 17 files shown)


andreas-junghanns:
Observations:
 - you introduced argument descriptions with a new format "nValues::" - before I tried to be consistent with the "* nValues:" format. We should be consistent throughout the document.
 - achors like [[qualifier,`qualifier`]] seem a bit off (too low in that case).
 - I tried to not repeat descriptions of argument used over and over again, like `valueReferences`. In your case, if those valueReferences must point to specific types of variables, stating that is probably a good idea. But we should check the rest of the document for consistency in that respect.
 - sometimes you merge argument descriptions (e.g. `intervalCounters` and `resolutions`), sometimes you have them separate (e.g. `intervals` and `qualifiers`). I did not detect a pattern...
 - search for <<input>> <<clock>> to turn to <<inputClock>>
 - in the clocks-type table, we format `aperiodic` and `periodic` - but these are not keywords, right? Drop the ``?

Let me know and for most of these, I can add the changes to this PR.

andreas-junghanns (changes requested):
See my previous comment...

andreas-junghanns (approved):

TorstenBlochwitz (approved):
Great work! Makes Clocks more to variables than before!

PR #1366, opened 2021-03-22
Recommend file name and static symbols for source code FMUs

t-sommer:
This will avoid duplicate object names and symbols when compiling and linking multiple source code FMUs into the same binary, e.g. for a HiL application.

### DIFF
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -95,8 +95,10 @@ Each <<BuildConfiguration>> provides the necessary information to compile and li
 An FMU importing tool may not regard more than one <<BuildConfiguration>> when building the FMU for a specific <<platform>>.
 The importer chooses the matching <<BuildConfiguration>> based on the <<platform>> and <<modelIdentifier>> attributes.
 
-If only a single <<SourceFile>> is provided in the <<BuildConfiguration>> it should have a unique name (e.g. `<model_identifier>.c`) and all symbols except for the FMI functions should be defined as `static` to avoid name conflicts when compiling and linking multiple source code FMUs into one binary.
+In order to avoid symbol name conflicts when compiling and linking multiple source FMUs, source files should keep the exported symbols to a minimum by declaring all symbols not needed for linking as `static`.
+If only a single <<SourceFile>> is provided in the <<BuildConfiguration>> all symbols except for the FMI functions should be defined as `static`.
 _[This source file may include other source files that are not listed in the <<BuildConfiguration>>.]_
+It is also recommended to use a descriptive name (e.g. `<model_identifier>.c`) for this single <<SourceFile>> instead of generic names (like `all.c`, or `model.c`) in order to aid readability and reduce integration effort.
 
 .BuildConfiguration attribute details.
 [[table-BuildConfiguration-details]]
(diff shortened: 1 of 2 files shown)


docs/2_5_fmu_distribution.adoc
```diff
@@ -95,6 +95,9 @@ Each <<BuildConfiguration>> provides the necessary information to compile and li
 An FMU importing tool may not regard more than one <<BuildConfiguration>> when building the FMU for a specific <<platform>>.
 The importer chooses the matching <<BuildConfiguration>> based on the <<platform>> and <<modelIdentifier>> attributes.
 
+If only a single <<SourceFile>> is provided in the <<BuildConfiguration>> it should have a unique name (e.g. `<model_identifier>.c`) and all symbols except for the FMI functions should be defined as `static` to avoid name conflicts when importing multiple source code FMUs.
```
andreas-junghanns, line 98: Should "importing" not be changed to "linking"?

andreas-junghanns (approved):

docs/2_5_fmu_distribution.adoc
```diff
@@ -95,6 +95,9 @@ Each <<BuildConfiguration>> provides the necessary information to compile and li
 An FMU importing tool may not regard more than one <<BuildConfiguration>> when building the FMU for a specific <<platform>>.
 The importer chooses the matching <<BuildConfiguration>> based on the <<platform>> and <<modelIdentifier>> attributes.
 
+If only a single <<SourceFile>> is provided in the <<BuildConfiguration>> it should have a unique name (e.g. `<model_identifier>.c`) and all symbols except for the FMI functions should be defined as `static` to avoid name conflicts when compiling and linking multiple source code FMUs into one binary.
```
beutlich, line 98: Can you clarify what `unique name` refers to? How can an exporting tool ensure that its only source file name is unique w.r.t. to the importing environment?

t-sommer: Maybe we should better call it „non-generic“ (like model.c or all.c)

beutlich: Thus, it is not about avoiding name conflicts, but just to minimize the danger of them, right?

andreas-junghanns: We could not ensure this - source files are customer files we bundle and we cannot even wrap them nicely (using #include) as suggested without getting into name-conflict trouble. We should rather suggest to ensure each is compiled in their own subdirectory to avoid file-name clashes.
Note: Even then, merging sources from different source-code FMUs may create conflicts: global variables and function names, even types can have the same name but point to unique objects.
We cannot solve this perfectly, but we can improve the likelihood of success with some suggestions (again, "rules" are too strong, packaging customer code into FMUs...).

t-sommer: > Thus, it is not about avoiding name conflicts, but just to minimize the danger of them, right?

Yes. It's basically applying the function prefix concept from the headers (`<model_identifier>_`) to the filename (and hence the object file).

chrbertsch:
Regular Design Meeting: 

Pierre: The file name issue is something different than the identifier and can be handled differently
Torsten: e.g. in Simulink everything is in one directory
Pierre: for the file names we have to handle this anyway.
Torsten: Generic names as all.c are not a good idea, it is just a recommendation
Pierre: you have to handle it anyway for the file names. You must rename anyway. For symbols it is a different issue.
Torsten: We would have the same issue with the dll. 
Pierre: we should do so.
Torsten: we should make it consistent and use the model identifier. Perhaps the static symbols should be different ticket
Pierre: for the dll we are enforcing a name, for the c file not. So we would still not be consistent
Torsten: this would be very strong. And we we could not backport it to 2.0. And we have the build configuration, with something like a ..._arm.c file
Pierre: We do not have  fine-granular dll configuration. We define the entry point by the name convention, then this dll could lode others.
....
Torsten: my recommendation only works for the one sourcefile solution.


@pmai : I will propose a variant on the symbols

pmai (approved):
Obviously I now like my own wording ;)

PR #1368, opened 2021-03-23
update SE example: snippets and xml

MBlesken:


MBlesken:
@andreas-junghanns @t-sommer @chrbertsch @clagms as well as everyone else @IZacharias and would like to invite you to review and maybe contribute to this PR. Thank you.

Note that the ModelStructure contains a reference between and input clock and the output clock it triggers. We would like to raise the question if and how dependencies between input variables, parameters ... to clocks should be defined. Probably we also need to know the relation between an input clock and a countdown clock. @IZacharias will create a separate issue but this PR may already initiate creating ideas and opinions.

andreas-junghanns:
> Note that the ModelStructure contains a reference between and input clock and the output clock it triggers. We would like to raise the question if and how dependencies between input variables, parameters ... to clocks should be defined. Probably we also need to know the relation between an input clock and a countdown clock. @IZacharias will create a separate issue but this PR may already initiate creating ideas and opinions.

The <Clock> element of <ModelStructure> lists all variables that are clocked by this clock - all referenced by their valueReferences. We did not restrict what kinds of variables - so outputClocks are not excluded and could be listed - right? Same for inputs and parameters... I asked the element <Clock> to be added to <ModelStructure> because when working on the layered standard networks4FMI (see #1177) I needed to express clocked inputs and that was not possible otherwise.
Am I missing the point of your question? (I am unsure because your example already has <Clock> elements.)

andreas-junghanns (changes requested):
I fixed a formatting issue (clock to Clock).

The format of the C code is different to the other examples (see the placement of the `{`).
Should we try to follow the same formatting rules across all examples?

MBlesken:
@andreas-junghanns I fixed the formating and added a few entries to the ModelStructure after studying the respective chapter in the spec. The output clock is now listes twice: As Clock and as Output. Does the whole ModelStructure in this example match your expectation? @IZacharias FYI

t-sommer:
Currently we use backward definitions of dependencies (for derivatives, outputs, initial unknowns) i.e. we define which variables it depends on. What is the rationale to define it forward for Clocks? If there can only be one associated Clock for a variable, we could use an attribute `clock="vr"` like we do for derivatives.

chrbertsch:
> If there can only be one associated Clock for a variable, we could use an attribute clock="vr" like we do for derivatives.

From https://fmi-standard.org/docs/3.0-dev/#ModelStructure: 
"A clocked variable my depend on multiple clocks and may therefore be listed in multiple <Clock> elements. _[More rigorous importers requiring a variable to be dependent on a single clock can reject FMUs violating this restriction.]_"

> Currently we use backward definitions of dependencies (for derivatives, outputs, initial unknowns) i.e. we define which variables it depends on. 

Wouldn't it be possible to merge the "ClockedState" and "Clock" elements of the model-structure to a new ClockedVariables elements, where all Clocks a variable depends on are listed, and being a clocked "state" (and having the "previous" value) is just an attribute?

t-sommer (approved):
I think this is an improvement over the current situation. We should merge this PR and continue the discussion in a separate issue.

chrbertsch (approved):

andreas-junghanns (approved):
Yes, no more obvious issues I can see - thank you for your efforts!

PR #1372, opened 2021-03-25
Improve Terminals and Icons section

andreas-junghanns:


andreas-junghanns:
@clagms , @KarlWernersson : Please comment. If no objections are raised in the next 24 hours, I will merge this in.

clagms (approved):
I'm not very familiar with the terminals and icons stuff, but the text looks fine.

PR #1375, opened 2021-03-25
Move Clock dependency info to attribute

### LINKED ISSUE #1370: clock dependependency definition in model structure
(Singled out from #1368, referring to #1287 and  #1328 )

TorstenS: 
Currently we use backward definitions of dependencies (for derivatives, outputs, initial unknowns) i.e. we define which variables it depends on. What is the rationale to define it forward for Clocks? If there can only be one associated Clock for a variable, we could use an attribute clock="vr" like we do for derivatives.

Christian:
From https://fmi-standard.org/docs/3.0-dev/#ModelStructure:
"A clocked variable my depend on multiple clocks and may therefore be listed in multiple elements. _[More rigorous importers requiring a variable to be dependent on a single clock can reject FMUs violating this restriction.]_"

Wouldn't it be possible to merge the "ClockedState" and "Clock" elements of the model-structure to a new ClockedVariables elements, where all Clocks a variable depends on are listed, and being a clocked "state" (and having the "previous" value) is just an attribute?

t-sommer:
fixes #1370


andreas-junghanns:
@t-sommer : Is this ready for review? If not, can I help?

@TorstenBlochwitz : Can you have a look, please and comment? Do you remember why we originally chose to have a ModelStructure element and not add this as attribute to the Variable element as now proposed?

KarlWernersson (approved):

KarlWernersson:
@andreas-junghanns 
As far as I remember, the discussion was that both ways of of representing the information was valid and gave the same information. It ended up being a matter of preference on how to represent it. A vote was cast and it ended 2:1, it the small working group.

t-sommer:
The schema figures still need to be regenerated.

andreas-junghanns:
> @andreas-junghanns
> As far as I remember, the discussion was that both ways of of representing the information was valid and gave the same information. It ended up being a matter of preference on how to represent it. A vote was cast and it ended 2:1, it the small working group.

@t-sommer convinced me that this proposal might be a better way for these reasons:
 - shorter lists are more human readable
 - information is now more naturally aligned with other "Variable" information
 - ModelStructure uses the other dependency direction for (most) other elements

@TorstenBlochwitz : Any strong arguments against?

andreas-junghanns:
> 
> 
> The schema figures still need to be regenerated.

Done.

andreas-junghanns:
Ready for review?

IZacharias:
I like the idea of putting the clock dependency back to the variables.

But why are we limiting this to discrete variables?
Especially in Scheduled Execution, we need the relationships to clocks not only for inputs and outputs, but also for parameters and local variables. Only in this way the master is able to generate optimal code. The master needs the information to decide on where to place the calls to the fmi3GetXXX() and fmi3SetXXX() functions.
Example: If a parameter is only needed for the calculation of exactly one model partition, updating its value, i.e. calling the fmi3SetXXX() function, is only necessary and meaningful before calling the fmi3ActivateModelPartition() function for the corresponding clock.

KarlWernersson:
@IZacharias 

> But why are we limiting this to discrete variables?

Only discrete variables will change at a clock tick. 
If it is continuous it cannot belong to the clock.

For fixed tunable and constant. These are set or retrieved independent of clock ticks. 
You would only call set/get at initializatoinMode (possible event mode if you change tunable). 
So is it really needed for optimizing that? Initialization is still done globally and not on specific clock partion.

But it could still have some use for model information to assign parameters to specific clock partitions, so as long as we forbid continuous I am ok to allow all others to be part of a clock.

IZacharias:
> @IZacharias
> 
> > But why are we limiting this to discrete variables?
> 
> Only discrete variables will change at a clock tick.
> If it is continuous it cannot belong to the clock.
> 
> For fixed tunable and constant. These are set or retrieved independent of clock ticks.
> You would only call set/get at initializatoinMode (possible event mode if you change tunable).
> So is it really needed for optimizing that? Initialization is still done globally and not on specific clock partion.
> 
> But it could still have some use for model information to assign parameters to specific clock partitions, so as long as we forbid continuous I am ok to allow all others to be part of a clock.

@KarlWernersson You're right, I probably didn't make that clear enough. My point is to allow it for tunable parameters, discrete inputs, discrete outputs, and discrete local variables. For fixed, constant, continuous it is not needed or makes no sense.

andreas-junghanns (approved):

docs/2_4_common_schema.adoc
```diff
@@ -1077,12 +1077,19 @@ The default value of this attribute is `false`.
...
 
 _[For example, if `previous == 3` for variable `8`, then variable `3` is the previous value of variable `8`. See also <<fmi3UpdateDiscreteStates>>._
 _Note: This is reverse compared to the <<derivative>> attribute.]_
 
+|`clocks`
+|
+[[clocks,`clocks`]]
+If present, this variable is clocked.
+The value of <<clocks>> is a list of value references of <<Clock>> variables this variable depends on.
+Only discrete variables can have this attribute.
```
IZacharias, line 1091: > Only discrete variables can have this attribute.

We need the information also for tunable parameters, i.e. discrete variables and tunable parameters can have this attribute.

Issue #1380, opened 2021-03-29
External Scheduling using FMU 3.0 for Co-Simulation with Event Mode


oliverkotte:
We have a concept for using Co-Simulation FMU 3.0s with Event Mode to implement virtual ECUs with external scheduling. One FMU contains the external scheduler, another FMU the scheduler-less functional virtual ECU.

The trick is for the virtual ECU to not have any computations done in StepMode, i.e. doStep() only advances internal simulation time. All computations are done in Event Mode, i.e. in updateDiscreteStates(). The second trick is that timing information is conveyed from the external scheduler FMU to the functional FMU by way of clock activations.

I am of the opinion that some text passages of the current FMI spec should probably be adjusted. Most prominently, Table 1 contains a row stating that CS has "Scheduler Included" and SE has not. But this notion is I believe wrong at least for virtual time single-threaded ECUs as FMU, I thus suggest to delete the row. There are further sections in the spec that carry this notion, at least those below, maybe they should be changed as well?

Beyond this, at least for virtual time single-threaded use cases, with New Clocks and model partitions (representing e.g. ECU tasks) unified across SE and CS, and with StepMode doing nothing but advancing internal simulation time, it appears to me that
`activateModelPartition (clock, activationTime)` of SE 
reduces to being just a convenience function for shortening the sequence
`doStep(dt to activationTime), enterEventMode(), setClock (clock), updateDiscreteStates(), enterStepMode()` of CS
However, using Co-Simulation FMUs instead of Scheduled Execution FMUs allows us to more easily co-simulate with the FMUs of the remaining system, and to move the scheduler itself into an FMU, so implementing external scheduling with CS is more beneficial to our use case.
I wonder how far this CS-approach can be taken to cover use cases of the "SE domain". 

--- Further text passages with the notion that virtual ECUs as SE but not as CS may have external schedulers ---

in Appendix A. SE is defined as
FMI type that externalizes the scheduler to run model partitions, potentially synchronized between more than one FMU and exchanging input and output variables accordingly.
-> Can be done with CS as well at least in some cases.

1.1
addition of the new interface type FMI for Scheduled Execution (see Section 5) that allows activation of individual model partitions (or tasks) from an external scheduler.
-> SE does not enable it (for virtual time single-threaded sim), it can be done with CS as well at least in some cases.

4. 
As another example, FMUs representing controller code, an internal scheduling algorithm will trigger tasks at the correct time and order while advancing time to the next communication point
[...]

PR #1384, opened 2021-03-31
wording to not seemingly forbid external scheduling with CS

oliverkotte:
refers to issue #1380


andreas-junghanns (approved):
I made a few changes, I think this is ok now

PR #1386, opened 2021-04-01
Consistency in variables for SE (assignment to clocks and preemption)

MBlesken:
First draft.

Thinking about a graphic ....


MBlesken:
@IZacharias @andreas-junghanns @clagms @chrbertsch I would like to invite you to review this PR.

I'm open to suggestions if e.g. a graphic (what kind) could be helpful to understand the examples I'm listing.

@andreas-junghanns I'm not sure about formating. I think the comments on Set/get operations should be level with the left side of the page so I deleted a few "+". Can you please check?

Thank you.

docs/5_2_scheduled_execution_api.adoc
```diff
@@ -68,22 +68,26 @@ This function can be called before scheduling a model partition for variables as
...
 
-+
 The restrictions related to variable <<causality>> and <<variability>> defined for <<StepMode>> in <<fmi-for-co-simulation,`Co-Simulation`>> apply.
 
-+
 It is not allowed to call <<get-and-set-variable-values,`fmi3Get{VariableTypeExclClock}`>> functions after <<get-and-set-variable-values,`fmi3Set{VariableTypeExclClock}`>> functions without an <<fmi3ActivateModelPartition>> call in between.
 _[The reason is to avoid different interpretations of the caching, since contrary to <<fmi-for-model-exchange, `FMI for Model Exchange`>>, <<fmi3ActivateModelPartition>> will perform the actual calculation instead of <<get-and-set-variable-values,`fmi3Get{VariableTypeExclClock}`>>, and therefore, dummy algebraic loops at communication points cannot be handled by an appropriate sequence of <<get-and-set-variable-values,`fmi3Get{VariableTypeExclClock}`>> and <<get-and-set-variable-values,`fmi3Set{VariableTypeExclClock}`>> calls as for Model Exchange.]_
 
+To ensure predictable behavior every variable should be assigned uniquely to a model partition via its associated <<Clock>>.
+Observe two examples:
```
chrbertsch, line 82: Shouldn't we format the examples as non-normative? 
Just as _[Examples: ...

docs/examples/snippets.c
```diff
@@ -213,9 +213,11 @@ fmi3Status fmi3GetIntervalDecimal(fmi3Instance instance, const fmi3ValueReferenc
 
 	for (int i = 0; i < nValues; i++) {
 		if (valueReferences[i] == 6) {
+            env->lockPreemption(); // Optional: Preventing preemption is actually not needed here.
```
chrbertsch, line 216: Could we give an exmaple where preemption is needed?

docs/5_2_scheduled_execution_api.adoc
```diff
@@ -68,22 +68,26 @@ This function can be called before scheduling a model partition for variables as
 Function <<get-and-set-variable-values,`fmi3Get{VariableTypeExclClock}`>>::
 These functions can be called after the computation of a model partition for variables assigned to that model partition via its associated <<Clock>> and all variables not associated to a <<Clock>> (global variables).
 
-+
-Set/get operations must be atomic for a single variable.
```
chrbertsch, line 72: I learned in the discussion that "atomic" is the right term for the intended behaviour. When we delete it here, shouldn't we used it somewhere else in the text?

docs/5_2_scheduled_execution_api.adoc
```diff
@@ -68,22 +68,26 @@ This function can be called before scheduling a model partition for variables as
...
 
+To ensure predictable behavior every variable should be assigned uniquely to a model partition via its associated <<Clock>>.
+Observe two examples:
+
+* An output variable is written by two different model partitions.
+The values returned by <<get-and-set-variable-values,`fmi3Get`>> that is placed after <<fmi3ActivateModelPartition>> may return the values computed by the other model partition if that model partition has higher priority and preempted the task execution meanwhile.
+* An input variable is read by two different model partitions.
+The variable values of a model partitions of lower priority may have been overwritten by a call to <<get-and-set-variable-values,`fmi3Set`>> in the context of a task executed for a model partition of higher priority.
+
+Note that if every variable is assigned uniquely to a model partition it is also not necessary to use <<preemption-support, `preemption locks`>> in the context of <<get-and-set-variable-values,`fmi3Get`>> and <<get-and-set-variable-values,`fmi3set`>>.
```
chrbertsch, line 89: Non-normative?

t-sommer (changes requested):
"Clock" should be upper case if it refers to Clocks in FMI (as opposed to a wall clock e.g.).

docs/5_1_scheduled_execution_math.adoc
```diff
@@ -19,10 +19,11 @@ More details can be found in <<Clock,Clocks>>.
...
-That also requires that the FMU reports the active state of an <<outputClock>> only with the first call of <<fmi3GetClock>> for a specific activation of this <<Clock>> and sets the reported activation state immediately back to `false` for the following <<fmi3GetClock>> calls for that <<Clock>> until this <<outputClock>> is internally activated again.
+The FMU's code has to be prepared for being able to correctly handle preemptive calls of <<fmi3ActivateModelPartition>>, <<get-and-set-variable-values,`fmi3Get{VariableType}`>>, <<get-and-set-variable-values,`fmi3Set{VariableType}`>>, <<fmi3GetClock>> and <<fmi3GetIntervalDecimal>>.
+In general for Scheduled Execution this requires a secured internal and external access to global states and variable values to ensure the correct handling of the preemption of model partition computations.
+For <<get-and-set-variable-values,`fmi3Get{VariableType}`>>, <<get-and-set-variable-values,`fmi3Set{VariableType}`>> in particular a unique assignment of the respective variables to model partitions via its associated <<Clock>> is strongly recommended.
```
clagms, line 24: Should this be non-normative text?

docs/5_2_scheduled_execution_api.adoc
clagms: Also, the previous mention of the unique assignment was made as a strong recommendation.
So the text "To ensure predictable..." and subsequent examples should be non-normative, I think.

docs/5_3_scheduled_execution_example.adoc
```diff
@@ -57,14 +57,18 @@ In the context of the task being executed every 10 ms, the FMU initiates the sch
 include::examples/snippets.c[tags=SE_fmu_activateMP10ms]
 ----
 
-In <<fmi3GetIntervalDecimal>> the Clock's interval qualifier is reset to <<fmi3GetIntervalDecimal, `fmi3IntervalUnchanged`>>.
+In <<fmi3GetIntervalDecimal>> the clock's interval qualifier is reset to <<fmi3GetIntervalDecimal, `fmi3IntervalUnchanged`>>.
+To ensure consistency it may be necessary to prevent preemption.
```
clagms, line 61: This is a bit vague. It's unclear whether we are forcing this, or just recommending it.

andreas-junghanns: ... a bit vague and still not clear, who must prevent preemption: importer of FMU?

docs/examples/snippets.c
andreas-junghanns: ... or clarify why we do not need preemption here (single model partition changing this variable?)...

andreas-junghanns (changes requested):
I think we still need to improve some parts of the text as indicated by comments below.

docs/examples/snippets.c
MBlesken: ... the HTML around the snippet says "In this example it is actually not needed since CallbackIntermediateUpdate is only called from one model partition." This is supposed to address your remarks. It's not in the same file so maybe that's why you missed it...

docs/5_2_scheduled_execution_api.adoc
MBlesken: @IZacharias @andreas-junghanns  ... what do you think? I wouldn't miss it

docs/5_2_scheduled_execution_api.adoc
MBlesken: @andreas-junghanns would you please decide on the formating?

MBlesken: .... here a various comments below.

docs/5_3_scheduled_execution_example.adoc
MBlesken: > "Clock" should be upper case if it refers to Clocks in FMI (as opposed to a wall clock e.g.).

@andreas-junghanns What do you think? ... the term occurs 383 times in the document.

MBlesken: > This is a bit vague

I changed the text to explicitly name the FMU.

docs/5_1_scheduled_execution_math.adoc
IZacharias: The first two bullets should not be non-normative text. An importer must rely on the fact that an FMU that provides the Scheduled Execution interface is prepared to handle preemptive calls.
The third bullet is a recommendation and can be non-normative text.

docs/5_2_scheduled_execution_api.adoc
MBlesken: I put this in _[ ... ]_

docs/5_2_scheduled_execution_api.adoc
MBlesken: Yes, I changed it.

docs/5_1_scheduled_execution_math.adoc
MBlesken: I changed the third row.

docs/5_2_scheduled_execution_api.adoc
andreas-junghanns: @IZacharias & @MBlesken : This is the last pending issue for this PR and I think this is very important to define clearly. I have made a proposal in this PR now to require the importer to ensure get/set calls are not preempted. Please fix this if this is wrong from your point of view. If you agree, we can mark this conversation as resolved and merge.

andreas-junghanns (approved):

docs/5_2_scheduled_execution_api.adoc
MBlesken: @andreas-junghanns Sorry, I needed to undo your commit "Require importer to guarantee atomicity of get/set operations". It's not the importers' responsibility. I cannot fix it just by replacing "importer" to "FMU". I am working on a new commit to guide the reader a bit more ...

IZacharias: I am afraid that people might draw the wrong conclusions from 
> Set/get operations must be atomic for a single variable

It could give the impression that lockPreemption calls are always necessary. But in fact lockPreemption should be used as rarely as possible. I would rather insert some recommendations and give additional hints. 

In fact, consistency problems can only occur for "shared" variables, i.e. for variables that are accessed from several tasks. These cases should be considered:

1.  A variable is associated to more than one clock (not recommended!). Then the importer might call the set/get operation for the same variable in more than one task. This also means that the execution of a model partition might be interrupted by a set/get for a variable that is used in this model partition.
2. A variable is associated to only one clock (one model partition), but the FMU uses the variable internally in more than one model partition. In this case, the importer calls the set operation only in one task (the task in which it also calls fmiActivateModelPartition() for the associated clock).
3. A variable is associated to only one clock (one model partition), and the FMU uses the variable in this one model partition only.

In these cases, it is up to the FMU to decide how to handle this.

- In (1.) atomicity is necessary if the FMU wants to ensure consistency.
- In (2.), if the variable is associated to the clock with the highest priority, get and set are never interrupted, at least not by a task that also accesses this variable.
But if the FMU requires consistency, the FMU's code snippets that "transfer" data from one model partition to another model partition must be "smart". Note that this might also affect “internal” variables with no entry in the modelDescription.xml. Consistency can be assured, for example, by using lockPreemption/unlockPreemption (see example in 5.1.1 Preemption Support), but other ways are also possible. 
- In (2.), if the variable is associated to the clock with a lower priority, get and set can be interrupted. Only the FMU knows if an interruption might lead to a consistency problem or not. Here again it is up to the FMU to secure its code.
- In (3.) it is no problem at all if set/get are interrupted by some other task.

andreas-junghanns: > @andreas-junghanns Sorry, I needed to undo your commit "Require importer to guarantee atomicity of get/set operations". It's not the importers' responsibility. I cannot fix it just by replacing "importer" to "FMU". I am working on a new commit to guide the reader a bit more ...

Can you explain why the FMU should be responsible?
If this is required, then the following will happen:
 - importer calls fmi3GetXXX
 - FMU calls lock_preemption
 - FMU computes return values (in SE, this is not heavy, the activation call is compute intensive)
 - FMU calls unlock_preemption
 - FMU returns

If we require the importer to ensure atomicity, the fmi3GetXXX call is simpler and more efficient because the preemption mechanisms is not required. Why can´t the importer handle the preemption functionality right away?

andreas-junghanns: @IZacharias : our posts crossed paths, sorry.
Refrasing your explanation: "Asking the FMU to lock/unlock preemption is more efficient because only in rare cases such locking is required and only the FMU knows these internals. The importer would have to be more cautious and lock more often to be sure."

But: why does an FMU need to lock during get/set operations at all? The importer is in control, no model partition can compute at this moment, right? (The issue of race conditions is not part of this discussion, FMUs must handle these for sure, otherwise preemption can never work.)

MBlesken: @andreas-junghanns the clue is that lock preemptions in general should be avoided and only the FMU exporter can know where locks are actually needed. @IZacharias just tried to explain this. Is this helping? Imagine the importer had to guess where there could be potential violations. This would end up in locks for whole task bodies of all tasks except the highest priority task. That would clearly be absurd with respect to the intended preemptability of tasks.

Now, @IZacharias explanations hopefully help to understand how an FMU has to be implemented with respect to preemptability. (I also tried to improve the understanding with this PR.) My question is: Where does the implementers guide pick up on this? And how much explanation and examples need to be in the specification? (Are there specific parts in Irina's explanations that you missed and should be included in the specification?) FYI @pmai

MBlesken:
@chrbertsch @clagms @t-sommer I made quiet a few changes and there are also some discussions in this item that hopefully help make things clearer about preemption. Is it possible for you to finish your review now?

@andreas-junghanns would you like to re-review?

andreas-junghanns (approved):
This will indeed improve the text a lot - thank you!

chrbertsch (approved):

IZacharias (approved):

PR #1392, opened 2021-04-04
glossary entry for event iteration

### LINKED ISSUE #1383: Define and explain event iteration
"event iteration" is used in the spec, but not defined. We should add and explain it at least in the glossary

How does an FMU decide, when request a new event iteration in fmi3UpdateDiscreteStates? Typically, when a discrete variable has changed? This was the impression when reading http://doc.simulationx.com/4.0/1033/Content/06_Computations/Events.htm

chrbertsch:
resolves #1383


andreas-junghanns (approved):
I have tried to fix your proposal and I would merge it, if you agree now.

PR #1397, opened 2021-04-06
Add workingDirectory parameter to fmi3Instantiate*()

### LINKED ISSUE #299: Storage location for FMU output files
**Reported by andreas&#x2e;nicolai on 7 Nov 2014 12:03 UTC**
## Problem Description
FMUs may generate a large number of result data that are not part of the FMU communication. For example, an FMU of a Building Simulation Model may be connected to another FMU with control algorithms and HVAC components. The Building Simulation Model will generate a large number of internal outputs (order 1000 to 100000). Similarly, FMUs encapsulating PDE problems may generate large fields of transient data (order of variables > 1 Million when using 2D/3D grids).

In either case, exporting all variables from the model as output variables to the simulation master **is not feasible** (... "may not be feasible" is too weak a statement in this context). 

Also, existing model implementations may have own output formats tailored to a dedicated post-processing for that specific engineering domain. These outputs would naturally be written to files from within the FMU, whenever a communication step (CoSiM) or calculation step (ModelExchange) has completed.

## Assumptions about current FMU interface design
An FMU receives via the FMI v2 interface a resource path location. This path has to be treated as **read-only**, since it may be located in a temporary directory created by the master and removed by the master upon simulation end. Therefore, the resource path **may not be used** to generate output file paths from.

The location of the FMU executable/shared library may be in a read-only location, as well.

Storing absolute file paths in the FMU reduces portability, e.g. when running FMUs in cloud computing environments, where traditional file systems or paths are not available. 

Fixing output paths in an FMU instance may lead to several FMU instances overwriting each others results, when several simulation runs (several masters) are being run concurrently. A likely scenario may be, when from a set of FMUs one FMU is modified/parametrized differently, the others are used unchanged. Then a variation/sensitivity study may be done in parallel with the same FMU. This may involve 100 variations computed in parallel. If the same FMU were to be used in each variation with a fixed output path, an FMU instance might overwrite results of an other FMU, or fail to write outputs altogether. A workaround would require 100 copies of potentially large FMU containers with individual output paths. This procedure may be restrictive and cumbersome.

## Resulting requirements regarding the FMU Interface
The FMU must receive a persistent path to store its output files. The FMU must be able to store its outputs in a manner that suits the physical problem/implementation in the FMU, e.g. creation of subdirectories must be possible.

Ideally, this output path should be provided alognside the resource path. For example, the function fmi2Instantiate() could be extended as follows.

```C
fmi2Component fmi2Instantiate(fmi2String instanceName,
  fmi2Type fmuType,
  fmi2String fmuGUID,
  fmi2String fmuResourceLocation,
  /* Provide a writable output location with access permissions for the FMU process */
  fmi2String fmuOutputLocation,
  const fmi2CallbackFunctions* functions,
  fmi2Boolean visible,
  fmi2Boolean loggingOn
);
```

## Proposal for short-term workaround

The interpretation of the resource path should be explicitely changed to **persistant location with read/write access** to the FMU process.

Master implementations must guarantee that the content of the FMU is extracted into such a dedicated directory.

If the same FMU is instantiated several times by the master, each FMU instance must receive a unique directory (ultimately resulting in duplication of resource data of the original FMU).

----

Migrated-From: https://trac.fmi-standard.org/ticket/299

t-sommer:
fixes #299


andreas-junghanns:
Should we define the working directory to be unique per instance (I think yes)?
We should also define that NULL is allowed for platforms without file system, unless needsWorkingDirectory==true

t-sommer:
> Should we define the working directory to be unique per instance (I think yes)?
> We should also define that NULL is allowed for platforms without file system, unless needsWorkingDirectory==true

That was my intention. Please change the text if it's not clear.

andreas-junghanns (approved):
I think this is ok now (as it is now optional).

pmai:
I still do not see how this is useful functionality in general, unless more processing / UI expectations are given: With the current approach you get some anonymous directory you can write stuff to. Nothing gives the impression to the importer that this path should a) be left lying around after the simulation finishes, or b) be visisble to the user, so that the user can look at stuff that is written there, or c) that there is any information on what is written their, its uses, longevity, etc.

As it is done now I don't see how this fixes the issue of #299, nor how this is better than already extant functionality in the various C APIs that gives you a temporary directory.

t-sommer:
> As it is done now I don't see how this fixes the issue of #299, nor how this is better than already extant functionality in the various C APIs that gives you a temporary directory, like e.g. mkdtemp (POSIX.1-2008).

Passing the `workingDirectory` gives the importer better control. I see three possible use cases:

- the importer can keep / pre-populate the directory if it has inside knowledge about the workings of the FMU
- it can choose a device that has better I/O performance compared to the default temp directory of the OS
- the files are easily accessible for debugging and resetting the state

pmai:
The problem remains that the expectations are not made clear in the standard, neither for the importer nor for the FMU: Am I to expect large random file writes/reads? Large sequential writes? Is this data that is supposed to be available to the user (in which case I should not delete it after the simulation run)? Am I as the importer supposed to cleanup the contents of this directory after the run? Or is the FMU? Must the directory be empty, or can I reuse it in the next run? Will this change results? Are there cases were it is wanted to leave it populated and others where it is supposed to be cleared?

In terms of your use cases, they seem to depend on a couple of things being true that I am not sure hold in general very often if at all:

- If the importer (i.e. the tool, not the user) has deeper knowledge of the internal workings of the FMU, how would it gain that, and if it has that, why do we need a standardized interface for workingdir, when a bespoke parameter would be sufficient?
- Of the 4 participants to the simulation (the user, the importer, the FMU, and the system), it seems to me the importer is the last on the list to have superior knowledge on choosing I/O devices: It lacks both the knowledge of the intented use by the FMU, the system configuration and the user intent. One would expect either the user and/or the system to have superior mechanisms and knowledge here. And there are mechanisms to impart that information on the system.
- How would the importer even know that the files in the directory are intended either for debugging or in some magic fashion for resetting of state (which means that state is supposed to be retainable, which the name working directory singularly fails to even hint at)?

Again in total it seems to me for this mechanism to work at all in any way that is distinguishable from calling mkdtemp, more knowledge about the FMUs actual workings are needed by the importer than is available through modelDescription.xml. If that is the case, I fail to see how a layered spec/bespoke parameters or some other mutual/multilateral arrangement would not work better than this approach, which leaves everything open to the imagination... Which is what @andreas-junghanns and @TorstenBlochwitz argued in #299 as far as I can tell. It still seems to me that either something more complete is needed, or this will only work between bespoke FMUs and implementors, in which case it should be a layered spec.

I mean reading the current proposal this does not give any clue that this directory is supposed to be used for output files (rather than just temporary files), which is what #299 strongly hints at as a use case.

andreas-junghanns (changes requested):
@pmai has convinced me that this is not a solution. @TorstenBlochwitz convinced me that there are better solutions already (e.g. structural parameters).

t-sommer:
> @pmai has convinced me that this is not a solution. @TorstenBlochwitz convinced me that there are better solutions already (e.g. structural parameters).

So how do you retrieve the path with the current API when you need it during instantiation?

t-sommer:
Another, more generic, way to solve the problem would be to have static parameters that can be set before calling the constructor. However, if I remember correctly, this proposal was rejected at an earlier FMI Design Meeting.

TorstenBlochwitz:
> > @pmai has convinced me that this is not a solution. @TorstenBlochwitz convinced me that there are better solutions already (e.g. structural parameters).
> 
> So how do you retrieve the path with the current API when you need it during instantiation?

Everything which uses this string parameter should be done when calling fmi3EnterInitalizationMode.

TorstenBlochwitz:
> Another, more generic, way to solve the problem would be to have static parameters that can be set before calling the constructor. However, if I remember correctly, this proposal was rejected at an earlier FMI Design Meeting.

"Static" parameters are not possible, because the FMU does not exist before fmi3InstantiateXXX is called.

t-sommer:
> "Static" parameters are not possible, because the FMU does not exist before fmi3InstantiateXXX is called.

An instance of the FMU does not need to exist to call the function (which is the same for all instances of a given FMU).
We could declare these parameters e.g. like so

```xml
<String name="licenseFile" valueReference="1" static="true" .../>
``` 

and call the setter with a `NULL` instance

```C
fmi3SetString(NULL, {1}, 1, {"C:\temp\license.lic"}, 1);
```

to set a static parameter.

pmai:
> > Everything which uses this string parameter should be done when calling fmi3EnterInitalizationMode.
> 
> This is not always possible. Take e.g. a Dymola FMU that has been exported w/o a code export license. It requires the path to a license file. Currently this path has to be provided via an environment variable which is not very elegant IMHO.

This has now morphed from working directories to solving quite different problems, so I am not sure whether this discussion should be kept here. Whether structural parameters are sufficient for solving all kinds of problems is independent of them being sufficient to handle the "working directory" problem, IMHO.

That being said, why does this need to be passed prior to instantiation? It is perfectly possible to create code that does the license check at a later stage, if that is needed/wanted.

We might want to have per-module (rather than per instance) settings for other reasons, but they tend to raise lots of problems that need to be solved (e.g. when are you allowed to set them, do we have tunable versions of these, ...).

andreas-junghanns:
May I point out that this is a new feature and we had feature freeze long time ago?
Why are we discussing this the week before we want to declare the first beta - much less still a day before we want to declare beta?

TorstenBlochwitz:
> 
> 
> May I point out that this is a new feature and we had feature freeze long time ago?
> Why are we discussing this the week before we want to declare the first beta - much less still a day before we want to declare beta?

I totally agree to that and to Pierre's last comment. The ticket was issued in 2014 and nobody cared for a long time. Normally we interprete this fact in a way that the topic is of very low interets. I propose to close this, since we are close to a release.

chrbertsch:
Regular Design meeting: Will be commented and closed by @andreas-junghanns

andreas-junghanns:
Web-design meeting: no real interest during the last 7 years allows us to close this.

PR #1401, opened 2021-04-08
Dependencies Variables and Clocks

MBlesken:
DRAFT

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -808,7 +808,7 @@ The association between <<clocked-variable,clocked variables>> and their <<Clock
 <<clocked-variable,Clocked variables>> can depend on multiple Clocks.
 _[For example, a global counter could be incremented by multiple tasks, each controlled by a different Clock.]_
 
-<<outputClock,Output Clocks>> may depend on input clocks.
+Clocks may depend on input Clocks.
 This dependency is also defined by the <<clocks>> attribute.
 _[E.g. if an <<outputClock>> depends on an <<inputClock>>, then Clock ticks of the <<inputClock>> in <<EventMode>> or <<ClockActivationMode>> may create <<outputClock>> ticks during this event handling.]_
 
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -808,6 +808,10 @@ The association between <<clocked-variable,clocked variables>> and their <<Clock
 <<clocked-variable,Clocked variables>> can depend on multiple Clocks.
 _[For example, a global counter could be incremented by multiple tasks, each controlled by a different Clock.]_
 
+Clocks may depend on input clocks.
+This dependency is also defined by the <<clocks>> attribute.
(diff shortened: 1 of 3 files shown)


andreas-junghanns:
Why do we need to allow `tunable` for clock variables?

chrbertsch:
> Why do we need to allow `tunable` for clock variables?

The proposal is to allow the `clock` attribut for `tunable` parameters, that is the can be clock**ed**.

Could this be used to signal that a tunable parameter is used only in certain model partitions?

MBlesken:
The PR is now ready for review. @andreas-junghanns @KarlWernersson @chrbertsch you and everyone else is welcome to leave comments. Thank you.

IZacharias:
> > Why do we need to allow `tunable` for clock variables?
> 
> The proposal is to allow the `clocks` attribute for `tunable` parameters, that is the can be clock**ed**.
> 
> Could this be used to signal that a tunable parameter is used only in certain model partitions?

@chrbertsch Yes, this is the intention. For Scheduled Execution, we need this information so that the importer has the possibility to place the set commands sensibly (i.e. set is only called before the call to fmiActivateModelPartition for the associated clock, not before the call to fmiActivateModelPartition  for any clock)
@andreas-junghanns FYI

KarlWernersson:
Would there be interest to also allow fixed and constant to belong to a clock partition?
In a way it is not as relevant as for tunable as if it will be set it will be done in initialization before any clock is active.
However it could be of interest that a parameter only affect a certain part ion?
Do we have a reason to forbid it?
Is there a problem to state that a fixed parameter is part of a clock?

docs/2_2_common_mechanisms.adoc
```diff
@@ -808,6 +808,10 @@ The association between <<clocked-variable,clocked variables>> and their <<Clock
 <<clocked-variable,Clocked variables>> can depend on multiple Clocks.
 _[For example, a global counter could be incremented by multiple tasks, each controlled by a different Clock.]_
 
+Clocks may depend on input clocks.
```
andreas-junghanns, line 811: Should state here, that "Output Clocks may depend on input Clocks."?

andreas-junghanns (approved):

chrbertsch (approved):

docs/2_2_common_mechanisms.adoc
MBlesken: @andreas-junghanns since countdown clocks can depend on input clocks not only output but also input clocks can depend on input clocks. (This PR states an example in the xml.) I observed options "Output Clocks and countdown Clocks may depend on input Clocks." and "Clocks ([e.g. output Clocks, countdown Clocks]) may depend on input Clocks". I wasn't happy with neither of them, so I changed it back to "Clocks may depend on input Clocks." fixing "input clocks" to "input Clocks".

MBlesken:
> Would there be interest to also allow fixed and constant to belong to a clock partition?
> In a way it is not as relevant as for tunable as if it will be set it will be done in initialization before any clock is active.
> However it could be of interest that a parameter only affect a certain part ion?
> Do we have a reason to forbid it?
> Is there a problem to state that a fixed parameter is part of a clock?

@KarlWernersson I don't have an interest in also allow fixed and constant. As you said it does not seam to be as relevant to me either.

MBlesken:
@KarlWernersson just a friendly reminder. Your review is still pending.

KarlWernersson (approved):

Issue #1393, opened 2021-04-04
May time-based clocks only be activated at first super-dense time point (t_I=0)?


chrbertsch:
Currently in the "Alllowed Function Calls" for Event Model, it is stated: "Only Time-based clocks must not be active for more than one call of fmi3UpdateDiscreteStates per Event Mode."

Is there any reason, why this should not be the first super-dense time point, i.e. integer part of super-dense time t_I=0?
Then we should explicitly state this. 

And as get-calls  and not only fmi3UpdateDiscreteStates may trigger calculations, it must be clarified that time-based clocks may for a given real-part of super-dense time not be active _after_ the first call of fmi3UpdateDiscreteStates. 

So I would re-state the cited sentence, e.g. like this: 
 "Time-based clocks may only be active for super-dense time points (t_R, t_I), with t_I=0, i.e. they must not be active after the first call of fmi3UpdateDiscreteState per Event Mode."

andreas-junghanns:
@TorstenBlochwitz: We changed that - do you remember why?

clagms:
If I remember correctly, it's up to the importer to decide when to activate the time based clocks and how to prioritize among multiple activating clocks... So it's natural not to force the time based clocks to be activated on the first super-dense time instant.

chrbertsch:
>... So it's natural not to force the time based clocks to be activated on the first super-dense time instant.

The integer-part of super-dense time corresponds to event iteration. Event iteration is necessary to determine a discrete variable's change due to another discrete change. But clocked variables do not participate in event iteration (they tick only once per real part of a super dense time intant), so it does not seem natural to me to allow an ambiguity here.  The concept of priority of model partition exists only in SE. To my understanding, in ME and CS clocks ticking at the same time instant should be activated "synchronously" at he first integer part of super dense time. 
Or could someone please describe, why it should be different?

Imagine we have a CS FMU with  two periodic time-based clocks (with a 5ms and 10ms interval), and each of them increments the same (global) counter. For events, where both clocks are active, shall the counter be incremented once or twice? If both clocks are active at the same (e.g. first) integer part of super-dense time, it will be incremented once. If the 5ms- and the 10ms-clocks are active at different integer parts of super dense-time, the counter will be activated twice.
Which behaviour do we want? 
Or do we explicitely want the importer to activate clocks at different integer parts of a super-dense time instant? But from where should the importer know in which order? (as we have no priorities)

clagms:
[...]

PR #1404, opened 2021-04-12
Clarify time-based clock activation restrictions

andreas-junghanns:


chrbertsch (commented):
Looks good in principle and clarifies the concern raised in #1393 

> The event iteration for handling discontinuities of the continuous part of the FMU should precede handling of time-based Clocks

@andreas-junghanns or @TorstenBlochwitz Could you please provide a use case where one should do it like this? 
For me still an activation at (t_R, t_I=0) is more intuitive for time based clocks. (The importer coud still do this according to this proposal.

andreas-junghanns:
> @andreas-junghanns or @TorstenBlochwitz Could you please provide a use case where one should do it like this?
> For me still an activation at (t_R, t_I=0) is more intuitive for time based clocks. (The importer coud still do this according to this proposal.

Any fixed-step solver will often force events and clock activations to fall on the same time.
Then it is natural to first deal with the event (finishing the continuous step of the simulation) before starting the clock activations.
Does that explain this sufficiently?

chrbertsch (approved):

PR #1409, opened 2021-04-13
Add and constrain derivative functions in state machines

andreas-junghanns:


andreas-junghanns:
@KarlWernersson : added as requested.

@MBlesken & @IZacharias : I have removed the use of fmi3GetAdjointDerivative and fmi3GetDirectionalDerivative from Initialization Mode for SE - is that correct? (How would one compute any derivative in SE?)

KarlWernersson (approved):
Looks good. Thanks for fixing it =)

MBlesken:
> @MBlesken & @IZacharias : I have removed the use of fmi3GetAdjointDerivative and fmi3GetDirectionalDerivative from Initialization Mode for SE - is that correct? (How would one compute any derivative in SE?)

@andreas-junghanns @IZacharias I'm not sure if this should be forbidden. On the other hand I don't see why one would use it in SE. Maybe this is enough reason to forbid it...?

docs/2_3_common_states.adoc
```diff
@@ -390,9 +390,11 @@ This function should always be called after calling function <<fmi3UpdateDiscret
 
 Function <<fmi3GetDirectionalDerivative>>::
 See <<fmi3GetDirectionalDerivative>>.
+This call is not allowed in Scheduled Execution.
```
MBlesken, line 393: @andreas-junghanns @IZacharias I'm not sure if this should be forbidden. On the other hand I don't see why one would use it in SE. Maybe this is enough reason to forbid it...?

docs/2_3_common_states.adoc
```diff
@@ -390,9 +390,11 @@ This function should always be called after calling function <<fmi3UpdateDiscret
...
 
 Function <<fmi3GetAdjointDerivative>>::
 See <<fmi3GetAdjointDerivative>>.
+This call is not allowed in Scheduled Execution.
```
MBlesken, line 397: ... same here

PR #1414, opened 2021-04-16
Rename Transmission Line Method to Transmission Line Modeling

robbr48:
Renamed "Transmission Line Method" to "Transmission Line Modeling", which is the most commonly used name for the method.

Also added anchors and a reference.


chrbertsch (approved):

andreas-junghanns (changes requested):
The title of the paper spells Modelling with double-ll - your references use single l.
Can we use one way of spelling it? Which?

andreas-junghanns (approved):
I just saw we use modeling (single l) all over the paper, so I guess we have to leave it as is.

PR #1418, opened 2021-04-23
Remove start value for clock; formatted non-normative text

andreas-junghanns:


IZacharias (commented):
@andreas-junghanns The changes are ok. But shouldn't we also remove the start attribute from fmi3Clock in fmi3Variable.xsd?

KarlWernersson:
We still need to specify initial and start attribute for a clock. 
Are they even allowed to have one.
1. The simplest is to claim that inital == calcualted and there fore they don't have a start attribute. 
This is very pragmatic but not very correct, as for an input clock the initial is in general not computed internally but given by the master.

2. To state that that input clock do not have inittal and start atribute is more correct but a bigger rule exception.

I would prefer 2 but I am ok with both

IZacharias:
> We still need to specify initial and start attribute for a clock.
> Are they even allowed to have one.
> 
> 1. The simplest is to claim that inital == calcualted and there fore they don't have a start attribute.
>    This is very pragmatic but not very correct, as for an input clock the initial is in general not computed internally but given by the master.
> 2. To state that that input clock do not have inittal and start atribute is more correct but a bigger rule exception.
> 
> I would prefer 2 but I am ok with both
> 

@KarlWernersson @andreas-junghanns I would also prefer to delete both attributes from the clock variable. We all know that the Clock is a very special kind of variable, therefore exceptions to the general rules are ok. The usage of initial==calculated would seem strange to me.

andreas-junghanns:
Removing the 'start' attribute from the Clock type is simple (done already).
We cannot as simply remove 'initial' - it is inherited through fmi3VariableCore.
We have verbally restrict this.
@KarlWernersson : Can you propose what we should add into the description for initial (the table where I removed the sentence).

chrbertsch:
Regular Design Meeting: 
@t-sommer will make the changes to the text
@andreas-junghanns wil fix the text
Then we will ask @KarlWernersson  to review it.

andreas-junghanns:
@KarlWernersson : Could you please have a look and see if this is ok now so we can merge these changes ASAP?

KarlWernersson (approved):
Looks good.
By restricting xml verbal restriction are avoided =)

andreas-junghanns:
@IZacharias : you approved previously and we followed your suggestions, so I will merge assuming your approval.

PR #1423, opened 2021-04-28
Change resourceLocation to resourcePath

### LINKED ISSUE #1405: ResourceLocation: lifetime?, read-only?, path or URI?, point to "resources" or root?

t-sommer:
fixes #1405


chrbertsch (changes requested):
I would not change the name, as we are in beta phase now. 
Thus I propose to close this PR.

t-sommer:
> And we decided in #1405 "to leave things as they are" for the resrourceLocation.

This "decision" refers to question 4. "Make resourceLocation point to root of ZIP archive?", not 1. - 3.

chrbertsch:
> This "decision" (actually just a poll) refers to question 4. "Make resourceLocation point to root of ZIP archive?", not 1. - 3.

Ah, sorry your are right. The description text is good.

But why change the name resourceLocation to resourcePath?Is it to make the change from URI to a path explicit?

t-sommer:
> But why change the name resourceLocation to resourcePath?Is it to make the change from URI to a path explicit?

Yes. @klausschuch proposed it, and I think it's a good idea.

andreas-junghanns:
> I would not change the name, as we are in beta phase now.

This is "just" the name of the function argument - one could argue that this is not a change to the API and it clarifies what the argument is (stressing the difference to 2.0).
I think the change is a good idea.

andreas-junghanns (approved):

chrbertsch (approved):

t-sommer:
Do we need to vote on this?

chrbertsch:
> Do we need to vote on this?

I do not think so. The PR is open for review, anyone can object to the changes. It would be good regarding the removial of the URI to get feedback from @TorstenBlochwitz and @IZacharias , @MBlesken  and perhaps tool vendors who do remote simulation of FMUs.

I asked several colleagues who run FMU simulations in the cloud, but no one needs the URI, a folder path is sufficient for them.

Regarding the lifetime and read-only: do we need a clarification for FMI2.0.x for that? Or would that break anything?

TorstenBlochwitz (approved):
Very nice!

PR #1428, opened 2021-04-30
Fix SE concepts intro

MBlesken:
DRAFT. Fixing a false remark on output clocks by using a new formulation about countdown Cloks from https://github.com/chrbertsch/fmi3pap. Compare chrbertsch/fmi3paper#4 (comment)


andreas-junghanns:
We (Andreas and TorstenB) are confused why you say `Model partitions that are associated to countdown Cloccks`...
Can't (shouldn't) we be more general and note this about all input clocks?

MBlesken:
@IZacharias @andreas-junghanns please **both** review this PR before merging. Thank you very much. Everyone else is welcome, too, of course.

andreas-junghanns (approved):
I think this is good now

andreas-junghanns (approved):

IZacharias (approved):

Issue #1421, opened 2021-04-27
State that activateModelPartition() does not activate input clock?


oliverkotte:
Shall the spec clearly state that activateModelPartition() does not set the input clock to ClockActive, as setClock() does for ME/CS? Then it becomes also clearer that activating/deactivating the input clock is not needed when implementing activateModelPartition(). This was not immediately obvious to me when browsing the spec (although it is implicit at least in Table 5) but it became obvious when reading Claudio's sentence:
> Note the difference between the partition activation function and the clock set/get functions: the later inform the Importer that a task should be scheduled, while the former executes as part of the previously scheduled task.

andreas-junghanns:
@oliverkotte : Sounds good: please create a PR.

chrbertsch:
So we never call setClock or getClock in SE, right? Then we should state this also in "2.2.9.5. Clocks specific API"

oliverkotte:
The importer calls getClock on output clocks in CallbackIntermediateUpdate in SE.
But it never calls setClock in SE, right?

andreas-junghanns:
@IZacharias or @MBlesken : Please comment here.

oliverkotte:
If setClock is never called in SE, input clocks in SE would tick but they would never be activated, only their associated model partition would be

IZacharias:
@oliverkotte 
> The importer calls getClock on output clocks in CallbackIntermediateUpdate in SE.
> But it never calls setClock in SE, right?

Yes, that is correct.

> If setClock is never called in SE, input clocks in SE would tick but they would never be activated, only their associated model partitions would be by way of activateModelPartition

Maybe I don't understand the question correctly. The call to the fmi3ActivateModelPartition tells the FMU that the clock is activated. What exactly do you think we should adjust in the specification?

andreas-junghanns:
> If this is so, I propose to clarify this in the text.

@oliverkotte : Could you please create a PR?

PR #1430, opened 2021-05-03
Remove reference to input clock activations from SE text

oliverkotte:
link to issue #1421

I noted that "clock activation" and "clock tick" seemed to be used synonymously.
If you agree to this issue, then clock ticks and clock activations become two different things, hence my changes "clock activation" -> "clock tick" when referring to input clocks.

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -853,7 +853,7 @@ include::../headers/fmi3FunctionTypes.h[tags=GetClock]
 include::../headers/fmi3FunctionTypes.h[tags=SetClock]
 ----
 
-[[fmi3SetClock,`fmi3SetClock`]] is never called in Scheduled Execution.
+<<fmi3SetClock>> is never called in Scheduled Execution.
 
 For some Clock types, the interval [[fmi3SetInterval,`fmi3SetInterval`]] is set by the environment for the current time instant by the function <<fmi3SetIntervalDecimal>> or <<fmi3SetIntervalFraction>>.
 The values of the arguments `interval`, `intervalCounter`, `resolution`, `shift` and `shiftCounter` refer to the unit of the <<independent>> variable.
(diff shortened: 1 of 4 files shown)


docs/2_2_common_mechanisms.adoc
```diff
@@ -574,7 +574,8 @@ The following Clock types are described in detail after the table.
 Input Clock::
 [[inputClock, `input Clock`]]
 is a variable of type Clock with <<causality>> == <<input>>.
-The importer will activate input Clocks using <<fmi3SetClock>> or <<fmi3ActivateModelPartition>>.
+In Model Exchange and Co-Simulation, when input Clocks tick, the importer will activate the input Clocks using <<fmi3SetClock>>.
+In Scheduled Execution, when input Clocks tick, the importer will activate the associated model partitions (but not the input Clocks) using <<fmi3ActivateModelPartition>>.
```
chrbertsch, line 578: I think it is the other way round, when a model partition is activated, the clock ticks ...

oliverkotte: I'd say the clock tick determines the time instant of the model partition's activation and is causally first. E.g. a triggered input clock might tick because a triggered output clock of another FMU ticked and the two are connected; the tick of the input clock then clearly causes and precedes the activation of the model partition.

andreas-junghanns (changes requested):
@oliverkotte : This is getting more clear - thank you! We need to have @MBlesken or @IZacharias agree as well. And I would like you to also fix "clock tick" in the glossary - that would be the main place where we define the difference for ME/CS and SE.

andreas-junghanns (approved):

chrbertsch (approved):

andreas-junghanns:
@IZacharias | @MBlesken : Can one of you please approve these changes?

IZacharias (approved):

PR #1432, opened 2021-05-04
Update logging documentation

t-sommer:


andreas-junghanns (approved):

t-sommer:
If this was the intended meaning of the arguments, we should also back-port this to FMI 2.0.x.

pmai (approved):
Looks good to me.

PR #1433, opened 2021-05-07
Add missing function defines for shift accessors

pmai:
The function defines in fmi3Functions.h for the newly introduced clock shift accessor functions were missing.


andreas-junghanns (approved):
That was my bad, I think, sorry. Thanks for spotting and fixing it!

t-sommer (approved):

Issue #1412, opened 2021-04-14
Clarify allowed causalities for clockedstate/state variables and their previous values


pmai:
Seperated from the discussion in #1407 we have to work out the allowed causalities for clocked/discrete states and for their previous value variables, and starting from this we need to look at the allowed and allowable initial and start value attributes for those.

clagms:
I've placed a sentence in https://github.com/modelica/fmi-standard/pull/1413 that can act as a placeholder for the information that might be added in the current issue:
```
the variable referenced by the previous attribute must have a type that is compatible with this variable.
```

andreas-junghanns:
> the variable referenced by the previous attribute must have a type that is compatible with this variable.

In that PR I already changed this to require "same type".
Compatible types would need definition and with signed/unsigned and min=0 we can get a lot of "smart-people" comments about what can be considered compatible... I just feel we are not asking too much to ask for "the same" as internally, previous values should be the same type anyway (in sane implementations, I know).

andreas-junghanns:
@KarlWernersson : can you please propose what exactly we should clarify for this issue? Which combinations of attributes we should check?

KarlWernersson:
It has to be discrete and  (only leaves local and output)
and it has to be of the same type (already specified)

Input is a bit tricky, it would be quite wrong to let previous variable (the variable referenced as in the previous attribute) be an input, however an input can have a previous attribute  (the FMU buffers it internally and want to expose it)

So maybe it is sufficient to change the current text 

> and the variable referenced by the previous attribute must have the same numeric type

to

> and the variable referenced by the previous attribute must have the same numeric type and cannot have causality == input

chrbertsch:
@KarlWernersson or @pmai : could you please prepare a PR?

chrbertsch:
Regular Design Meeting: 

Karl: they have to be discrete, stated elsewhere.
Pierre: Why not  parameters?
Karl: They are fixed or tunable.
Pierre: "discrete" means "variability=discrete"

@clagms : I will create a PR

PR #1436, opened 2021-05-12
Clarifies the characteristics of the previous value

clagms:
Fixes https://github.com/modelica/fmi-standard/issues/1412


pmai (approved):
Looks good to me, I think.

chrbertsch:
@KarlWernersson  and others: Can we merge this?

PR #1456, opened 2021-06-14
Improve description of NULL arguments

andreas-junghanns:


KarlWernersson:
@t-sommer 
For FMU state Andreas only changed form null->NULL
It was already specified like this in fmi 2.0, my guess is that the designed was based on free().

andreas-junghanns:
@t-sommer : so far this PR only fixes spelling. Can you agree with that. All other changes require more discussion.

t-sommer (approved):

PR #1457, opened 2021-06-14
Remove Clocks from tables 19 and 20; remove restriction Clock/initial

andreas-junghanns:


docs/2_4_common_schema.adoc
```diff
@@ -1150,15 +1148,6 @@ If <<initial>> is not present, its value is defined by <<table-definition-initia
...
-^|--
-^|--
-^|--
-^|[.underline]#exact#
-^|[.underline]#calculated#
-^|[.underline]#calculated#
-^|--
 |====
 
 _[Note: For local and output variables and <<initial>> = <<exact>>, then the variable is explicitly set in <<InitializationMode>>._
```
KarlWernersson, line 1153: Maybe ad a second note to be explicit

_[Note2: Clocks do not have initial attribute]_

KarlWernersson (changes requested):
Looks good, comment about adding an additional note to the table if you agree

docs/2_4_common_schema.adoc
andreas-junghanns: I added a normative sentence to that effect - please check.

KarlWernersson (approved):

Issue #1450, opened 2021-06-13
Evaluation of outputs only in clocked partitions


rfranke:
Let a clocked partition be described with state equations (1) and output equations (2)
```
xd(k) = f[xd(k-1), u(k), t(k)],     (1)
yd(k) = g[xd(k), u(k), t(k)]        (2)
```
Sometimes it is needed to only evaluate output equations (2) for given states xd(k). This is similar to the initialization of clocked continuous-time models, where equation (1) above is replaced with xd(k)=xd(k-1).

The concept [Franke, Mattsson, Otter, Wernersson, Olsson, Ochel, Blochwitz: Discrete-time models for control applications with FMI, Modelica Conference 2017](https://ep.liu.se/ecp/132/057/ecp17132507.pdf) introduced the argument `subactive` to `fmi2SetClock`. 

The current documentation of FMI 3 defines
```
typedef fmi3Status fmi3SetClockTYPE  (fmi3Instance instance,
                                      const fmi3ValueReference valueReferences[],
                                      size_t nValueReferences,
                                      const fmi3Clock values[],
                                      size_t nValues);
```
with
```
typedef            bool fmi3Clock;    /* Data type to be used with fmi3ClockActive and
                                         fmi3ClockInactive */

/* Values for fmi3Clock */
#define fmi3ClockActive   true
#define fmi3ClockInactive false
```

How can the evaluation be restricted to output equations?

clagms:
Thank you for laying out the equations in the issue @rfranke . Makes it possible to be more precise.

> How can the evaluation be restricted to output equations?

Maybe I misunderstand the problem, but isn't this done by the FMU, when `getReal( yd )` is invoked, and the clock `d` is active?

In other words. Suppose clock `d` is active at super-dense time instant `(t, i)`, then the FMU knows that equations (1) and (2) have become active (active equations means part of them will be computed when `getReal` is called for one of the variables).
When the importer invokes `getReal( yd )`, then `f` is computed (because `g` depends on `xd` and then `g` gets computed).

If the importer did not trigger the update of all variables in `xd` or `yd`, then the FMU will do so anyway, when the `UpdateDiscreteStates` function is called.

As far as I understood, this "lazy" evaluation model no longer requires the use of the subactive flag, but I might be wrong of course, and there was quite some confusion regarding this flag.
See also: https://github.com/modelica/fmi-standard/issues/890

rfranke:
[...]

PR #1462, opened 2021-06-23
Introduce f_disc to separate discrete state equations from f_event

rfranke:
This shall resolve issue #1450.
It enables the evaluation of output equations f_event independently
from state equations f_state with the fmi3Clock value fmi3ClockSubactive.
This is particularly needed for model exchange.

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -793,39 +793,39 @@ _An example is provided in <<example-scheduled-execution>>.]_
 
 ===== Model Partitions and Clocked Variables [[clocked-variable]]
 
-Each clock latexmath:[C] induces a discrete system, whose state advances at each of latexmath:[C]'s tick.
-Such system, also called the <<model-partition>> of latexmath:[C], is commonly written as:
+Each Clock latexmath:[\mathbf{k}] induces a discrete system.
+It's state advances at each tick of latexmath:[\mathbf{k}] at time latexmath:[\mathbf{t}_{k}].
+Such a system, also called the <<model-partition>> of latexmath:[\mathbf{k}], is commonly written as:
 
 latexmath:[\begin{align*}
-x^C_d(k) &= f^C(x^C_d(k-1), u^C(k), t(k)) \\
-y^C_d(k) &= g^C(x^C_d(k), u^C(k), t(k))
+\mathbf{x}_k(\mathbf{t}_{k}) &= \mathbf{f}_k(\mathbf{x}_k(\mathbf{t}_{k-1}), \mathbf{u}(\mathbf{t}_k), \mathbf{t}_k) \\
+\mathbf{y}_k(\mathbf{t}_{k}) &= \mathbf{g}_k(\mathbf{x}_k(\mathbf{t}_{k}), \mathbf{u}(\mathbf{t}_k), \mathbf{t}_k)
 \end{align*}]
 
 where:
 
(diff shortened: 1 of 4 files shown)


andreas-junghanns:
@rfranke : Please check my tiny change: I do not think that f_event should compute x_d - I removed x_d from the left-hand side of f_event to reflect this.
Reasoning: Your introduction of f_disc should give you complete control over computation of discrete states. If a getXXX triggers such a computation implicitly, this would not give you that desired control.

Another issue comes to mind: Some models might not be able to separate f_event and f_disc. Should we introduce a capability flag for the FMU `providesDiscreteStateControl`? If that is the case, we would have to explain that in the text as well.

andreas-junghanns (approved):
With my changes, I think this is ok - it makes the machinary more difficult to operate (one more function to support to split functionality of the update).
I am still wondering about the addition of a capability flag to indicate support of this feature, but this can be dealt with in a different PR.

rfranke:
@andreas-junghanns your change looks good to me.

chrbertsch:
Regular Design Meeting: 
- Do we know of any tool that would not support the separation of the evalation  / update functions.
- Rüdiger: It might be tools that do not separate the previous.
- Claudio: Change "Activate output clocks and respective model partitions for the evaluation of discrete states in fdisc", as this seems not to be correct regarding the output clocks, I will change this
- Rüdiger: Change Activate clocks and respective model partitions in fevent
- Claudio: I wil change it.
- Masoud: I will check the case where the f_disc calculation is implicit.
- single out the question on the capability flag  (Christian)
- Mousoud: What is the use case for this features?
- Rüdiger: e.g. partial derivatives; in ME we need control over the evaluation of functions.

clagms:
I've added a bit more, describing a bit more on the model partitions, and trying to summarize the semantics described in https://github.com/modelica/fmi-standard/issues/1450#issuecomment-866886765

@rfranke please have a look.

@andreas-junghanns  to make the PR more "complete", I think we still need to define the signature of the function `fmi3EvaluateDiscreteStates`. If this PR is to be merged without that, then we need to create an issue to not forget this.

andreas-junghanns (commented):
@clagms : your latest changes are not acceptable: we have a section with mathematical description elements. We should not start just a few paragraphs below with a totally new section with new mathematical symbols that that never appear again.
Can you please use the definitions already layed out and if some semantic is missing, add it to what is alredy there in "2.1. Mathematical Definitions".

clagms:
Understood @andreas-junghanns . However, right now I'm pretty flooded and I think only in 3-4 weeks will I find a bit of time to work on this. If you want to merge this PR without my additions, please revert and merge.

andreas-junghanns:
@clagms and @rfranke : I changed your formulas to fit the math notation specified in "2.1 Mathematical Definitions" - please check that I did not mess up your intent.

andreas-junghanns:
@TorstenBlochwitz : This was merged without your feedback - any second thoughts?

chrbertsch:
@rfranke  Thanks for your work. The proposal seemed to be in a good state.
however there are no documented approved reviews. Have you coordinated with someone before merging this?
As the last changes were made by Andreas, I consider him to approve.
@clagms : could you please also comment, if every thing was fine for you?

andreas-junghanns:
Sorry, @rfranke : Big Brothers are watching, not just you ;)

t-sommer:
@rfranke, next time please rebase or squash (instead of merge) the PR, so we get a more streamlined `master` branch.

Issue #1305, opened 2021-02-17
Compiler and platform tuple


MBlesken:
The current specification has lacks and inaccurate references to compiler and platforms:

A few examples
- The possible values of the _attribute_ compiler of _SourceFileSet_ is not precisely defined. Only examples are given. (An importer has to identify the compilers it supports.)
- The Application Binary Interfaces (ABI) listed for the _platformTuple_ intended to define the directory for binaries is likely incomplete and lacks of concrete descriptions (E.g. "gnu": If sources are compiled for linux _elf_ format is created. If you compile for windows the _pe_ format is created. But _pe_ is not listed. So should gnu be used here?)
- The ABI Versions only lists MSVC

The question is: **Is it intended to describe these details in the specification or can we provide a concrete reference an FMU developer has to regard?**

chrbertsch:
Must these definitions be _in_ the FMI3.0 standard, or could we reference e.g. to a website, where we could update it independent of the release cycle of the FMI standard?

andreas-junghanns:
@t-sommer : Please add examples for the most common cases for source, static binary and dynamic binary.

MBlesken:
> The current specification has lacks and inaccurate references to compiler and platforms:
> 
> A few examples
> 
> * The possible values of the _attribute_ compiler of _SourceFileSet_ [NOW: -> BuildConfiguration, PR #1465] is not precisely defined. Only examples are given. (An importer has to identify the compilers it supports.)
> * The Application Binary Interfaces (ABI) listed for the _platformTuple_ intended to define the directory for binaries is likely incomplete and lacks of concrete descriptions (E.g. "gnu": If sources are compiled for linux _elf_ format is created. If you compile for windows the _pe_ format is created. But _pe_ is not listed. So should gnu be used here?)
> * The ABI Versions only lists MSVC
> 
> The question is: **Is it intended to describe these details in the specification or can we provide a concrete reference an FMU developer has to regard?**

@t-sommer @andreas-junghanns @chrbertsch These parts of the specification have not been addressed by PR #1316. (My PR #1465 isn't addressing it either.) **Could you please comment on the issues I marked?** I'm not an expert and also would need support here, but from what I got when discussing with colleges is that these parts of the standard are not helpful with the intended workflow.

andreas-junghanns:
[...]

PR #1465, opened 2021-06-25
Build configuration compiler

MBlesken:
@andreas-junghanns  @t-sommer Why is compiler an attribute of SourceFileSet? I propose to move it one level up: I cannot imagine a BuildConfiguration should be able to contain sources to be built with several different compilers. Also a static library needs to provide information on the compiler that was used. 

This PR is somehow related to #1305 
FYI: @IZacharias


t-sommer (changes requested):
The motivation was to support configurations like Fortran + C that require this flexibility.

MBlesken:
> The motivation was to support configurations like Fortran + C that require this flexibility.

@t-sommer Thanks for the reminder, I will drop the PR.

PR #1466, opened 2021-06-28
Add more abstract type images

andreas-junghanns:


andreas-junghanns:
Darn, found #1459 too late...

PR #1473, opened 2021-07-07
replace rigorous

### LINKED ISSUE #1471: Clarification for Table 19: clocks attribute
In Table-19 for the "clocks" attribute, it is mentioned that 
"[More rigorous importers requiring a variable to be dependent on a single clock can reject FMUs violating this restriction.]"
I wonder what "More rigorous importers" exactly means. If assigning a variable to several clocks is allowed, it is then a correct FMU and the FMU may be simulated. It is also fine and normal that an importer that cannot simulate this FMU, reject it.

chrbertsch:
resolves #1471


masoud-najafi (approved):

andreas-junghanns (approved):
I am not sure why you propose this change. What is wrong with "rigorous"?
But I will not oppose this change, more important issues require our attention!

andreas-junghanns:
Please use your fork to create PRs.

Issue #1470, opened 2021-07-06
needsExecutionTool attribute


masoud-najafi:
In the standard i is mentioned that "The name of the tool can be taken from attribute generationTool in <fmiModelDescription>.
But the exporter tool and the importer tool may be different, or their version may be different.
needsExecutionTool is currently a boolean value,  shouldn't needsExecutionTool  be a string (with empty default value) indicating the name of the require is d tool?

chrbertsch:
Regular Design meeting: 
Torsten: this flag just signals to the user that the FMU is not to simulated just as it is. It is not complete.
One could just add an error message
Christian: Best provide information on the needed tool etc. in the documentation folder.

I propose to add a sentence in the implementers guide.

andreas-junghanns:
@chrbertsch : If this is a point for the implementer´s guide, can you add it to the list there and close this issue here?

chrbertsch:
added to https://github.com/modelica/fmi-standard/issues/1029#issuecomment-876591291

PR #1478, opened 2021-07-08
Relax requirement for including dependencies

### LINKED ISSUE #1477: Relax requirements for including binary dependencies
Currently the docs state that

> Dynamic link libraries must include all referenced resources that are not available on a standard target machine [for example, DLLs on Windows that are built with Visual Studio should be compiled with the /MT option to include the required symbols from the Visual C runtime in the DLL, and not use the option /MD where this is not the case]. When compiling a shared object on Linux, RPATH="$ORIGIN" has to be set when generating the shared object in order that shared objects used from it, can be dynamically loaded.

However this renders any FMU illegal that cannot link statically against the VS runtime or depends on DLLs that "are not available on a standard target machine" (whatever that means), e.g. commercial third party libraries or applications that cannot be shipped with the FMU.

I propose to relax to this to _should include_ and also backport this change to `2.0.x`.

See also https://github.com/modelica/fmi-cross-check/issues/57.

t-sommer:
fixes #1477

Update 1 after my discussion with @andreas-junghanns:

This proposal is by no means an invitation to link dynamically, but to legalize the large number of FMUs that are doing it already (for various reasons) and will continue to do it. Not allowing  it (even though it is clearly discouraged) will e.g. lead to the exclusion of a large number of FMUs from the Cross-Check, because they are clearly illegal according to the current specification (see also https://github.com/modelica/fmi-cross-check/issues/57).

Update 2: 

At least on Windows we have to distinguish between two types of dependencies on external DLLs:

1. Load time dependencies that are declared in the `<modelIdentifier>.dll`. These can be determined before calling `LoadLibrary()` and an error message can be displayed to the user.

2. Runtime dependencies where the external DLLs are loaded _after_ `LoadLibrary()` has been called. In this case the FMU can generate an error message that can (and must) be displayed to the user if `fmi*Instantiate*()` returns `NULL`. Also this approach does not lead to DLL-hell, as the paths to the DLLs are specifically chosen by the FMU.

As I understand @andreas-junhanns, he wants to avoid being blamed as an importer for bugs in the FMU rather than forbidding dynamic loading of DLLs in general.

### DIFF
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -25,36 +25,37 @@ Every FMU is distributed with its own ZIP file.
 [[figure-zip-structure]]
 ----
 // Structure of ZIP file of an FMU
-modelDescription.xml          // description of FMU (required file)
-documentation                 // directory containing the documentation (optional)
-   index.html                 // entry point of the documentation
-   diagram.png                // descriptive diagram view of the model (optional)
-   diagram.svg                // if existing the diagram.png is required (optional)
+modelDescription.xml                // description of FMU (required file)
+documentation                       // directory containing the documentation (optional)
+   index.html                       // entry point of the documentation
+   diagram.png                      // descriptive diagram view of the model (optional)
+   diagram.svg                      // if provided, diagram.png is also required (optional)
+   externalDependencies.{txt|html}  // required, if external resources are needed to load or simulate the FMU
    <other documentation files>
-   licenses                   // directory for licenses (optional)
-      license.{txt|html}      // Entry point for license information
-      <license files>         // For example BSD licenses
-icons                         // FMU and terminal icons (optional)
-   terminalsAndIcons.xml      // description of terminals and icons (optional)
-   icon.png                   // image file of icon without terminals (optional)
-   icon.svg                   // if existing the icon.png is required (optional)
-   // all terminal and fmu icons referenced in the graphical representation
-sources                       // directory containing the C sources (optional)
+   licenses                         // directory for licenses (optional)
+      license.{txt|html}            // entry point for license information
+      <license files>               // for example BSD licenses
+icons                               // FMU and terminal icons (optional)
+   terminalsAndIcons.xml            // description of terminals and icons (optional)
+   icon.png                         // image file of icon without terminals (optional)
+   icon.svg                         // if existing the icon.png is required (optional)
+                                    // all terminal and fmu icons referenced in the graphical representation
+sources                             // directory containing the C sources (optional)
    buildDescription.xml
-binaries                      // directory containing the binaries (optional)
-   x86_64-windows             // binaries for Windows on Intel 64-bit
-      <modelIdentifier>.dll   // shared library of the FMI implementation
-      <other DLLs>            // the DLL can include other DLLs
-   x86_64-windows-msvc140mt   // static libraries for 64-bit Windows generated
-      <modelIdentifier>.lib   // with Visual Studio 2015 with /MT flag
-   i686-linux                 // binaries for Linux on Intel 32-bit
-      <modelIdentifier>.so    // shared library of the FMI implementation
-   aarch32-linux              // binaries for Linux on ARM 32-bit
-      <modelIdentifier>.so    // shared library of the FMI implementation
-   x86_64-darwin              // binaries for macOS
-      <modelIdentifier>.dylib // shared library of the FMI implementation
-resources                     // resources used by the FMU (optional)
-extra                         // Additional (meta-)data of the FMU (optional)
+binaries                            // directory containing the binaries (optional)
+   x86_64-windows                   // binaries for Windows on Intel 64-bit
+      <modelIdentifier>.dll         // shared library of the FMI implementation
+      <other DLLs>                  // the DLL can include other DLLs
+   x86_64-windows-msvc140mt         // static libraries for 64-bit Windows generated
+      <modelIdentifier>.lib         // with Visual Studio 2015 with /MT flag
+   i686-linux                       // binaries for Linux on Intel 32-bit
+      <modelIdentifier>.so          // shared library of the FMI implementation
+   aarch32-linux                    // binaries for Linux on ARM 32-bit
+      <modelIdentifier>.so          // shared library of the FMI implementation
+   x86_64-darwin                    // binaries for macOS
+      <modelIdentifier>.dylib       // shared library of the FMI implementation
+resources                           // resources used by the FMU (optional)
+extra                               // Additional (meta-)data of the FMU (optional)
 ----
 
 ===== Directory Documentation [[documentation-directory]]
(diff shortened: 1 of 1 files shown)


andreas-junghanns (changes requested):
This is dangerous. Esssentially, this is opening a legal door to DLL hell.
I know we already have FMUs out there that are not fully static and are illegal that way but this way the importers are not to blame if things break.
Again: If we allow dynamic linking it will be unclear who is responsible about providing those dependencies.
Note: even for the simple case of VS, linking with /MD you are NOT allowed to ship the VS runtime stuff - the importer has to make sure the redistribution packages are installed. For each version of VS.
I am not saying that this is sometimes unavoidable, I know static linking is a luxury that sometimes is unatainable, but at least the standard should set clear expectations.
This has been discussed way to often this was repeatedly refused and I still think for good reasons.

This proposal here is way too soft. If we open the door for dymamic linking, we should at least:
 - require some form of dependency documentation
 - shift the burdon to the exporter if things fail, that is, proper error messages MUST be logged to the importer as ERROR to make sure the user knows the FMU makes a certain dynamic load request that the importer or the system it runs on must be fulfilled.

Needless to say: I would rather not allow this.

t-sommer:
> [...] error messages MUST be logged to the importer as ERROR to make sure the user knows the FMU makes a certain dynamic load request that the importer or the system it runs on must be fulfilled.

This is already required by the specification (see https://fmi-standard.org/docs/3.0-dev/#FMUStateSetable):

>  If a NULL pointer is returned, then instantiation failed. In that case, logMessage is called with detailed information about the reason.

> require some form of dependency documentation

The dependencies are already implicitly declared either in the shared library itself (for load time dependencies) or through the mandatory error message (for lazy loading).

@andreas-junghanns, can you propose a change that makes this clarifies this?

chrbertsch:
Isn't this related to the needsExecutionTool flag? 
If there is an external dependency for the FMU to be executed, IMHO this flag should be set and it should be documented, of what this dependency existst (and how to install tools, needed DLLs etc.)
see https://github.com/modelica/fmi-standard/issues/1470

t-sommer:
I think historically `needsExecutionTool` indicates the requirement to install an external tool in the sense of tool coupling. If we repurpose the flag to indicate _any_ dependency on external resources, we should probably rename it to `needsExternalResources` and update the documentation accordingly.

casella:
I understand a fundamental requirement of FMUs is to be self-contained and portable to the maximum possible extent, so that they can be reused and shipped to third parties with ease. So, the ideal situation is a self-contained set of C sources, that can be compiled with any standard ANSI C compiler, and/or a statically compiled DLL, that can run anywhere.

I fully agree with @andreas-junghanns' comment, in particular with this statement:
> If we allow dynamic linking it will be unclear **who is responsible about providing those dependencies**.". 

This is the killer argument for me.

I find @t-sommer's argument
>This proposal is by no means an invitation to link dynamically, but to legalize the large number of FMUs that are doing it already (for various reasons) and will continue to do it. Not allowing it (even though it is clearly discouraged) will e.g. lead to the exclusion of a large number of FMUs from the Cross-Check, because they are clearly illegal according to the current specification

rather questionable. 

We've already been in a similar situation 10 years ago, when it became clear that the Modelica Standard Library 3.2 was not standard at all, because in a quite large number of occasions it did not actually follow the Modelica Specification 3.2. That was obviously a big deal. The way we solved it was to fix the Modelica Standard Library as much as we could so that it actually followed the Specification, even though that actually required some work. I think we should do the same with those FMUs in the cross-check, i.e. re-generate them so that they actually follow the standard. Old FMUs that worked can still be used, but new ones should stick to the standard.

As it turned out back then, there were some features in the MSL that were actually quite cleverly designed and useful, e.g. the gravitity function in the outer System object of the MultiBody library, that was actually not allowed by the lookup rules of Modelica 3.2. In those cases, the best solution was to amend the Specification to allow them, which was actually done twice with 3.2r1 and 3.2r2. 

All this was done cooperatively with contribution from all tool vendors and library developers, and I believe that was a success story for the MA. I am convinced we should address this issue in the same spirit.

Summing up

- I agree with @andreas-junghanns that a world of self-contained FMUs that do not depend on DLLs is a better world
- I think the FMI Specification has a point in this respect and should not be relaxed light-heartedly
- MA standards should not legalize de-facto situations, but rather strive for good, rational, open, portable, reusable, and user-friendly software
- If it turns out that there is a real, _documented_ need of dependencies _in real applications_ (never mind the cross-check examples), the standard should explain clearly how they should be managed. I understand managing DLL hell is a hell of job, so I would personally avoid going in this direction
- Even if we allowed dependencies on DLLs, we should not put FMUs that actually have those dependencies in the FMI-cross-check pool of examples, because that would cause unnecessary waste of time and frustration. Most importantly, they will lead to unfair results, based on factors that are completely independent form the quality of implementation of FMUs, which is what the cross-check should certify

t-sommer:
> * If it turns out that there is a real, _documented_ need of dependencies _in real applications_ (never mind the cross-check examples), the standard should explain clearly how they should be managed. I understand managing DLL hell is a hell of job, so I would personally avoid going in this direction

You could argue that approach 1. (see above) is too hard to implement, but what speaks against approach 2. in your opinion?
I can give you a list of real world industrial applications where this is necessary for legal and efficiency reasons.

> * Even if we allowed dependencies on DLLs, we should not put FMUs that actually have those dependencies in the FMI-cross-check pool of examples, because that would cause unnecessary waste of time and frustration. Most importantly, they will lead to unfair results, based on factors that are completely independent form the quality of implementation of FMUs, which is what the cross-check should certify

The XC rules are not subject of this PR.

andreas-junghanns (approved):
I have no better suggestions and it seems clear that we need to relax our rules.
So with trepidation, I propose to merge this.
However, this needs further discussion in a whider group!

chrbertsch (changes requested):
As discussed a flag `hasExternalDependencies` should be added similarly to "needsExecutionTool"

t-sommer:
> As discussed a flag `hasExternalDependencies` should be added similarly to "needsExecutionTool"

I'm aware of the discussion yesterday, however I think it introduces a redundancy, since this information is already contained in the presence (or absence) of the file.

andreas-junghanns:
> > As discussed a flag `hasExternalDependencies` should be added similarly to "needsExecutionTool"
> 
> I'm aware of the discussion yesterday, however I think it introduces a redundancy, since this information is already contained in the presence (or absence) of the file.

@chrbertsch : I was similarly cautious to accept this reduction, but @t-sommer pointed out that platform support must also be taken from the existence of files. So the mechanism to use a file as a "flag" is not a new mechanism.

andreas-junghanns (changes requested):
Should we not point to the externalDependencies.* file in the text?

hubertus65 (commented):
I agree with this proposal.

chrbertsch:
> 
> 
> Should we not point to the externalDependencies.* file in the text?

I agree. I would suggest something like:  "If the FMU has external dependencies that are needed to load or simulate the FMU (examples: ...), a file hasExternalDependencies* must be present within the documentation folder, which contains a human-readanble list and/or description of these dependencies."

chrbertsch (approved):

andreas-junghanns (approved):

PR #1479, opened 2021-07-08
State that the value of the independent variable is not defined for SE

### LINKED ISSUE #1458: Does an independent variable make sense in Scheduled Excuction?
see also https://github.com/modelica/Reference-FMUs/issues/107

t-sommer:
fixes #1458


andreas-junghanns (changes requested):
In "5.2.2. State: Clock Activation Mode" t is used.
I don't think that the independent variable is "not defined" - it might be that some FMUs cannot return it, but when you call getXXX(), the variable you receive has a value w.r.t. the current time.
This currently proposed change confuses me more than it helps.
We must have approval from @IZacharias and @MBlesken.

andreas-junghanns (approved):

t-sommer:
FMI Design Meeting: The intermediate solution for zero-time execution model is to return the last activation time.

MBlesken:
@andreas-junghanns @t-sommer a getXXX() call by the simulation algorithm does not make sense (in SE) since it is the simulation algorithm itself that defines the time. This in mind I suggest to re-think
> FMI Design Meeting: The intermediate solution for zero-time execution model is to return the last activation time.

and consider retuning fmi3Error.

docs/2_4_common_schema.adoc
```diff
@@ -963,7 +963,9 @@ Local variable values must not be used as input to another model or FMU.
 All variables are a function of this <<independent>> variable.
 <<variability>> must be <<continuous>>.
 Exactly one variable of an FMU must be defined as <<independent>>.
-If the unit for the independent variable is not defined, it is implicitly`unit = s`.
+For <<fmi-for-scheduled-execution,Scheduled Execution>> the value of the independent variable is not defined _[since each <<model-partition>> has its own time]_.
+_[The main purpose of this variable is to define a quantity and unit for the independent variable.]_
```
MBlesken, line 967: quantitiy -> value?

docs/2_4_common_schema.adoc
```diff
@@ -963,7 +963,9 @@ Local variable values must not be used as input to another model or FMU.
 All variables are a function of this <<independent>> variable.
 <<variability>> must be <<continuous>>.
 Exactly one variable of an FMU must be defined as <<independent>>.
-If the unit for the independent variable is not defined, it is implicitly`unit = s`.
+For <<fmi-for-scheduled-execution,Scheduled Execution>> the value of the independent variable is not defined _[since each <<model-partition>> has its own time]_.
```
MBlesken, line 966: the value is defined by the simulation algorithm

MBlesken (changes requested):

pmai:
If the current proposal would be that SE FMUs must have an independent variable, but that access to it would produce fmi3Error, then this must surely be reconsidered; why have a mandatory variable that is not allowed to be accessed (and causes simulation abort). Given that SE probably does not make sense for time units that are not s (or smaller factors thereof), this should be taken into account.

PR #1484, opened 2021-07-21
Improve clocks math description

clagms:
This is the continuation of https://github.com/modelica/fmi-standard/pull/1462 and https://github.com/modelica/fmi-standard/issues/1450

IMHO, there are still issues to be discussed before merging this. 
See the above issue for the continuation of the discussion.

PR #1485, opened 2021-07-22
New attempt at fixing formulas and descriptions for f_disc and f_event

andreas-junghanns:
With PR #1484 not quite right, this is another attempt to describe both EventMode functions.


andreas-junghanns:
The definition of "previous(v)" is still a bit odd, but I want to keep it "simple" (even if Claudio isn´t quite happy with that, yet).
The rest is done, except for the image where we could show what each signals previous value is... not today.

clagms:
I defined the previous, but also cannot find the time right now to address the remaining issues:
- figure is inconsistent with notation (use of index k in clock values does not make sense)
- unify superdense time notation with the notation for even instances (uses n subscript to represent different event instants).

andreas-junghanns:
I had to change the definition of previous as also continuous variables can change (non-continuously) during Event Mode.

clagms:
There are still some issues with the last commit:
- the cases are not mutually exclusive (on first inspection).
- t_0 is not defined on the first case (should be 0 I think).
- we need to make it clear or the first two cases that the variable is continuous or discrete.

andreas-junghanns:
@clagms : I tried now and I am playing with another idea to make it more obvious, but I think what is there now should at least be correct. Could you please check?

rfranke (approved):
It's fine for me. The second `f_event` might be renamed to `f_discevent` to express clearer that it combines `f_disc` and `f_event` into one.

masoud-najafi:
HI 
A few remarks for PR 1485:
--- It's state advances at each tick of   -> Its state advances at each tick of

--Then
Such a system, also called the <<model-partition>> of latexmath:[\mathbf{k}], is written as
either in explicit form ( when providesEvaluateDiscreteStates capability flag is False)


latexmath:[\begin{align*}
(\mathbf{x}_k) &:= \mathbf{f}_{disc}(^\bullet{\mathbf{x}_{k-1}}, \mathbf{u_k}, \mathbf{t_k}) \\
(\mathbf{y}_k) &:= \mathbf{f}_{event}(\mathbf{x}_{k-1}, \mathbf{u_k}, \mathbf{t_k})
\end{align*}]

or in implicit form (when providesEvaluateDiscreteStates capability flag is True)

latexmath:[\begin{align*}
0 &:= \mathbf{f}_{disc}(^\bullet{\mathbf{x}_k}, \mathbf{u_k}, \mathbf{t_k}) \\
(\mathbf{y}_k) &:= \mathbf{f}_{event}(\mathbf{x}_k, \mathbf{u_k}, \mathbf{t_k})
\end{align*}]


Note that in the explicit form, \mathbf{x}_k  is evaluated using GetXX(ref(x_k)) where ref(x_k) is given in ModelStrcuture->ClockedState.
On the other hand, in the implicit form, The first equation (which is an implicit equation) is evaluated using fmi3EvaluateDiscreteStates(comp).

Masoud: If you all agree, I can continue to elaborate.

rfranke:
I do not agree with your proposal. You miss the important point of a state transition in
```
  x(k) := f_disc(x(k-1), u(k))
```
that is explicit. Also the output function
```
  y(k) = f_event(x(k), u(k))
```
is explicit.
The capability flag `providesEvaluateDiscreteStates = false` just means that an FMU cannot separate state transitions from output evaluations, i.e. it has
```
  y(k) = f_event(f_disc(x(k-1), u(k)), u(k)) == f_discevent(x(k-1), u(k))
 ```

PR #1488, opened 2021-07-23
Relax requirement for including dependencies

### LINKED ISSUE #1486: Back-port relaxation of requirement for including dependenciesto FMI 2.0.3
back-port modelica/fmi-standard#1478 to FMI 2.0.3, according to steering committe decision from July 21st 2021.

(This will influence https://github.com/modelica/fmi-cross-check/pull/124)

t-sommer:
and introduce documentation/externalDependencies.{txt|html}

fixes #1486

### DIFF
--- a/docs/2_3_fmu_distribution.adoc
+++ b/docs/2_3_fmu_distribution.adoc
@@ -29,6 +29,7 @@ modelDescription.xml      // Description of FMU (required file)
 model.png                 // Optional image file of FMU icon
 documentation             // Optional directory containing the FMU documentation
     index.html            // Entry point of the documentation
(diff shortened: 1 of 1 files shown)


andreas-junghanns (approved):

chrbertsch (commented):
I personally would prefer a "must" for the presence of the new file at least in case of external library dependencies.  This would really force to document the dependencies.
But I can live with the "should`".

Should we  list "license files and servers" in the "e.g." bracket explicitely?

t-sommer:
> I personally would prefer a "must" for the presence of the new file at least in case of external library dependencies.

As discussed in the SC meeting, this is not possible for FMI 2.0. Hence "should".

PR #1491, opened 2021-07-27
Clarify value of the independent variable

### LINKED ISSUE #1490: Can the independent variable be discrete?
If a CS FMU has a fixed step solver, the internal time makes discrete steps. Should the required independent variable be `causality = discrete` in this case, or should it always return the last time that has been set by the importer?

t-sommer:
fixes #1490

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -53,7 +53,7 @@ _[Example of a definition in this header file:_
 +
 [source, C]
 ----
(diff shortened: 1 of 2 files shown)


andreas-junghanns:
@TorstenBlochwitz : Please review this PR.

pmai (approved):
Waiting on @TorstenBlochwitz review, but seems an overall improvement for me. The issue of whether SE has any idea of a time variable is a separate one, so does not have to be addressed here.

MBlesken:
This PR seems very similar to #1479. Do we need both?

PR #1505, opened 2021-08-03
Remove obsolete ban to change callback functions

### LINKED ISSUE #1504: Remove obsolete ban to change callback functions
> It is not allowed to change these functions between fmi3InstantiateXXX and fmi3Terminate calls.

t-sommer:
fixes #1504

### DIFF
--- a/docs/2_3_common_states.adoc
+++ b/docs/2_3_common_states.adoc
@@ -136,7 +136,6 @@ See <<preemption-support>> for details.
 +
 The arguments <<logMessage>>, <<intermediateUpdate>>, <<preemption-support,`lockPreemption`>>, and <<preemption-support,`unlockPreemption`>>, are function pointers provided by the simulation environment to be used by the FMU.
 Each of these pointers can be NULL to indicate missing support for the respective functionality.
-It is not allowed to change these functions between <<fmi3Instantiate>> and <<fmi3Terminate>> calls.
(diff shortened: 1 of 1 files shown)


pmai (approved):
OK for me: pointing out that the pointers to the functions must remain valid during that time frame seems a bit over the top.

andreas-junghanns:
> 
> 
> OK for me: pointing out that the pointers to the functions must remain valid during that time frame seems a bit over the top.

For the record:
The sentence now removed originated from the fact that the pointers lived inside a structure. It had to be said that the importer should not change those function pointer with the expectation that the FMU then uses the new function pointer value.
The obsolesense stems from us moving the function pointers into the argument lists of the instantiation functions and a later change would not have been possible (pass by value, of course).

pmai:
> 
> 
> > OK for me: pointing out that the pointers to the functions must remain valid during that time frame seems a bit over the top.
> 
> For the record:
> The sentence now removed originated from the fact that the pointers lived inside a structure. It had to be said that the importer should not change those function pointer with the expectation that the FMU then uses the new function pointer value.
> The obsolesense stems from us moving the function pointers into the argument lists of the instantiation functions and a later change would not have been possible (pass by value, of course).

That is the background; however the pointers passed in could still be made invalid during the life time of the FMU. Pointing this out in addition was what my comment considered over the top, i.e. uneccesary (you have to go to some trouble to invaldate function pointers, so naive implementers will not do so, and sophisticated implementors will know better).

Issue #1444, opened 2021-06-07
Allow Clock variables to have a previous attribute?


clagms:
From the document, it's not explicitly forbidden to have a Clock variable with a previous attribute.
I think this should not be allowed, and, if so, the constraint might be added somewhere near text:
```
only clocked variables (i.e. variables with the attribute clocks) may have the previous attribute
```

PS - Issue identified by @nickbattle while upgrading the [FMI check tool](https://github.com/INTO-CPS-Association/FMI-VDM-Model)

andreas-junghanns:
"... this should not be allowed..." : Can you propose what we should write in the text as to why this should be forbidden?
Is it a matter of semantic inconsistency (e.g. "a clock cannot be a state and therefore cannot have a previous value").

This is related to #1442, because only clocked variables can have previous variables and if a Clock variable is a clocked variable, it could be a state. I will remark in #1442 w.r.t. to this.

clagms:
Thanks for the quick feedback @andreas-junghanns . 
I will try to provide PRs for these and the other recent issues I've opened, by the next weekend.
Just need your sanity check on them to make sure they are issues indeed.

andreas-junghanns:
> 
> 
> Thanks for the quick feedback @andreas-junghanns .
> I will try to provide PRs for these and the other recent issues I've opened, by the next weekend.
> Just need your sanity check on them to make sure they are issues indeed.

@clagms : Did this get solved with the recent changes?

clagms:
Thanks for the reminder @andreas-junghanns .
This kind of depends on https://github.com/modelica/fmi-standard/issues/1442 because it could be fixed by simply disallowing `clocks` attribute for Clocks:
```
The following constraints [to the previous attribute] apply: only clocked variables (i.e. variables with the attribute <<clocks>>) may have the <<previous>> attribute; 
```

t-sommer:
FMI Design Meeting: remove "previous" attribute for Clocks.

PR #1509, opened 2021-08-06
Add previous constraint for Clocks #1444

andreas-junghanns:
Sorted constraints for previous attribute and added Clock as exception.


pmai (approved):
Looks good.

PR #1515, opened 2021-08-13
Explain qualifier values

andreas-junghanns:


masoud-najafi (approved):
What would be qualifier value at the very first invocation of fmi3GetIntervalDecimal()? fmi3IntervakUnchanged would mean unchanged WRT the value in XML ?

andreas-junghanns:
> 
> 
> What would be qualifier value at the very first invocation of fmi3GetIntervalDecimal()? fmi3IntervakUnchanged would mean unchanged WRT the value in XML ?

That would be a reasonable assumption, I suppose. This is different though for non-periodic clocks that have no default in the XML - see your other ticket.

andreas-junghanns:
@TorstenBlochwitz : pushed both PRs now, please check

TorstenBlochwitz (approved):
Much better now.

PR #1521, opened 2021-08-17
Clarify deactivation of Clocks

andreas-junghanns:


masoud-najafi (approved):
Thanks, it is perfect. But this statement implies that some clock related APIs should absolutely be called before calling fmi3UpdateDiscreteStates(). For example, fmi3GetIntervalX for "countdown"-interval clocks (do you agree?). I need to check if this is not violation from this restriction in the standard.

TorstenBlochwitz (approved):
For me, it looks good now and I would propose to merge this PR.

@masoud-najafi : Please discuss your concerns in the corresponding ticket. Please explain, why the new formulation for you "implies that some clock related APIs should absolutely be called before calling fmi3UpdateDiscreteStates(). For example, fmi3GetIntervalX for "countdown"-interval clocks (do you agree?)."

PR #1523, opened 2021-08-18
Relax and clarify time-based clock activation

andreas-junghanns:


masoud-najafi (approved):
This good, but there is one thing that I do not understand. A triggered clock tick should also be active in only one super-dense time. I mean, if this clock needs to be repeated in another super-dense time, it is not the same clock. That means that in the first super-dense time, the clock has triggered itself. This new clock should be treated as ANOTHER clock, it is not the SAME clock tick. Do you have a simple example to check?

andreas-junghanns:
For triggered clocks we even allow deactivation (if supported by the FMU). So activation has different meaning. I am sure you mean active still at updateDiscreteStates(), right?
I would not restrict multiple activations during one Event Mode, even in that case, after all: it is only activating a model partition and maybe this is exactly what some use case out there needs. And I cannot see this to harm: if you can activate a set of equations once, why not several times.
If that is a concern to anyone, we might have to introduce a capability flag indicating if the FMU supportsMultipleClockActivationsDuringOneEventMode.

masoud-najafi:
It is too early to think about this. Let's find a simple model (composed of several FMUs), then we could get back to this point.

PR #1526, opened 2021-08-23
Clarify clocked variable get/set

andreas-junghanns:


masoud-najafi (approved):

TorstenBlochwitz (approved):
We shifted the sentences is a bit without changing semantics, compared to what Masoud reviewed.

PR #1537, opened 2021-08-27
Remove clocked variables from StepMode math

andreas-junghanns:


andreas-junghanns:
Added a sentene as per request.

masoud-najafi (approved):

TorstenBlochwitz:
The changes look good to me. But when I read the document, I asked myself, if we clearly enough mention that clocked variable can be accessed only if one of its clocks ticks. We mention this only in one half sentence: 

The discrete-time variables xk and yk are called clocked variables. A clocked variable vk can acquire a new value and can be queried only when its Clock k is active, except during Initialization Mode.

Is this sufficient?

TorstenBlochwitz:
We file a new ticket on that.

TorstenBlochwitz (approved):
OK. much better now.

PR #1540, opened 2021-08-30
prepare 2.0.3 release

chrbertsch:
Prepare 2.0.3, please review.

Release date is assumed to be 1.10.2021. If we are fast with inspection and voting, we could perhaps finalize the voting within the MAP FMI before the Modelica Conference, but not the voting of MA. Let's discuss this in tomorrows steering committe meeitng.


andreas-junghanns (approved):
The changes in the .adoc files look good to me. I did not check the .docx, just hoping you did this in parallel.

PR #1559, opened 2021-11-03
Clarify issues 1558

andreas-junghanns:


docs/2_3_common_states.adoc
```diff
@@ -880,12 +892,22 @@ Getting variables might trigger <<selectiv-computation,computations>>.
...
+
 Function <<fmi3GetDirectionalDerivative>>::
+No restrictions.
+
 Function <<fmi3GetAdjointDerivative>>::
 No restrictions.
 
 Function <<fmi3GetOutputDerivatives>>::
-Not allowed in Model Exchange.
+Not allowed in Co-Simulation and Scheduled Execution.
```
KarlWernersson, line 913: Not allowed in Model Exchange and Scheduled Execution

docs/2_4_common_schema.adoc
```diff
@@ -1428,8 +1427,8 @@ All variables listed in this element must have attribute `derivative` (in order
 
 |`reinit`
 |[[reinit,`reinit`]]
-Only for Model Exchange (if only a Co-Simulation FMU, this attribute must not be present.
-If both Model Exchange and a Co-Simulation FMU, this attribute is ignored for co-simulation): +
+Only for Model Exchange (if only a Co-Simulation or Scheduled Execution FMU, this attribute must not be present.
+If both Model Exchange and a Co-Simulation or Scheduled Execution FMU, this attribute is ignored for Co-Simulation and Scheduled Execution): +
```
KarlWernersson, line 1431: Should we reomve the requirement to not have it for co-simulation or scheduled execution, isn't it sufficient to, to stat that is it ignored for those cases?

KarlWernersson: I see potential for making erroneous modelDescription.xml say that you can export as both ME and cs but choose only CS, then it is high risk that this gets there any way.

KarlWernersson (changes requested):

docs/2_4_common_schema.adoc
andreas-junghanns: Fixed as suggested, now:
Only used in Model Exchange, ignored for the other interface types.

docs/2_3_common_states.adoc
andreas-junghanns: I changed it as suggested, now consistend with "Getting Derivatives of Continuous Outputs"

andreas-junghanns:
@TorstenBlochwitz : I followed Karl´s suggestions. Please review and approve, then I will merge.

TorstenBlochwitz (approved):
Much better now.

PR #1562, opened 2021-11-10
Remove argument "instanceName" from fmi3CallbackLogMessage

### LINKED ISSUE #1561: Remove redundant argument "instanceName" from fmi3CallbackLogMessage
since it is known to the importer and can be retrieved via `instanceEnvironment` if need.

t-sommer:
fixes #1561


andreas-junghanns (approved):
nice catch!

PR #1576, opened 2021-11-18
Forbid use of GetOutputDerivatives in SE

andreas-junghanns:
What about all other derivative functions in SE?


andreas-junghanns:
Sorry, I messed up my branches... bear with me...

PR #1578, opened 2021-11-18
Rename XML attribute interval of Clock variable to clockType

andreas-junghanns:


chrbertsch:
It is a little confusing that now we have two times "Clock type" in the first line of table 5, once as a general headline and once with a slightly different meaning for the xml attribute. 

grafik

Should we perhaps delete the first occurance of "clock type" there and leave this field empty, or rename it to "Clock properties"?

andreas-junghanns:
> rename it to "Clock properties"?

Done as suggested, plus header rearranged.

chrbertsch (approved):

docs/2_2_common_mechanisms.adoc
```diff
@@ -586,23 +586,26 @@ While the importer is the source of the actual Clock activations, the timing of
...
 [[table-overview-clocks]]
-[cols="9,9,16,9,40,40",options="header"]
+[cols="9,9,16,9,40,40"]
 |====
-2+|Clock type
-|<<causality>>
-|<<interval>>
-|Related API calls and arguments
-|Example
+2.2+h|Clock properties
```
MBlesken, line 596: I think we should discuss in the next meeting if "clockType" is the best alternative to "interval". I guess that people would expect e.g. "periodic" to be a "clock type" because that is the _major_ feature of a clock. On the other hand I think that people are expecting a numerical value when the term "interval" is used.

So I propose "**intervalType**" as alternative to "interval" as it was used in the standard to subsume _constant_, _fixed_, ...

andreas-junghanns:
Pierre: `intervalVariability`: Andreas, Pierre, Klaus, Masoud
Matthias: `intervalType`: Matthias

Web Design Meeting: change to `intervalVariability`

PR #1581, opened 2021-11-24
Clarify allowed calling of fmi2GetXXX in initialization mode

### LINKED ISSUE #1554: Conflicting statements about when usage of fmi2GetXXX functions is allowed.
The 2.0.2. standard isn't very clear or is even giving conflicting information about when it is allowed to use the `fmi2GetXXX` functions.

In **2.1.7** it says:
`The actual values of the variables that are defined in the modelDescription.xml file can be inquired
after calling fmi2EnterInitializationMode with the following functions:`

It doesn't give any restrictions other than that `fmi2EnterInitializationMode` must have been called.

In **4.2.4** it says that `fmi2GetXXX` functions can only be called in certain states. Specifically when in _InitializationMode_ on page 107 :
`The variables that can be retrieved by fmi2GetXXX calls are (1) defined in the xml file under <ModelStructure><InitialUnknowns>, and (2) variables with causality = ″output″.`

Later in the chapter (page 109) it says that while in _InitializationMode_ the `fmi2GetXXX` functions can be called
`for a variable with causality = "output" or continuous-time states or state derivatives (if element <Derivatives> is present)`

My first concern is that 2.1.7 gives the impression that it always allowed to call the `fmi2GetXXX` functions (after initialization) while chapter 4.2.4 says it's only allowed in certain states. My interpretation is that, since 2.1.7 refers to the application programming interface, that the information given in 2.1.7 merely means that it is always (after initialization) possible to read the FMU-Internal value buffers for the variables. It doesn't give any guarantees that the returned values are meaningful, though. Is that interpretation correct?

My second concern is that conflicting information is given in 4.2.4: It's my understanding that `variables listed under <ModelStructure><InitalUnkowns>` and `continuous-time states or state derivatives` are not exactly the same thing and it's not clear which of these two (different) statements has the priority.

chrbertsch:
resolves #1554


andreas-junghanns (approved):

andreas-junghanns:
@chrbertsch : what is the status here?

chrbertsch:
It think we have the same ambiguity in ME as in CS, and should fix this also. Do you agree?

andreas-junghanns (approved):

masoud-najafi (approved):

PR #1582, opened 2021-11-25
Add clarification on OS services

andreas-junghanns:


docs/1___overview.adoc
```diff
@@ -67,8 +67,12 @@ The interface types have large parts in common, defined in <<fmi-common-concepts
 In particular:
 
 * FMI Application Programming Interface \(C) -- <<general-mechanisms>> +
-All required equations or tool coupling computations are evaluated by calling standardized C functions.
+All required equations or tool coupling computations are evaluated by calling standardized C functions from the importer into the FMU.
+The FMU can signal certain events back to the importer using callback functions provided by the importer.
 C is used because it is the most portable programming language today and is the only programming language that can be utilized in all embedded control systems.
+The FMI API does not restrict what operating system services an FMU can use on the platform it runs on.
```
MBlesken, line 73: Good.

docs/1___overview.adoc
```diff
@@ -67,8 +67,12 @@ The interface types have large parts in common, defined in <<fmi-common-concepts
...
+However, for maximal portability, any dependency on the target platform should be minimized and operating system services should be accessed only through standard libraries.
```
MBlesken, line 74: It think this helps a lot!

So as you said in the meeting you decided not to add a hint that an FMU that is compiled or can be compiled with gcc and only depends on the GNU Libc is likely supported by most linux based FMU importers. Okay.

docs/1___overview.adoc
```diff
@@ -67,8 +67,12 @@ The interface types have large parts in common, defined in <<fmi-common-concepts
...
+Special run-time requirements should be documented in the appropriate directory inside the <<fmu-distribution,ZIP file>>.
```
MBlesken, line 75: I think this is very important for each vendor investing into FMU support as they are aiming for high compatibility with other tools.
* Can we emphasize it a little more? Maybe "... it is recommend..." or similarly?
* I am not sure if "run-time requirements" is too vague. What are you thinking of? Does this include dependencies on certain linux packages as well?

docs/1___overview.adoc
```diff
@@ -172,12 +176,12 @@ Else: icon:times[]
 
 === Properties and Guiding Ideas
 
-In this section, properties are listed and some principles are defined that guided the low-level design of the FMI.
-This shall increase self consistency of the FMI functions.
-The listed issues are sorted, starting from high-level properties to low-level implementation issues.
+In this section, properties are listed and some principles are defined that guide the design of the FMI API and XML schema itself (not the content of the FMUs).
+Following these principles shall help increase self consistency of the FMI API functions.
```
MBlesken, line 180: I am not sure what you mean by "self consistency of the FMI API functions".

docs/1___overview.adoc
```diff
@@ -194,6 +198,7 @@ It is possible to distribute an FMU without knowing the target processor.
 This allows an FMU to run on a PC, a Hardware-in-the-Loop simulation platform or as part of the controller software of an ECU.
 Keeping the FMU independent of the target processor increases the usability of the FMU.
 To be processor independent, the FMU must include its C (or C++) sources.
+To be maximally portable, FMUs must reduce their dependency on OS services and use these only through standard library calls.
```
MBlesken, line 201: Add? "...any dependency ... documentation/externalDependencies.{txt|html} .... "

docs/2_2_common_mechanisms.adoc
```diff
@@ -25,7 +25,7 @@ _[For an example of `NULL` being explicitly allowed see <<resourcePath>>._
 _Defensive implementations should still guard against `NULL` pointers.]_
 
 * The FMI Standard does not provide a runtime platform or portability layer.
-Access to operating system resources and services, such as memory and file system, must be implemented with special care because the availability of such resources and services is not guaranteed.
+Access to operating system resources and services, such as memory, network or file system, must be implemented with special care because the availability of such resources and services is not guaranteed.
```
MBlesken, line 28: Very good. Maybe "... is not guaranteed on every simulator.".

MBlesken (approved):
Approved with comments. Thank you for the valuable improvements!

docs/1___overview.adoc
andreas-junghanns: Yes, I think llvm and hopefully future initiatives like that will improve the situation that gcc is currently in (dependencies on certain version of stdlib etc.) - enshrining gcc seems the wrong hint here.

docs/1___overview.adoc
andreas-junghanns: - "should" is stronger than "is recommended" (See RFC 2119), if I remember correctly.
- "run-time requirements" must be vague, I think: We have no way of listing all potential dependencies that might or might not be available from one platform to the next, especially not for the future. Most of this is truely academic, because no one in their right mind would introduce a run-time dependency without need. What we need to say here is that the standard is silent on this issue and leaves this to the care of the FMU exporter to document "what is needed".

docs/1___overview.adoc
andreas-junghanns: I have an idea who wrote this a long time ago :)
Well, even outside psychology, "self consistency" requires "likeness of approach within the group".
In essense, this covers trivially naming conventions, argument ordering, array vs. pointer usage, but also more complex behavioral concepts like error handling, return value use, side-effect conventions and the like.
I am quite happy to have this vague here as it should capture whatever we mean in terms of similarity between the API functions or even the XML elements and their structure.
If you have a better way to describe this to mean "as-similar-as-possible-with-respect-to-natural-differences" - I am open for suggestions.

docs/1___overview.adoc
andreas-junghanns: Sorry, I am not sure what you mean... the need for documentation was covered above. "their dependency" means "all". What is missing?

docs/2_2_common_mechanisms.adoc
andreas-junghanns: Yes, I like that addition. Done.

chrbertsch (approved):

PR #1583, opened 2021-11-25
Allow fmi3XXXDerivative in SE

andreas-junghanns:


andreas-junghanns:
Note: I did not point again to the capability flags to avoid redundancy - this is described in the link these functions point to.
Some other places might need a consistency check the way we state "allowed", "conditionally allowed", "not allowed for context X" - if you find those, please open another issue.

docs/2_3_common_states.adoc
```diff
@@ -903,10 +901,10 @@ Function <<fmi3GetEventIndicators>>::
 Not allowed in Co-Simulation and Scheduled Execution.
 
 Function <<fmi3GetDirectionalDerivative>>::
-No restrictions.
+See <<fmi3GetDirectionalDerivative>>.
```
MBlesken, line 904: Isn't that very confusing to read with all the linking? I think it is already confusing that the section is called "Allowed Function Calls" and there are a lot of functions listed with a remark "Not allowed ..."

MBlesken (approved):

chrbertsch (approved):

PR #1592, opened 2021-12-03
Some changes regarding intervalQualifier and getInterval

PTaeuberDS:
This pull request includes following changes:

1. Some clarifications on the handling of intervalUnchanged and intervalNotYetKnown
2. Some restructuring of the SE code example
3. Adding getInterval to allowed functions of Clock Activation Mode
4. Removing of redundant parameters of clock specific getter and setter functions that are relicts of when clocks were not restricted to be scalars (e.g. nIntervals in getInterval). 
I'm not sure if the redundant parameter maybe was kept intentionally for some reasons. However, it is probably confusing for the reader to have two parameters for the same value. An alternative would be to state explicitly that nValueReferences and nIntervals always have the same value. Not very nice, though.

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -952,7 +952,7 @@ with the following meanings:
  * `fmi3IntervalChanged` is returned to indicate that the value for the interval has changed for this <<Clock>>.
  Any previously returned intervals (if any) are overwritten with the current value.
  The new <<Clock>> interval is relative to the time of the current <<EventMode>> or <<ClockUpdateMode>> in contrast to the interval of a periodic <<Clock>>, where the interval is defined as the time between consecutive <<Clock>> ticks.
- In Scheduled Execution this means that the corresponding Clock has to be scheduled or re-scheduled (if a previous call to <<fmi3GetInterval>> returned <<fmi3IntervalChanged>>).
+ In Scheduled Execution this means that the corresponding model partition has to be scheduled or re-scheduled (if a previous call to <<fmi3GetInterval>> returned <<fmi3IntervalChanged>>).
 
  [[fmi3IntervalUnchanged,`fmi3IntervalUnchanged`]]
   * `fmi3IntervalUnchanged` is returned if a previous call to <<fmi3GetInterval>> already returned a value qualified with <<fmi3IntervalChanged>> which has not changed since.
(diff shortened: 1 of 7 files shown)


andreas-junghanns (approved):
Except for the example where I did not fully understand all changes, the rest are good.

andreas-junghanns:
@PTaeuberDS  : Do you keep this "in Draft" for a reason? We should discuss this later today in the design meeting, right?

MBlesken (changes requested):
To me it seems you're attempting to make sense into a three value enum for the qualifier while I think that only two values are needed. As a consequence new requirements are imposed on the FMU exporter (generating the inst->CountdownClockQualifier = fmi3IntervalNotYetKnown). FYI @andreas-junghanns 

Request to change in the "Preemption section".

PTaeuberDS:
I moved the sentence to the place in the document where the qualifier is explained.
@MBlesken @andreas-junghanns Please have a look

MBlesken (changes requested):
Please check if the SE Example has to updated for GetInterval now: "inst->CountdownClockQualifier = fmi3IntervalUnchanged;" must only be called if the qualifier was fmi3IntervalChanged, right? Then: OK

PTaeuberDS:
> Please check if the SE Example has to updated for GetInterval now: "inst->CountdownClockQualifier = fmi3IntervalUnchanged;" must only be called if the qualifier was fmi3IntervalChanged, right? Then: OK

Good catch. Done

MBlesken (approved):

andreas-junghanns (approved):
I reworked the qualifier value descriptions (again) - if you agree with the new wording, please merge. Otherwise let´s find some time to discuss and change this online.

andreas-junghanns:
@MBlesken : Warning: I meddled with the text again - you might want to read the section describing the interval qualifiers again.
@PTaeuberDS : I checked your code in snippets.c and agree with the logic.

andreas-junghanns:
> I moved the sentence to the place in the document where the qualifier is explained. @MBlesken @andreas-junghanns Please have a look

I did not agree (hence the rewite): The sentence did not explain "who" needs to keep book this way. And in essence, we should not - in the standard text that is - describe how implementation is done - we need to stay on the API side.
But I agreed that the mechanism wasn´t too clear and hope this is a bit clearer now... waiting for your judgement...

TorstenBlochwitz (approved):
OK, looks good.

PR #1595, opened 2021-12-09
Rename XML attributes and function arguments

andreas-junghanns:


chrbertsch (approved):
Looks good

TorstenBlochwitz (approved):
OK

pmai:
@andreas-junghanns: Sorry missed this PR in time, however this set of changes makes `needsCompletedIntegratorStep` the sole attribute where leaving it out defaults to true. Is this really necessary? Would it not be better to have it default to `false` like all other capabilities/flags? If an FMU generator needs it, they generate a true value there, because they care. FMU generators that do not care will default to false, which is better in terms of performance etc.

If that is of interest I'll create a PR for this.

PR #1597, opened 2021-12-14
Improve event handling description

andreas-junghanns:


andreas-junghanns:
@TorstenBlochwitz & @andreas-junghanns : We refined the description of reasons for entering Event Mode and how this is done for both ME and CS in the Event Mode section. We believe this does not require a general mechanism section - it would split the little text we have even further unnecessarily.

TorstenBlochwitz (approved):
OK

Issue #1385, opened 2021-03-31
Design Walkthrough of Variable/ModelStructure


pmai:
As discussed today in the Design Meeting 2021-03-31, given the way we have incrementally added more information to variables (cf. event indicators, clocks) and the ModelStructure, we should do a complete design walkthrough of this whole area to ensure that all of the functionality we have now available is well designed and understood in its consequences and documented as such. The proposal is to collect questions for such a design walkthrough here, and then schedule a focused workshop where the design is gone over in a structured way and the collected questions (and others that arise during the walkthrough) are answered and necessary changes/enhancements to the design or documentation are collected.

Potential querstions might include:

- What does dependency information on event indicators mean, how should it be specified and how can it then be used reliably (e.g. to not query/calculate indicators that cannot have changed due to the dependencies not having changed)?

chrbertsch:
FMI Design Meeting: Let's prepare and schedule this for the next regular design meeting on May 26

@ all: please comment questions and comments to this ticket

pmai:
Further topics for design walk-through:

- Do all combinations of clock types make sense if combined via the clock attribute of a clock (i.e. can a time-based input clock have a clock attribute pointing to e.g. a triggered output clock, etc.)?
- Do all other attribute combinations on variables make sense?

chrbertsch:
Web-meeting: 
Pierre: are we stable enough w.r.t clocks? Let's wait for the outcome of tomorrow's discussion. Then I could prepare the "walkthrough" for some of the next meetings.

chrbertsch:
@pmai : could we start the walkthrought today? Will you join the meeting?

andreas-junghanns:
Clocks design walk through (by @TorstenBlochwitz & I (@andreas-junghanns)):
 - Clock `causality` is already restricted to `input` and `output`.
 - Added a note to `clocks`attribute to explain why we do not introduce further restrictions on clocked variables. See PR #1600.
 - `variability` and `previous` attribute are already properly restricted.

andreas-junghanns:
>     * What does dependency information on event indicators mean, how should it be specified and how can it then be used reliably (e.g. to not query/calculate indicators that cannot have changed due to the dependencies not having changed)?

This allows the importer to use the partial evaluation which seems to be supported by an FMU that publishes this dependency information: Before getting the event indicators only the variables they depend on have to be set. This may be a good point for the implementer´s guide.

KarlWernersson:
[...]

PR #1600, opened 2021-12-16
Add no further clock restriction

andreas-junghanns:


chrbertsch:
Shall merging this issue really close #1385  which goes beyond clocked variables?

andreas-junghanns:
Since we do not understand what a design walkthrough would do more than what we did and since nobody else seems to be interested in doing whatever-needs-to-be-done, we think this is as much as can be expected to happen.
Of course, you (or anyone else) can disagree, but I would kindly ask to suggest exactly what and especially *when* that will be done.

chrbertsch (approved):

masoud-najafi (commented):
The wording is not really clear for me. Does that mean that an input clock can depend on an output clock? If yes, if the output clock ticks, the dependent input clock should tick "synchronously"?

andreas-junghanns:
> The wording is not really clear for me. Does that mean that an input clock can depend on an output clock? If yes, if the output clock ticks, the dependent input clock should tick "synchronously"?

Not quite: Only if the output clock ticks, *may* this input clock tick.
We don´t know if this has practical applications, but according to @klausschuch - we should not forbit something just because we do not know what to use it for (as long as the mechanisms itself is clear).

masoud-najafi:
Ok. Sorry, but I need to have clear mind about it. Then, there should be an "internal" dependency between the output clock and input clock and causes this "possible" tick?

andreas-junghanns:
> Ok. Sorry, but I need to have clear mind about it. Then, there should be an "internal" dependency between the output clock and input clock and causes this "possible" tick?

The question was, if input clocks could be a clocked variable on another clocks and if there are restrictions.
One could say, it only makes sense to have an input clock be a clocked variable of another input clock.
But why be so restrictive?
Therefore, we decided NOT to restrict clocked clocks.

masoud-najafi:
But this would open the Pandora's box. Just suppose there is an output clock CO, and there is an input clock which is the clocked variable of CO. Similar to what we discussed last year, and finally it was not accepted and IntervalType of CountDown clock was introduced to disallow it.   With this feature, CountDown Clock is no more needed, I can do it with this feature. I will check it anyway if everything is fine with this new feature.

andreas-junghanns:
> But this would open the Pandora's box. Just suppose there is an output clock CO, and there is an input clock which is the clocked variable of CO. Similar to what we discussed last year, and finally it was not accepted and IntervalType of CountDown clock was introduced to disallow it. With this feature, CountDown Clock is no more needed, I can do it with this feature. I will check it anyway if everything is fine with this new feature.

A countdown clock would allow scheduling ahead of time. This here is different, because the current output clocks can only be triggered.

docs/2_4_common_schema.adoc
```diff
@@ -1091,7 +1091,11 @@ _Note: This is reverse compared to the <<derivative>> attribute.]_
 If present, this variable is clocked.
 The value of <<clocks>> is a list of value references of <<Clock>> variables this variable depends on.
 Only variables with <<variability,`variability == discrete`>> or <<variability,`variability == tunable`>> can have this attribute.
-_[Some importers might require a variable to be dependent on a single clock for technical reasons. They could reject FMUs violating this restriction.]_
+_[Some importers might require a variable to be dependent on a single clock for technical reasons._
+_They could reject FMUs violating this restriction._ +
+_Note: It is not further restricted, which variables can be clocked to not restrict currently unknown use cases._
+_For example, an <<input>> Clock could be a clocked variable of another (<<input>> or <<output>>) clock to indicate that it can only be activated when that Clock is active.]_
```
MBlesken, line 1097: @andreas-junghanns @masoud-najafi this contradicts "[Output Clocks are intended to enable an FMU to provide a trigger to the outside world, e.g. to a triggered input Clock of another FMU. Note, that if intended to activate model partitions or a Clock of the same FMU calling fmi3GetInterval for a countdown Clock and evaluating the qualifiers is recommended. An example is provided in Section 5.3.]". I would rather not have output clocks trigger input clocks of the same FMU but recommend to use countdown clocks instead.

andreas-junghanns: ... and I think we all agree with your preference. The question here is: Should we restrict the standard to mandate our preference, or should we allow such "wired" usages? In our discussion we tend to follow the Klausian camp: "Why restrict a usage simply because we do not know what it is good for."

MBlesken: @andreas-junghanns ... No we're not discussing to restrict it. It is a recommendation. To be consistent within the standard we should not give an example that ignores the recommendation. That's all. Can we please adjust this?

masoud-najafi: Suppose that an input clock c1 that depends on another input/output clock c2, then is c2 sub-samples c1, or it indicates dependency of c2 on c1? It is not clear for me....

andreas-junghanns: > @andreas-junghanns ... No we're not discussing to restrict it. It is a recommendation. To be consistent within the standard we should not give an example that ignores the recommendation. That's all. Can we please adjust this?

@MBlesken : I don´t understand: adjust what? The example? The text?

andreas-junghanns: > Suppose that an input clock c1 that depends on another input/output clock c2, then is c2 sub-samples c1, or it indicates dependency of c2 on c1? It is not clear for me....

If c1 depends on c2, how can c2 be a sub-sample of c1? I do not understand your question?

masoud-najafi:
Sorry, a mistake:
 Suppose that an input clock c1 is clocked variable of another input/output clock c2, then c1 would subsample c2? or it indicates dependency of c1 on c2?

andreas-junghanns:
> Sorry, a mistake: Suppose that an input clock c1 is clocked variable of another input/output clock c2, then c1 would subsample c2? or it indicates dependency of c1 on c2?

Well, just go through the motions: if c1 is a clocked variable of c2, then c1 can only be accessed when c2 is active. That means c1 can only be a sub-sampling clock of c2. Depending on input or output, either the importer must observe this dependency or the FMU itself, respectively.

This seems pretty clear to me. Am I missing an open point or a contradiction?

PR #1601, opened 2021-12-16
Clarify possible status in log messages

t-sommer:


andreas-junghanns (changes requested):
I think this additional restriction should be part of the existing description of `status`.

andreas-junghanns:
> I think this additional restriction should be part of the existing description of `status`.

@t-sommer : Do you agree?

andreas-junghanns:
What if an FMU is hurt so badly, that it can only return fmi3Fatal from the call? For instance, an out-of-stack error? Should the requirement be a bit softer, e.g. "unless technically impossible, the FMU should..."?

pmai:
> What if an FMU is hurt so badly, that it can only return fmi3Fatal from the call? For instance, an out-of-stack error? Should the requirement be a bit softer, e.g. "unless technically impossible, the FMU should..."?

I think this is a much better formulation, yes.

Additionally it is not completely clear to me what "less or equal" means here:

>  `status` must be less or equal to the status returned by the function that called `logMessage`.

Probably this means

>  `status` must indicate a severity less or equal to the severity of the status returned by the function that called `logMessage`.

so it is forbidden to log an fmi3Error but continue from the call with e.g. an fmi3Ok or fmi3Discard.

We can of course require that, but in all honesty I would not know why we need this: An importer will always completely rely on returned status for its algorithms, and the logging is decoupled from that in any case (i.e. even with the given requirements it is not really guaranteed that you will find matching log output for any error condition; there can be n:m relationships here).

So what if an exporter logs higher severity messages than they return? The error messages will be presented to the user in any case, and the simulation will continue, too.

If we want to tie explicit error messages to specific failures this should not be handled via restrictions on the logger callback, but rather should be handled by an fmi3GetErrorCauseMessage or some other mechanism that directly ties the message to the error.

Or in other words: Why do we need this to be clarified? I'm probably missing something important here.

andreas-junghanns (approved):

TorstenBlochwitz:
@pmai : The objective here is to enforce FMUs to provide some information about the cause of a problem. According to my experience there are lots of FMUs which silently return fmiError.

TorstenBlochwitz (approved):
OK from my side.

pmai:
> @pmai : The objective here is to enforce FMUs to provide some information about the cause of a problem. According to my experience there are lots of FMUs which silently return fmiError.

That was the objective for the original requirement to log a message if you return an error (which is given in a different section). The current changes had a different objective.

It is arguable whether anything we write here will reduce FMUs that just return an error without a log message. But restricting what error severities they may log vs. what return codes they can return in the other direction has no bearing on that.

That said I was OK with the wording now chosen, not that I have any high hopes that people will adhere to this any more than if nothing was written.

PR #1610, opened 2021-12-29
Change Table1 solver/scheduler use

andreas-junghanns:


chrbertsch (approved):

MBlesken (commented):
@andreas-junghanns did you consider adding a "solver" to figure 3 as well?

andreas-junghanns:
> @andreas-junghanns did you consider adding a "solver" to figure 3 as well?

We did and it is clear we should not: here in figure three, we do say "Scheduled" Execution: the Scheduler should be prominent and is part of the API (like the Solver in ME). IN CS, we don´t know what is inside, that is why we added both there.
This makes it also symmetrical between ME-CS and SE-CS.

MBlesken:
>We did and it is clear we should not: here in figure three, we do say "Scheduled" Execution: the Scheduler should be prominent and is part of the API (like the Solver in ME). IN CS, we don´t know what is inside, that is why we added both there.
This makes it also symmetrical between ME-CS and SE-CS.

Agreed. I like it.

PR #1612, opened 2021-12-29
Remove/change word task, improve glossary

andreas-junghanns:


TorstenBlochwitz (approved):
OK

chrbertsch (approved):

docs/2_2_common_mechanisms.adoc
```diff
@@ -830,7 +830,7 @@ where:
 . latexmath:[\mathbf{f}_{disc}] denotes the state transition function and
 . latexmath:[\mathbf{f}_{event}] denotes the output function of the <<model-partition>> of latexmath:[\mathbf{k}].
 
-_[Such a model partition may represent a task or an interrupt service routine of an embedded system, or a discretized part of a plant model._
+_[Such a model partition may represent a <<task>> or an interrupt service routine of an embedded system, or a discretized part of a plant model._
```
MBlesken, line 833: I suggest to erase "task" here because this is in particular an example where "task" is intentionally used in a vague way . But being this vague does not result in a benefit for the standard. The ISR and "discretized part of a plant model" may be sufficient as examples. The use of the term task here and the current state of the definition in the glossary even contradicts "A task can be thought of as a simple interface of the scheduler that allows to connect the FMU interface ..."  **What do you think?** -> "Such a model partition may represent an interrupt service routine of an embedded system, or a discretized part of a plant model. "

docs/7___appendix.adoc
```diff
@@ -237,8 +233,9 @@ The integer part latexmath:[\mathbf{t}_I] of this tuple is a counter to enumerat
 |_super-dense time instant_
 |See <<time-instant>> and <<super-dense-time>>.
 
-|_task_
-|Special kind of _model partition_ that is used in control code.
+|[[task,task]]_task_
```
MBlesken, line 236: If we can agree to erase "task" in docs/2_2_common_mechanisms.adoc then the term is only used in the SE example where it is stated as "A task can be thought of as a simple interface of the scheduler that allows to connect the FMU interface ...". Therefore I suggest to remove this definition which was my key intention when stating the related issue.

docs/5_3_scheduled_execution_example.adoc
```diff
@@ -7,7 +7,7 @@ Two model partitions are associated with two <<periodic-clock,periodic Clocks>>
 During the execution of the model partition of <<Clock>> `10msClock`, the FMU changes the <<interval>> for <<countdown,countdown clock>> `AperiodicClock` and calls <<fmi3CallbackClockUpdate>> to inform the importer that the interval has changed.
 The importer retrieves this interval and activates the corresponding model partition.
 
-In this example, the calls to <<fmi3ActivateModelPartition>> (as well as related <<get-and-set-variable-values,`fmi3Get{VariableType}`>> and `fmi3Set{VariableType}`) for each associated Clock are placed into a task.
+In this example, the calls to <<fmi3ActivateModelPartition>> (as well as related <<get-and-set-variable-values,`fmi3Get{VariableType}`>> and `fmi3Set{VariableType}`) for each associated Clock are placed into a <<task>>.
```
MBlesken, line 10: See my comment in glossary below.

MBlesken (commented):
I would like to request changes ...

docs/2_2_common_mechanisms.adoc
andreas-junghanns: We should at least mention control code, I will produce a "neutral" proposal.

andreas-junghanns: Reads now: "Such a model partition may represent part of control code or an interrupt service routine of an embedded system, or a discretized part of a plant model."

docs/7___appendix.adoc
andreas-junghanns: removed in PR #1613

docs/5_3_scheduled_execution_example.adoc
andreas-junghanns: Done in #1613

andreas-junghanns:
Please approve #1613 and we can resolve all pending change requests here

PR #1614, opened 2022-01-04
Improve language for entire text

andreas-junghanns:


andreas-junghanns:
We are in the middle of "Physical Units" but all sections before are "done".

docs/2_2_common_mechanisms.adoc
```diff
@@ -1,31 +1,36 @@
...
+The following callback functions lead to well-defined states and may call FMI functions according to their respective state definitions:
+ ** <<intermediateUpdate,`fmi3IntermediateUpdateCallback`>>, and
+ ** <<fmi3ClockUpdateCallback>>.
 
 * FMI functions must not change global settings which affect other processes/threads.
 An FMI function may change settings of the thread in which it is called (such as floating point control registers), provided these changes are restored before leaving the function or before a callback function is called. +
 _[This property ensures that functions of different FMU instances can be called safely in any order._
 _Additionally, they can be called in parallel provided the functions are called in different processes._
-_If an FMI function changes for example the floating point control word of the CPU, it must restore the previous value before return of the function._
-_For x86 CPUs, the floating point control word is set using the `fldcw` instruction._
```
chrbertsch, line 20: This non-normative text seems to be removed. Should it go the implementers guide?

andreas-junghanns: The first sentence is still there, just added commas.
The second is indeed removed as it is a detail for a single platform, when we really say in the example already what to do - the "how for x86" is just distracting.

docs/2_2_common_mechanisms.adoc
```diff
@@ -377,57 +382,51 @@ include::../headers/fmi3FunctionTypes.h[tags=SetClock]
...
 
 ===== Getting and Setting the Complete FMU State [[get-set-fmu-state]]
 
---
-The FMU has an internal state consisting of all values that are needed to continue a simulation.
+The FMU has internal data representing its state.
 This internal state consists especially of the values of the continuous states, iteration variables, <<parameter>> values, <<input>> values, delay buffers, file identifiers, and FMU internal status information.
-With the functions of this section, the internal FMU state can be copied and the pointer to this copy is returned to the environment.
-The FMU state copy can be set as actual FMU state, in order to continue the simulation from it.
+Depending on the FMI type, only a subset of this data is directly accessible via the FMI C-API.
```
KarlWernersson, line 419: Is it FMI type or FMU type?

KarlWernersson: or instance type?

KarlWernersson (approved):
Did a quick look, just one comment on what is the correct term to be used

chrbertsch (approved):

docs/2_2_common_mechanisms.adoc
andreas-junghanns: In the text we call it FMI type, or interface type, so I think functional mockup interface type makes most sense.

Issue #1622, opened 2022-01-10
Make explicit requirements for propagation of clock deactivations


clagms:
Right now, the standard does not talk much about propagation of clock de-activations.

Consider the following FMU:
![image](https://user-images.githubusercontent.com/871103/148739257-170e2d5e-06b6-4186-955b-9c641ce8f5d5.png)

If there is a de-activation of clock `c`, then the FMU must be responsible for de-activating clock `d` (in case clock `d` ticked as a result of `c`'s deactivation).
This complicates the implementation of the FMUs that support deactivation of clocks and should be made clear in the standard.

PS - Issued identified thanks to the work of @SimplisticCode

andreas-junghanns:
Please read the mathematical model in Event Mode, specifically the entry for `fmi3SetClock` - I think this clarifies this issue.

PS - This issue was identified and clarified thanks to previous work of @masoud-najafi.

clagms:
The bits of text I mentioned above are from the Mathematical model in Event Mode.

This is the vicinity of the excerpts provided:
````
When a Clock is deactivated, the FMU must reset all clocked states of this Clock to the values computed during the last fmi3UpdateDiscreteStates when this Clock was active (∙vk).
````

There's nothing about propagation of clock de-activations.

KarlWernersson:
Should it not solve itself?

setClock(c==true) 
getClock(d) => true. //v is evaluated
setClock(c==false) 
getClock(d) => false  //as v is not a valid variable any more.

clagms:
> Should it not solve itself?

What is "it"? The FMU? Or the Importer? I claim that it's the FMU, and that this should be clear in the standard. It's just a sentence that is needed.

KarlWernersson:
As it is is defined I don't think it is needed for output clocks, if you do nothing the updateDiscretesState should fix it if you have disabled c and as this is computed internally in the FMU so it is for me clear who should fix this.. But it doesn't hurt to be more explicit with examples in the standard.

A more tricky case is

Say you have input clock A , and input clock A, where B is a sub clock of A (i.e. it has a clock attribute with reference only to A)
if you activate both A and B and then deactivate A, who should deactivate B, importer or FMU?

clagms:
That's a good case, but is B also an input clock? 
- If it is, then if you have an input clock B that is a subclock of A, the importer will always know when to tick it. So it may not make sense for B to be a subclock of A.
[...]

Issue #1620, opened 2022-01-10
Clarify and distinguish clocks attribute from model structure dependencies.


clagms:
## The problem

Right now, the standard states, regarding the definition of the `clocks` attribute:
> Clocks may depend on input Clocks. This dependency is also defined by the `clocks` attribute.

And then in non-normative text, it states that the dependency means 
> if an output Clock depends on an input Clock, then Clock ticks of the input Clock in Event Mode or Clock Activation Mode may create output Clock ticks during this event handling.

This is wrong and leads to an inconsistent definition of the `clocks` attribute.

## Proposed Changes

1. Imho, clock dependencies should be declared in the model structure, and not in the `clocks` attribute:
   1. an output clock `C1` declares that it depends on a clock `C2` means that: `tick(C2) => maybe tick(C1)`
2. The definition of the dependency should be normative.

PS - Issued identified thanks to the work of @SimplisticCode

clagms:
There's another bit in the standard that needs to be modified, if this issue is solved:
> For example, an input Clock could be a clocked variable of another (input or output) clock to indicate that it can only be activated when that Clock is active.

andreas-junghanns:
There is a subtle difference between a dependency declared with the model structure and those with a clocks attribute.
This was discussed at length during a design meeting and we decided to leave things as they are.
I also fail to see why what you mention is inconsistent (beside: there can never be an inconsistency between normative and non-normative text - in that case the non-normative text must be considered wrong):
The normative text is about (all) Clocks, the non-normative text speaks about an example.
What am I missing?

clagms:
I vaguely remember the discussion, but the difference is not so subtle, and I was hoping that time would make it less hard to accept.

Was there are major downside to keeping the two dependencies separate? I can't remember...

The clocks attribute is used to build the clocked partition, whereas the model structure dependencies are used to optimize the propagation of changes, and checking of algebraic loops.
It seems cleaner to not mix those two uses in the same mechanism.

andreas-junghanns:
We *have* two mechanisms. What is the issue?

clagms:
The issue is that the definition of each mechanism can be made simpler and thus:
- there is no redundancy (i.e., exporter A implementing dependencies between clocks using the clocks attribute, and exporter B implementing dependencies between clocks using the model structure attributes); and
- easier to understand and implement for everyone.

I would accept the status quo if I could find an answer to the following:
> Was there are major downside to keeping the two dependencies separate?

[...]

PR #1624, opened 2022-01-11
Clarify clock dependencies requirements

clagms:
Fixes https://github.com/modelica/fmi-standard/issues/1622
Fixes https://github.com/modelica/fmi-standard/issues/1620

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -853,9 +853,8 @@ The association between <<clocked-variable,clocked variables>> and their <<Clock
 <<clocked-variable,Clocked variables>> can depend on multiple Clocks.
 _[For example, a global counter could be incremented by multiple model partitions, each controlled by a different Clock.]_
 
-Clocks may depend on input Clocks.
-This dependency is also defined by the <<clocks>> attribute.
-_[E.g. if an <<outputClock>> depends on an <<inputClock>>, then Clock ticks of the <<inputClock>> in <<EventMode>> or <<ClockActivationMode>> may create <<outputClock>> ticks during this event handling.]_
+Output clocks may also depend on input Clocks and other variables. Such dependencies are declared in the <<ModelStructure>>. A clock latexmath:[\mathbf{k}] depends on a clock or variable latexmath:[\mathbf{v}] if a tick or change of value in latexmath:[\mathbf{v}] may trigger a tick of latexmath:[\mathbf{k}], during the same super-dense time instant.
+
 
 ===== Scheduled Execution [[scheduled-execution]]
 
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -853,7 +853,9 @@ The association between <<clocked-variable,clocked variables>> and their <<Clock
 <<clocked-variable,Clocked variables>> can depend on multiple Clocks.
 _[For example, a global counter could be incremented by multiple model partitions, each controlled by a different Clock.]_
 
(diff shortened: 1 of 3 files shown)


andreas-junghanns (approved):
I had to fix a number of things in this PR and it still reads foreign to the text around it due to the mathy-kind of language that introduces symbols just for a single sentence or two. But that and formatting issues aside:

 - A change of a variable value is not the trigger condition for a clock, it is the *value* (old or new). Corrected.
 - Deactivation of clocks that depend on other activated clocks being deactivated results as a matter of cause - I corrected the language to make this more clear: *this is not a special case*. Any output value may change because a clocks activation status changed and that needs to be propagated. Clock or not. And the importer must query those values, Clock or not. It might be efficient about it knowing the dependencies, but when ignoring those, it must query all variables.
 - I re-introduced the note about the Clock dependencies. This is non-normative text and it does not contradict the normative text above, as far as I can tell, but clarifies the issue.

As the text is now, I can approve it. If we need to further discuss the changes, let´s.

clagms:
Thanks for these improvements @andreas-junghanns . 
I'd like to get @KarlWernersson's blessing before merging.

KarlWernersson (approved):
Looks good to me

clagms:
You're right @TorstenBlochwitz . 
We need to restrict the clocks attribute to variables that are not clocks.
That way, dependencies between clocks can be placed in the model structure, and clocked partition information can stay in the clocks attribute.

Besides, seeing the clock as a variable that is part of a clocked partition is very difficult to visualize in the mathematical model we provide in the standard. Is the clock also a clocked state? Does it have a previous value? 

That also means that the example reintroduced by 3f5011b750912aecdcd9e24d0be4aecf5ffdcfe0 should be omitted.

andreas-junghanns:
We have been throught this several times in the last 18 months: `dependencies` carry different semantics from what `clocks` declare. `clocks` add more information, namely that accessing variables with `clocks` attributes is restricted to when these clocks are active. Therefore: We cannot only use the `dependencies` - they would not be sufficient.
If this is not clear, we should find a good place where to non-normatively explain this.

TorstenBlochwitz (approved):
My comment was not on restricting anything. We need to clarify which functionalities are enabled by both methods.

clagms:
@andreas-junghanns I think c48b6ef42d92bf104ee641d384594d08b687efd3 is more in line with what you stated in https://github.com/modelica/fmi-standard/pull/1624#issuecomment-1010151552

Please check and merge if you approve.

andreas-junghanns:
> Please check and merge if you approve.

Sorry, no, the example was specifically referring to the not-so-obvious case of an *input* clock depending on another clock. Reverted and will now merge.

clagms:
Sure @andreas-junghanns  but you should have left the "observed" word there. It's about restricting observation and not activation 😢

andreas-junghanns:
> observed



> 



> Sure @andreas-junghanns but you should have left the "observed" word there. It's about restricting observation and not activation 😢

This sentence is about input clocks. There is no way to observe it (and no need). What am I missing?

clagms:
So you picked a non obvious example in lieu of a more obvious example (an output clock part of an input clock's model partition), and left out the bit of information that really made it explicit the difference between the clocks attribute and the dependency (the observation restriction). Why not add both?

andreas-junghanns:
Because there are descriptions of these trivial examples in the text already: "Output clocks may depend on input Clocks and other variables.".
The difference between "dependencies" and "clocks" needs some clarification. Proposal:
"Note, the dependencies between clocked variables and their Clocks are defined by the attribute clocks in <fmiModelDescription><ModelVariables>. Compared to dependencies, clocks declared dependencies also restrict when clocked variables can be accessed."
See: 262dfd1b34b894e810975dd4a2576fd74e7fb209

clagms:
Thanks @andreas-junghanns this is what was missing.

PR #1636, opened 2022-01-20
Fix state xref and formatting

andreas-junghanns:


andreas-junghanns:
work redone in PR #1614 to reduce merge conflict potentials

PR #1646, opened 2022-01-25
Do not change global settings which affect other processes/threads on preemptive systems

MBlesken:
@andreas-junghanns you may also forward my invite to review this small addition to the specification. Thank you.

FYI @PTaeuberDS


andreas-junghanns (approved):

MBlesken:
@pmai the [fmi-guide](https://modelica.github.io/fmi-guides/main/fmi-guide/) states "However they must restore the floating point control word before returning, so as to not affect state outside of the current thread of execution.". I think in a remark in the guide mentioning "To prepare the FMU code to run reliably on <<preemption-support, preemptive systems>> FMI functions must not change global settings." or alike would be useful. What do you think?

PR #1648, opened 2022-01-25
Use elements for the start values of binary variables

### LINKED ISSUE #1647: Start value for Binary variables is ambiguous
For binary variables the start value is still being provided as an attribute with a space-separated list of hex values.

This is ambiguous, because the empty string (a sequence of zero numbers) should be used to specify binary values of length 0; in a space separated list, where multiple spaces are collapsed to one separator, this does not allow the specification of 0-length entries for certain array elements.

We therefore have to switch to the element-based specification, as we already do for string variables.

pmai:
Fixes #1647

Checklist

* [x] [Linted the documents](https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#building-the-specification-document).


nickbattle (commented):
Looks good to me.

nickbattle (approved):

andreas-junghanns (approved):

PR #1649, opened 2022-01-25
Explicitly allow additional entrypoints

### LINKED ISSUE #1643: Allow additional entry points in FMU code, specify restrictions
As per the discussion on #1632, in order to properly support both minor releases with new optional features, as well as layered standards, we should clarify in the standard that additional entry points in FMUs (binary/source code) are allowed (they already have to be to follow OS ABI requirements), and that we reserve all symbols starting with fmi3 for FMI 3.x.

pmai:
Fixes #1643.

Checklist

* [x] [Linted the documents](https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#building-the-specification-document).


andreas-junghanns (approved):
Just the duplication is a bit strange.

andreas-junghanns (approved):

PR #1659, opened 2022-01-31
Make xsd capability flags extensible

pmai:
This allows arbitrary further attributes on implementation elements, thus allowing future capability flags to be introduced in 3.x minor releases, as discussed during the plugfest V.

### DIFF
--- a/.circleci/generate_figures.ps1
+++ b/.circleci/generate_figures.ps1
@@ -6,6 +6,9 @@ $generator = Join-Path $generatorDir "XSDDiagramConsole.exe"
 
 function Clear-Gen() {
 	# Get-ChildItem $pngOutDir\*
+	if (Test-Path $generatorDir){
+		Remove-Item -Recurse $generatorDir
+	}
 	Remove-Item $pngOutDir\*
 }
 


pmai:
Note that I have not yet updated the schema diagrams, because the diagram generator seems to choke on the xs:anyAttribute entry with a null pointer exception:

> Generating schema for ModelExchange...
> Loading the file: C:\Users\pmai\src\fmi-standard\schema\fmi3ModelDescription.xsd
> 
> Adding 'ModelExchange' element to the diagram...
> Expanding to level 1...
> 
> Unbehandelte Ausnahme: System.NullReferenceException: Der Objektverweis wurde nicht auf eine Objektinstanz festgelegt.
>   bei XSDDiagram.Rendering.Diagram.<>c__DisplayClass116_0.<ExpandComplexType>b__1(DiagramItem a) in H:\srcctrl\github\xsddiagram\XSDDiagrams\Rendering\Diagram.
> cs:Zeile 896.
>    bei System.Collections.Generic.List`1.FindIndex(Int32 startIndex, Int32 count, Predicate`1 match)
>   bei System.Collections.Generic.List`1.FindIndex(Predicate`1 match)
>   bei System.Collections.Generic.List`1.Exists(Predicate`1 match)
>   bei XSDDiagram.Rendering.Diagram.ExpandComplexType(DiagramItem parentDiagramElement, complexType complexTypeElement) in H:\srcctrl\github\xsddiagram\XSDDiagrams\Rendering\Diagram.cs:Zeile 896.

Don't know if we have a fix for that; otherwise we would have to drop the xs:anyAttribute thing for diagram generation...

t-sommer:
Maybe we should add a note that vendors must not introduce proprietary attributes, even if the XSD would tolerate it.

chrbertsch:
> Maybe we should add a note that vendors must not introduce proprietary attributes, even if the XSD would tolerate it.

What about layered standards? They should use annotations, right? Also Layered standards released by the FMI project?

pmai:
I have added restrictive language that restricts all non-namespace attributes for FMI use. This leaves the option of using other namespaces for other uses, e.g. in layered standards or in other circumstances where this makes sense, but disallows anything that looks like a normal capability flag and prevents clashes.

chrbertsch:
> I have added restrictive language that restricts all non-namespace attributes for FMI use. This leaves the option of using other namespaces for other uses, e.g. in layered standards or in other circumstances where this makes sense, but disallows anything that looks like a normal capability flag and prevents clashes.

I understand what "namespace" means for annotations, but what does it mean for other attributes?
Is this clearly defined?

pmai:
Here the XML namespace is meant , I've clarified this.

andreas-junghanns (approved):

andreas-junghanns:
@clagms : Can you please have a look at the error @pmai pasted here in a comment further "up"? This seems a short comming of the XSD Generation Tool 2000 you introduced - maybe there is version 3000 out now that can handle the issue here?

clagms:
I'll do so over the weekend!

clagms:
Not yet Andreas! Got sick over the weekend. Probably covid, as my partner
tested positive (and I negative but it's only a matter of time :p)

clagms:
> @clagms : have you looked into the error and do you see a way to fix it? Thanks!

Update on this: I know what the problem is. Just have to think of a way to solve it without hacks....

pmai:
> > @clagms : have you looked into the error and do you see a way to fix it? Thanks!
> 
> Update on this: I know what the problem is. Just have to think of a way to solve it without hacks....

Thank you, and all the best wishes health-wise...

clagms:
The problem should be fixed now @pmai . Just delete the folder `.circleci\xsddiagram` to force an update on the tool.
I've opted for omitting the AnyAttribute. This was the most elegant solution given the time.

The ideal solution would be to draw the element with some special notation, but this requires that a more substantial change to the original code of the tool, which has no tests, so I'd probably break something...

andreas-junghanns (approved):
The image generation works now - thank you @clagms !

chrbertsch:
To my understanding the text and xsd changes are fine, the picture generation tool has been fixed, but the re-created pictures have not yet been committed.

pmai:
I'll fix this, don't touch ;)

clagms:
> The image generation works now - thank you @clagms !

My pleasure @andreas-junghanns . There's just one problem: there's no mechanism to update the tool, and not even a way to check for updates and warn the user of the tool.
So in the future, someone who has not involved in this PR may regen the figures with v1.2 of the tool, and will get a crash. 
Then that person has no way of knowing that he/she should delete the folder, to force an update of the tool.
How do we prevent this problem?

I was thinking of adding a bit to the contributing.md file, explaining what to do in case there are problems with the generation tool... A better way would be to modify the script to check for updates, but that seems like a brittle solution...

pmai:
I have fixed the conflicts and rebased against master. The images do not need regeneration, since they come out identical when regenerated with the fixed tool (the anyAttribute element is just ignored, leading to identical files). So this could be merged.

On the update check problem: Wouldn't the simplest solution be to just always delete and redownload the tool? There are not that many users doing this manually, and for the CI it is always redownloaded anyway...

pmai:
BTW, for merging I would suggest to not squash and merge, but keep the individual commits, to make the tool changes stand out more...

pmai:
I think we can/should do it in this PR...

andreas-junghanns:
Thank you, @pmai !

PR #1672, opened 2022-02-10
Add 3rd use case for fmi3CompletedIntegratorStep

andreas-junghanns:


TorstenBlochwitz:
@Maplesoft-fmigroup: I read the proposed text too late. Shouldn't we describe a bit more in detail how the projection is done, at least that it is done by the FMU in Event Mode? And why is this flagged to the importer using a state event? Why not as a Step Event?

Maplesoft-fmigroup:
OK, how about the following instead:
* Signal when projection is necessary (for FMUs for DAE systems that
  use a projection approach) by flagging that a step event has occurred.
  The projection of the states onto the constraint manifold can then be 
  done in fmi3UpdateDiscreteStates and signalled via
  valuesOfContinuousStatesChanged being set to fmi3True.

PR #1673, opened 2022-02-10
Change == to = in text and insert spaces around =

andreas-junghanns:
This now makes the use of `=` consistent.
Note: an argument for the single `=` is that it in all circumstances this now reads the same for importer and FMU.
When using `==`, it is really only right for the side to check the conditions, the other side needs to meet that condition and why would you write it as `==`?
So after reading through the changes, I am more convinced than before that single `=` is the better solution.


chrbertsch (approved):

andreas-junghanns:
@KarlWernersson : Could you also check if this is ok with you? If we tag Beta.4 now, maybe this can make it into it?

Issue #1663, opened 2022-02-04
Clock intervalVariability


KarlWernersson:
1. 
In section 2.4.4 table 15.
The attribute interval is wrong, it is called intervalVariablitliy

2. 
Are we missing an intevalVariability?
we have  
fixed, ->inteval set by fmu during initializtaion
calculated -> intevar computed by fmu from fixed parameters
tunable -> inteval computed by fmu from tunable parameters.  

Don't we miss a set tunable i.e interval->changeable from the FMU at event mode?
I think this was one of the scenarios in @rfranke original proposal.

Is this intentional or a mistake?

I propose that we

Change name of **tunable** to **calculatedTunable**, and **calculated** to **calculatedFixed**

Add a "new" variability **tunable** That allows setInteval in event mode and clock activation mode

An alternative would be to introduce an intevalCausality but i try to avoid that.

KarlWernersson:
No, changing changes unregularly when the FMU want's it, what Ii am looking for is a clock where the period can change by calling setInterval from the importer during event mode. It is definitely not countdown, and triggered has nothing do do with interval so it is no longer an periodic clock.

To 1 Great =)

KarlWernersson:
Just by calling an interval **tunable** I would assume that you can set it as you can do that with fixed, just as for fixed and tunable parameters, I was a bit surprised it did not work as I assumed and in opposite to fixed.

KarlWernersson:
Working on the PR i realized that we only have **getShift** no **setShift** function and it is is only allowed for the **constant** and the **computed periodic clocks**, not the clocks where the interval is settable. 

Later on it says for attribute **shiftDecimal** and **shiftCounter** it states
This attribute must only be used for **periodic Clocks**
This is true but currently we are more restrictive and say that it must only be used for periodic clock with **clockVariability**
**constant**, **fixedComputed** (previously calculated) or **tunableCalculated**, (previously tunable)

**So my question is this**

Should we only support fraction for periodic clocks where the interval is set by the FMU, then it is  just a small upade for **shiftDecimal** and **shiftCounter** 

If not is it allowed for all periodic clocks?
Then I will update **fixed** and the new **tunable** accordingly.

Should it then also be possible to set the shift for **fixed** and **tunable**
If so we need the corresponding set functions.

andreas-junghanns:
Why would the importer *set* a shift? The shift must be declated by the FMU and retreived by the import with getShift.
So I don´t think setShift is not needed, for any clock.

[...]

PR #1674, opened 2022-02-11
Fixes clock interval part of #1663

KarlWernersson:


andreas-junghanns:
Note: images do not need regeneration as the XSD change only effects values of the attribute `intervalVariability` of clocks which do not show in the diagram(s).

andreas-junghanns (approved):
@KarlWernersson : I made a few minor corrections, please check that I did not invalidate any of your changes!

andreas-junghanns (commented):
I would remove both sentences saying: `This attribute cannot be combined with intervalDecimal.`
If I remember correctly, the fractions mechanic was purely optional, and now we seem to make it mandatory if an FMU supports only that. Of course it is easy to use the fractions and build decimals from that. But should we enforce this?

KarlWernersson:
that fractional is still optional is not a problem. the problem i see is if you give both a fractional and a decimal stat value. Which one to use? What if they are not the same, if 1/3 is used they cannot be the same. if fractional is given an the importer want to use decimal it is an easy computation to round a fractional to a decimal the other way around is not. 
So I am not trying to prevent decimals, just that both decimal and fractional start values are given. I want to preserve single source of truth.

Most cases this will not pose a problem but if someone want sot synchronize a clock of 1/3 and one of 2/3 in decimals that's up to them. Maybe use setIntervalDecimal to select a more appropriate one for decimal computations.

andreas-junghanns:
I think the decimals are less accurate than the fractions, except fractions are bit more difficult to handle for exporters and importers. So if both are given, you should use fractions if you know how (the Truth), otherwise the decimals are good approximations.
This would be good enough for the majority of all importers and for those that need more, they will insist on fraction definitions in the FMUs they import.

To you other points: To synchonize clocks that tick every 1/3 and 2/3 of a second should not be done using these numbers: it should be done internally in the FMU.

KarlWernersson:
What i am trying to say is that you cannot give a start value both in fraction and in decimal. The FMU must give one just as for constant, that does not prevent the master to set a different value either by fraction or by decimal if it supports both but we can only have one start value not two.

KarlWernersson:
OK but I don't think this is entirely true.

For constant intervalVariablity the exporter is limited to choose fractal or decimal representation it can't choose both.
Similar the start value for the clock interval for a fixed or a tunable describes the default value for the clock if the importer does not call setInterval on that clock. I would assume that such a clock would behave the same as a constant variability clock.

Now if we allow both decimal and fractal start values at the same time we have 2 scenarios.

1 They match. 
2. They dont match.

If 1 no problem

2 thye could missmatch due to error ie. 0.5 != 1/5
 they mismatch due to roundoff error 1/7-0.14 ~= 0.0029

By preventing the FMU from giving a start value in both decimal and fraction at the same time we avoid a lot of potential problems.

Now my idea was to enforce fractional if if supports it to make it easy to explain. But I don't see a need to enforce this I am fine that an FMU that supports fractional export decimal. Thinking on it it is probably better as it should reflect what the model exporter has selected.

But I am convicted that if we allow the xml to print a start value for both decimal and fractional at the same time for the same clock, we invite errors.

andreas-junghanns (approved):

t-sommer:
Superseded by e88d1e251c088fb8d09bdaac40dabef354895a56.

PR #1690, opened 2022-02-22
Remove redundant specification about variability/causality and initial

andreas-junghanns:


KarlWernersson (approved):
Looks good.
I se that the table is distorted here
https://fmi-standard.org/docs/3.0-dev/index.html#table-allowed-variability-causality-combinations
Maybe you fixed it already, but i am not fluent in asciidock table formating so i dont see it :(

andreas-junghanns:
> ... I see that the table is distorted here https://fmi-standard.org/docs/3.0-dev/index.html#table-allowed-variability-causality-combinations Maybe you fixed it already, but i am not fluent in asciidock table formating so i dont see it :(

Yes, we fixed that in this PR yesterday. I introduced this issue when I added the header style in the options column - which fails for combined-cell headers...

PR #1691, opened 2022-02-23
Add supportsFraction and intervalDecimal to SE example

andreas-junghanns:


andreas-junghanns:
I think this example misses not just supportsFraction but also the required intervalDecimal.

nickbattle (approved):
I've applied these changes and re-checked with VDMCheck3.sh. It now gives the following:
```
nick@pollux:~/Digital Twins/fmi-standard/docs/examples> VDMCheck3.sh scheduled_execution.xml 
Checking XML
1028: 2.4.9 InitialUnknowns must include refs: {2, 4}
Errors found.
```
VRs 2 and 4 are the output clocks, so this may be an error in the checker - it looks like clocked variables should be excluded from initialUnknowns "outputs".

nickbattle:
_> VRs 2 and 4 are the output clocks, so this may be an error in the checker_

Yes, it was excluding clocks from the test, not clocked variables. Will push the updated ZIP soon.

docs/examples/scheduled_execution.xml
```diff
@@ -18,15 +18,15 @@
 		<!-- Clock variables -->
 		<!-- Periodic input clock -->
 		<Clock name="10msClock"      valueReference="5" causality="input"  intervalVariability="constant"
-		priority="1" intervalCounter="10" resolution="1000"/>
+		priority="1" supportsFraction="true" intervalDecimal="0.01" intervalCounter="10" resolution="1000"/>
```
MBlesken, line 21: @andreas-junghanns
* isn't intervalCounter="10" resolution="1000" equivalent to intervalDecimal="0.01"? So why change it? ... why leave both options in the standard?
* supportsFraction="true" means that "...  fmi3GetXXXFraction and fmi3SetIntervalFraction are allowed to be called ..." - I don't see a need to call it for a constant clock. Do you?

andreas-junghanns: * yes, it is the same, but we made intervalDecimal required to help simple importers to ignore the fractions mechanism.
* indeed, we should add: ... as defined in Table 5.

andreas-junghanns: We need a new PR as this one was already merged. Will you (@MBlesken) provide one?

PR #1692, opened 2022-02-23
Remove computed from Clock

andreas-junghanns:


KarlWernersson (approved):
Looks good,
Thank you for fixing it.

andreas-junghanns:
@MBlesken or @PTaeuberDS : can one of you please also check to make sure this is ok for SE as well? I will merge otherwise tomorrow or so, but prefer to share the blame with all reviewers :)

TorstenBlochwitz (approved):
Nice!

PR #1695, opened 2022-02-28
Fix capitalization in "fmi3Deserialize"

t-sommer:


andreas-junghanns (approved):
Is this really important enought to be changed this late?

PR #1699, opened 2022-03-01
Clarify priority applies only to SE input clocks

andreas-junghanns:


andreas-junghanns:
Darn, this was based on an old master...

PR #1701, opened 2022-03-01
Remove redundant "shifts" and "intervals" prefixes

t-sommer:
from parameter names


andreas-junghanns (changes requested):
While this change simplifies reading the text, I wonder if `counter` can now be misunderstood as a counter of how many times a clock has ticket.
We need a statement that clarifies that the counter can be incremented by an arbitrary amount from clock tick to clock tick.
I checked the text and I am not entirely sure where this should go.

Also, there are two anchors stacked on top of one another: <<counters>> fails because of that.
I tried different things, but the problem persists...

andreas-junghanns (approved):
I added fixes to my converns. Please check those...

PR #1705, opened 2022-03-02
Add additional example on how to handle prefix for DLL building

andreas-junghanns:


Maplesoft-fmigroup (approved):

pmai (approved):

masoud-najafi (approved):
I do not like changing the header files, but since there is no other easy solution, I accept it.

PR #1706, opened 2022-03-02
Improve text for dependency

KarlWernersson:
Clarify that dependency is direct /related to partialDerivative, non normative text explaining dependency on independent variable

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -1163,25 +1163,25 @@ _[Note that array variables will be serialized, so `nSeed` is only equal to `nKn
 _Likewise `nSensitivity` is only equal to `nUnknowns` (resp., `nKnowns`) if all value references of `unknowns` (resp., `knowns`) point to scalar variables._
 
 _Note that it is the partial derivative and not the total derivative that is considered here._
-_When computing the partial derivative with respect to the independent variable_ latexmath:[\mathbf{t}] _of the function_
+_When computing the partial derivative with respect to the independent variable_ latexmath:[t] _of the function_
 
 [latexmath]
 ++++
-\mathbf{f}(\mathbf{x}, \mathbf{\dot{v}}, \mathbf{t})
+\mathbf{f}(\mathbf{x}, \mathbf{\dot{x}}, t)
 ++++
 
-_it's partial derivative with respect to_ latexmath:[\mathbf{t}] _will be_
+_it's partial derivative with respect to_ latexmath:[t] _will be_
 
 [latexmath]
 ++++
-\mathbf{\frac{\partial f}{\partial t}}
+\frac{\partial \mathbf{f}}{\partial t}
 ++++
 
 _and not_
 
 [latexmath]
 ++++
-\mathbf{\frac{\partial f}{\partial t} + \frac{\partial f}{\partial \mathbf{x}} \cdot \mathbf{\dot{v}}+\frac{\partial f}{\partial \mathbf{\dot{x}}} \cdot \mathbf{\ddot{v}}}
+\frac{\partial \mathbf{f}}{\partial t} + \frac{\partial \mathbf{f}}{\partial \mathbf{x}} \mathbf{\dot{x}}+\frac{\partial \mathbf{f}}{\partial \mathbf{\dot{x}}} \mathbf{\ddot{x}}
 ++++
 _]_
 
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -1170,7 +1170,7 @@ _When computing the partial derivative with respect to the independent variable_
 \mathbf{f}(\mathbf{x}, \mathbf{\dot{x}}, t)
 ++++
 
-_it's partial derivative with respect to_ latexmath:[t] _will be_
+_its partial derivative with respect to_ latexmath:[t] _will be_
 
 [latexmath]
 ++++
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -1161,6 +1161,28 @@ Both functions have the same arguments:
 
 _[Note that array variables will be serialized, so `nSeed` is only equal to `nKnowns` in the case of directional derivatives (resp., equal to `nUnknowns` in the case of adjoint derivatives) if all value references of `knowns` (resp., `unknowns`) point to scalar variables._
 _Likewise `nSensitivity` is only equal to `nUnknowns` (resp., `nKnowns`) if all value references of `unknowns` (resp., `knowns`) point to scalar variables.]_
+_[Note that it is the partial deriavative and not the total derivative that is consedered.
+ So when computing the partial derivative with respect to the independent variable latexmath:[t] of the function
+
+[latexmath]
+++++
+f(\textbf{x}, \textbf{\dot{v}}, t)
+++++
+
+it's partial derivative with respect to t will be
+
+[latexmath]
+++++
+\frac{\partial f}{\partial t}
+++++
+
+and not
+
+[latexmath]
+++++
+\frac{\partial f}{\partial t} + \frac{\partial f}{\partial \textbf{x}} \cdot \textbf{\dot{v}}+\frac{\partial f}{\partial \textbf{dot{x}}} \cdot \textbf{\ddot{v}}
+++++
+]_
 
 ===== Directional Derivatives [[directionDerivatives]]
 
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -1160,29 +1160,30 @@ Both functions have the same arguments:
 * `nSensitivity` contains the length of `sensitivity`.
 
 _[Note that array variables will be serialized, so `nSeed` is only equal to `nKnowns` in the case of directional derivatives (resp., equal to `nUnknowns` in the case of adjoint derivatives) if all value references of `knowns` (resp., `unknowns`) point to scalar variables._
-_Likewise `nSensitivity` is only equal to `nUnknowns` (resp., `nKnowns`) if all value references of `unknowns` (resp., `knowns`) point to scalar variables.]_
-_[Note that it is the partial deriavative and not the total derivative that is consedered.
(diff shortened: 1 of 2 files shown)


andreas-junghanns:
@KarlWernersson : I tried to improve the sentences, but I might have changed the meaning - please recheck!

KarlWernersson:
Thanks for the cleanup

KarlWernersson:
Do we also need some formula explaining in the getting partial derivative section or is this sufficient?

andreas-junghanns:
> Do we also need some formula explaining in the getting partial derivative section or is this sufficient?

Maybe we can add that to the non-normative text? Can you propose what should be there and I can try to make it pretty?

KarlWernersson:
I wrote a small clarification, I don't have the infrastructure to build the document locally so I might have gotten some of the formatting wrong

andreas-junghanns:
I have now changed your proposal to render like this:
grafik

Please check again.

KarlWernersson:
Thanks

f and t was intentionally not **bold** as they are scalars in this expression.
f could could be **bold** but then we need to remove the dot product

KarlWernersson:
But i realize we don't do this at other places at the standard so I'm not sure what the best solution...

andreas-junghanns:
> f and t was intentionally not **bold** as they are scalars in this expression. f could could be **bold** but then we need to remove the dot product

Sounds like I am missing the semantics of bold vs. non-bold. Please check also our mathematical notation: there we turn a lot of the math symbols to bold just to have the formatting consistent. Can you explain what bold/non-bold should mean?

KarlWernersson:
Also i screwed up v and x i will fix

KarlWernersson:
So normally non bold means scalar variable and bold means multidimensional variable . In general our equations works on multidimensional variables except for t which of course is a scalar but we have used bolded t in other parts of the code.

KarlWernersson:
I went back and looked at the 2.0.2 document and there we use non bold for t, Tnext etc.
I think it would be good to revert this, i can try to do this if you want?

masoud-najafi (changes requested):
This statement is not clear: "For continuous variables this corresponds to that there exist values for the knowns where the partial derivative of the unknwon with respect to a known is non-zero."

continuous -> continuous-time
"what is ment" -> ?

BTW, in DirectionalDerivative, the knowns are only "inputs and continuous-time states". Even if the independent variable is listed inside dependencies, I is meaningless to take into account, isn't it?

KarlWernersson:
> BTW, in DirectionalDerivative, the knowns are only "inputs and continuous-time states". Even if the independent variable is listed inside dependencies, I is meaningless to take into account, isn't it?

I think this is not the case knows are whatever we have listed in the dependencies, and that can include the time variable.

> vknown is the vector of [input](https://fmi-standard.org/docs/3.0-dev/index.html#input) variables of function f which can be changed by the importer in the current state. Details about which variables are in vknown are given in the description of element [dependencies](https://fmi-standard.org/docs/3.0-dev/index.html#dependencies) in [<ModelStructure>](https://fmi-standard.org/docs/3.0-dev/index.html#ModelStructure).


So if that is listed you can get directional derivative with respect to the time variables. So what I am trying to do is to clarify what happens when you take the directional derivative with respect to time. And that you don't get the total derivative.

Btw, the input links should not be there as this is not variables with variability "input"

KarlWernersson:
Another solution would be to not list time among the dependencies or forbid it from the directional derivatives but we have not done that so far

masoud-najafi:
If independent variable "time" is listed in the dependencies of a variable V, directionalDerivative may be provided the partial derivative of V WRT time.  The [only] use-case for DirectionalDerivative that I know is computing the Jacobian. In computing the Jacobian, we do not take into account "time", even if is listed.  So no contradiction for Jacobian. Do you any other use-case?

KarlWernersson:
I don't really see any but maybe I am missing something, I am just trying to fix what we have as we curently are allowing getDirecotinal derivative with respect to time but  I am fine removing the possibility to take directional derivative of time it but that a larger change.

KarlWernersson:
Anyhow an exporter an easily avoid providing directional derivatives with respect to time by not listing the time variable among the dependencies.

KarlWernersson:
From the fmi 2.0.2 standard

> . If a variable with causality = 
″independent″ is explicitely defined under ScalarVariables, a directional derivative with 
respect to this variable can be computed. If such a variable is not defined, the directional 
derivative with respect to the independent variable cannot be calculated.

In fmi3 we decided that independent must be specified. So this text has been removed.
So it is nothing new.

KarlWernersson:
But i dont knwo if anyone has implemented support for it, I haven't

masoud-najafi:
I do not see any harm in listing "time" in the dependencies. If an exporter lists "time" in the dependencies, it SHOULD also provide the directionalDerivative WRT "time" too. But it is up to the importer to use it or not. When I think, I remember I have seen the case where I have needed the derivative of a variable WRT time to estimate its sensitivity. In this case it was a zero-crossing surface that was also an output. I think clarification would be good, but not forbidding it, which is, as you said,  a big change.

masoud-najafi:
> From the fmi 2.0.2 standard
> 
> > . If a variable with causality =
> > ″independent″ is explicitely defined under ScalarVariables, a directional derivative with
> > respect to this variable can be computed. If such a variable is not defined, the directional
> > derivative with respect to the independent variable cannot be calculated.
> 
> In fmi3 we decided that independent must be specified. So this text has been removed. So it is nothing new.

This is still valid for fmi3, even if independent variable is always present.

docs/2_2_common_mechanisms.adoc
```diff
@@ -1160,7 +1160,30 @@ Both functions have the same arguments:
 * `nSensitivity` contains the length of `sensitivity`.
 
 _[Note that array variables will be serialized, so `nSeed` is only equal to `nKnowns` in the case of directional derivatives (resp., equal to `nUnknowns` in the case of adjoint derivatives) if all value references of `knowns` (resp., `unknowns`) point to scalar variables._
-_Likewise `nSensitivity` is only equal to `nUnknowns` (resp., `nKnowns`) if all value references of `unknowns` (resp., `knowns`) point to scalar variables.]_
+_Likewise `nSensitivity` is only equal to `nUnknowns` (resp., `nKnowns`) if all value references of `unknowns` (resp., `knowns`) point to scalar variables._
+
+_Note that it is the partial derivative and not the total derivative that is considered here._
+_When computing the partial derivative with respect to the independent variable_ latexmath:[t] _of the function_
```
klausschuch, line 1166: Isn't it always the partial derivative and not the total derivative?
Mentioning the independent variable explicitly here is confusing. It could be interpreted that just for the independent variable (i.e., time) it is the partial derivative otherwise it is the total derivative.

KarlWernersson:
Meeting to discuss the isses scheduled for 2022-03-09 15:00 CET

Maplesoft-fmigroup:
I believe that some stiff methods (e.g. Rosenbrock) require partial derivatives with respect to explicit occurrences of the time variable. This suggests that we do not want to disallow this for the directionalDerivative.

docs/2_4_common_schema.adoc
```diff
@@ -1685,7 +1685,8 @@ Elements <<Output>>, <<ContinuousStateDerivative>>, <<ClockedState>>, <<InitialU
 |`dependencies`
 |
 [[dependencies,`dependencies`]]
-Optional attribute defining the dependencies as list of value references of the unknown latexmath:[{\mathbf{v}_{\mathit{unknown}}}] (directly or indirectly via other variables) with respect to latexmath:[{\mathbf{v}_{\mathit{known}}}].
+Optional attribute defining the dependencies as list of value references of the unknown latexmath:[{\mathbf{v}_{\mathit{unknown}}}] directly with respect to latexmath:[{\mathbf{v}_{\mathit{known}}}].
+For continuous variables this corresponds to that there exist values for the knowns where the partial derivative of the unknwon with respect to a known is non-zero.
```
HansOlsson, line 1689:
```suggestion
For continuous variables this corresponds to that there exist values for the knowns where the partial derivative of the unknown with respect to a known is non-zero.
```
Spelling

docs/2_4_common_schema.adoc
```diff
@@ -1708,6 +1709,10 @@ Knowns latexmath:[{\mathbf{v}_{\mathit{known}}}] in <<InitializationMode>> (for
 
 * <<independent>> variable (usually time; <<causality>> = <<independent>>).
 
+_[For a dynamic system most of the unknowns would have an implict dependency on the independent variable.
+When the independent variable is listed among the dependencies this implicit dependency is not what is ment, but rather that the unknown's partial derivative with respect to the independent variable can be non-zero.
```
HansOlsson, line 1713:
```suggestion
When the independent variable is listed among the dependencies this implicit dependency is not what is meant, but rather that the unknown's partial derivative with respect to the independent variable can be non-zero.
```
Spelling

docs/2_2_common_mechanisms.adoc
```diff
@@ -1160,7 +1160,30 @@ Both functions have the same arguments:
...
+
+[latexmath]
+++++
+\mathbf{f}(\mathbf{x}, \mathbf{\dot{x}}, t)
+++++
+
+_it's partial derivative with respect to_ latexmath:[t] _will be_
```
Maplesoft-fmigroup, line 1173: it's -> its

docs/2_2_common_mechanisms.adoc
KarlWernersson: Yes its not only valid for time

KarlWernersson:
Meeting Conclusion.
Remove direct reference to time. Possible keep some comment like; it also holds for independent variable...

Dependency means that unknown has an algebraic dependency on the known.
[For real valued unknowns and knowns this means that the partial derivative on the unknown with respect to the known is not identical zero. ]

No mentioning of total derivative it adds confusion and that we speak of partial derivative is clear,

andreas-junghanns:
So who will update the PR?

masoud-najafi (approved):

KarlWernersson:
I will do it tomorrow I probably wont have time today

docs/2_2_common_mechanisms.adoc
KarlWernersson: thanks

PR #1722, opened 2022-03-22
Intermediate Update/ clock Update in table overview of features per interface type

MBlesken:
@andreas-junghanns following https://github.com/modelica/fmi-standard/pull/1585 I updated "Table 1. Non-normative overview of features per interface type.". What do you think?

### DIFF
--- a/docs/1___overview.adoc
+++ b/docs/1___overview.adoc
@@ -178,7 +178,7 @@ Inputs/Outputs: icon:times[]
 |icon:check[]
 |In <<EventMode>>: icon:check[] +
 Else: icon:times[]
-|icon:check[]
+|icon:times[]
 |====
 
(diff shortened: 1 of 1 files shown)


MBlesken:
@andreas-junghanns since in SE time isn't advancing within a single step, I guess "Direct Feedthrough" should be checked. Am I right?

KarlWernersson:
Is direct feedthrough the correct term?
If you co-simulate a gain block you still have direct feedthrough of the input to the output if you don't use some higher order input interpolation.

MBlesken:
@andreas-junghanns I reverted the Direct Feedthrough entry for SE. Can you approve and merge, please?

andreas-junghanns (approved):

Issue #1721, opened 2022-03-22
dependency of clocks


MBlesken:
In the Model Structure chapter for dependency it is stated: "_[This means that all dependencies must be listed, also the dependencies of clocked variables on their input clocks, even if redundant information.]_" **Is this what we want?**

**If yes**: In the SE example the variables 

- Float64 name="AOut" valueReference="2" causality="output" variability="discrete" clocks="5"
- Float64 name="BOut" valueReference="4" causality="output" variability="discrete" clocks="6"

should be listed redundantly in ModelStructure as well, **right?**

The dependency of 

- Clock name="OutputClock" valueReference="7" causality="output" intervalVariability="triggered" clocks="5"

is already listed in ModelStructure, as well: Why did we say that is necessary? **Did we note the reason for it in the specification?**

I think @masoud-najafi or @clagms can maybe help here, but any feedback is welcome. FYI: @andreas-junghanns @PTaeuberDS

KarlWernersson:
Yes, I don't remember why we need the clock information for clocked variables in the dependencies as it is already in the clock reference. But we have decided this several times. And I don't think we should change it so late.
The text should be 

> on their clocks
and not
> on their input clocks

as they could also be triggered by an output clock.

We also don't list clocks among the known variables which is not correct. they are known in event mode.

KarlWernersson:
I also think we should skip the 

> even if redundant information.

It just adds confusion and is kind of, a complaint of our decision.

Alternatively change it to something like

> this clock dependency should be the same as in the attribute clock of the variable

Is this always is true?
For clocked variables I think so, but it is not true for clocks

masoud-najafi:
>Is this always is true?
>For clocked variables I think so, but it is not true for clocks.
For clock variables we must also define their dependency, and it is really important to have dependency of output clocks on input clocks.

masoud-najafi:
I agree that "even if redundant information" is redundant itself. "must" in the statement . So we may skip it.

masoud-najafi:
> I also think we should skip the
> 
> > even if redundant information.
> 
> It just adds confusion and is kind of, a complaint of our decision.
> 
> Alternatively change it to something like
> 
> > this clock dependency should be the same as in the attribute clock of the variable
> 
> Is this always is true? For clocked variables I think so, but it is not true for clocks

[...]

PR #1727, opened 2022-03-25
add dependencies in SE xml

MBlesken:
As a first step initiated by #1721 I added dependency information w.r.t. the clocks in order to make the SE xml consistent.

### DIFF
--- a/docs/examples/scheduled_execution.xml
+++ b/docs/examples/scheduled_execution.xml
@@ -33,8 +33,8 @@
 		priority="3" intervalDecimal="0.05"/>
 	</ModelVariables>
 	<ModelStructure>
-		<Output valueReference="2" dependencies="0 1"/>
-		<Output valueReference="4" dependencies="3"/>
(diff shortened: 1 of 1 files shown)


andreas-junghanns (approved):

andreas-junghanns:
@nickbattle : Why did VDMCheck not find this inconsistency?
Is there a case where a clocked variable does not depend on its clock?

@MBlesken : will you merge this PR?

nickbattle:
> @nickbattle : Why did VDMCheck not find this inconsistency?

Because I'm not clear what the rules are. The formal model has the benefit of being unambiguous, but it may well be wrong until it is verified. Currently, clocked variables are considered as "optional" in the ModelStructure.

The current rules are [here](https://github.com/INTO-CPS-Association/FMI-VDM-Model/blob/61cd2ccd7d055dd45f7c1ff33d169fb5e076e386/fmi3/static-model/FMIModelDescription.vdmsl#L212)

PR #1731, opened 2022-03-30
Remove most changes in 1706, change some parts

KarlWernersson:
Please check the last part, non normative text, rewrote it a bit.
Even if we want to be as short as possible it could be good that there is no misunderstanding as we ourselves managed to misunderstand it some one new is much more likely to do it

so I think its good we have something there to be more explicit, but if you really don't agree I can remove it

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -1161,30 +1161,7 @@ Both functions have the same arguments:
 * `nSensitivity` contains the length of `sensitivity`.
 
 _[Note that array variables will be serialized, so `nSeed` is only equal to `nKnowns` in the case of directional derivatives (resp., equal to `nUnknowns` in the case of adjoint derivatives) if all value references of `knowns` (resp., `unknowns`) point to scalar variables._
-_Likewise `nSensitivity` is only equal to `nUnknowns` (resp., `nKnowns`) if all value references of `unknowns` (resp., `knowns`) point to scalar variables._
-
-_Note that it is the partial derivative and not the total derivative that is considered here._
-_When computing the partial derivative with respect to the independent variable_ latexmath:[t] _of the function_
-
-[latexmath]
-++++
-\mathbf{f}(\mathbf{x}, \mathbf{\dot{x}}, t)
-++++
-
-_its partial derivative with respect to_ latexmath:[t] _will be_
-
-[latexmath]
-++++
-\frac{\partial \mathbf{f}}{\partial t}
-++++
-
-_and not_
-
-[latexmath]
-++++
-\frac{\partial \mathbf{f}}{\partial t} + \frac{\partial \mathbf{f}}{\partial \mathbf{x}} \mathbf{\dot{x}}+\frac{\partial \mathbf{f}}{\partial \mathbf{\dot{x}}} \mathbf{\ddot{x}}
-++++
-_]_
+_Likewise `nSensitivity` is only equal to `nUnknowns` (resp., `nKnowns`) if all value references of `unknowns` (resp., `knowns`) point to scalar variables.]_
 
 ===== Directional Derivatives [[directionDerivatives]]
 
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -1687,9 +1687,9 @@ Elements <<Output>>, <<ContinuousStateDerivative>>, <<ClockedState>>, <<InitialU
 |
 [[dependencies,`dependencies`]]
 Optional attribute defining the algebraic dependencies as list of value references of the unknown latexmath:[{\mathbf{v}_{\mathit{unknown}}}] directly with respect to latexmath:[{\mathbf{v}_{\mathit{known}}}].
-For real valued unknowns and knowns this means that the partial derivative on the unknown with respect to the known is not identically zero.
-If not present, it must be assumed that the unknown depends on all knowns.
-If present as empty list, the unknown depends on none of the knowns.
+For a real valued unknown and a real valued known, if the known is not listed among the dependencies then the partial derivative of the unknown with respect to that known is identically zero.
+If dependencies is not present, it must be assumed that the unknown depends on all knowns.
+If dependencies is present as empty list, the unknown depends on none of the knowns.
 Otherwise the unknown depends on the knowns defined by the given value references. +
 _[This means that all dependencies must be listed, also the dependencies of clocked variables on their clocks.]_ +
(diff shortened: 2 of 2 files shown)


andreas-junghanns (approved):
Language seems ok to me, content needs to be approved by the other reviewers.

docs/2_4_common_schema.adoc
```diff
@@ -1710,9 +1710,11 @@ Knowns latexmath:[{\mathbf{v}_{\mathit{known}}}] in <<InitializationMode>> (for
 
 * <<independent>> variable (usually time; <<causality>> = <<independent>>).
 
-_[For a dynamic system most of the unknowns would have an implict dependency on the independent variable.
-When the independent variable is listed among the dependencies this implicit dependency is not what is meant, but rather that the unknown's partial derivative with respect to the independent variable can be non-zero.
-For a mechanical system this is called a rheonomous system, while if the independent variable is not listed in the dependencies, it is a scleronomous system.]_
+_[The <<independent>> variable should be treated as any other variable when it comes to dependencies._
+_As only explicit dependencies are listed it is required that unknowns that depends on the <<independent>> variable have a partial derivative that is different from zero._
```
chrbertsch, line 1714: depends --> depend

"... have a partial derivative that is different from zero." 
"... have a partial derivative w.r.t this variable that is different from zero." 

"... have a partial derivative that is different from zero." : this is not true for all times, but the variable _can_ be nonzero.

docs/2_2_common_mechanisms.adoc
```diff
@@ -1161,30 +1161,7 @@ Both functions have the same arguments:
...
-\frac{\partial \mathbf{f}}{\partial t}
-++++
-
-_and not_
-
-[latexmath]
-++++
-\frac{\partial \mathbf{f}}{\partial t} + \frac{\partial \mathbf{f}}{\partial \mathbf{x}} \mathbf{\dot{x}}+\frac{\partial \mathbf{f}}{\partial \mathbf{\dot{x}}} \mathbf{\ddot{x}}
-++++
-_]_
```
chrbertsch, line 1187: Why do we remove the explanation that the partial and not the total derivative w.r.t time is meant? Is it obvious or explained somewhere else?

KarlWernersson: This was requested by @klausschuch and @masoud-najafi  as they thought it was more confusing that we added this section than leaving it out

KarlWernersson:
Thanks @andreas-junghanns , unfortunately your fixes changed the meaning.
Identical zero = zero always and everywhere, while zero is just zero at some point, which they always can be.
I tried to use more general terms with more word to describe this but it was decided to use identical zero as it is more clear and less confusing

docs/2_2_common_mechanisms.adoc
KarlWernersson: But if there are different opinions on this I can return the section again, I added it before because I thought it helped but accepted the removal since others thought it was more confusing

andreas-junghanns:
@KarlWernersson  - thank for fixing my fix - I did not know that term and throught this was bad grammar.
Can @klausschuch also approve so we can merge this PR?

docs/2_4_common_schema.adoc
```diff
@@ -1687,7 +1687,7 @@ Elements <<Output>>, <<ContinuousStateDerivative>>, <<ClockedState>>, <<InitialU
 |
 [[dependencies,`dependencies`]]
 Optional attribute defining the algebraic dependencies as list of value references of the unknown latexmath:[{\mathbf{v}_{\mathit{unknown}}}] directly with respect to latexmath:[{\mathbf{v}_{\mathit{known}}}].
-For real valued unknowns and knowns this means that the partial derivative on the unknown with respect to the known is not identical to zero.
+For real valued unknowns and knowns this means that the partial derivative on the unknown with respect to the known is not identically zero.
```
klausschuch, line 1690: For a real valued unknown and and a real valued known this means that the partial derivative of the unknown with respect to the known is not identically zero.

docs/2_4_common_schema.adoc
```diff
@@ -1711,7 +1711,7 @@ Knowns latexmath:[{\mathbf{v}_{\mathit{known}}}] in <<InitializationMode>> (for
 * <<independent>> variable (usually time; <<causality>> = <<independent>>).
 
 _[The <<independent>> variable should be treated as any other variable when it comes to dependencies._
-_As only explicit dependencies are listed it is required that unknowns that depends on the <<independent>> variable have a partial derivative that is different from zero._
+_As only explicit dependencies are listed it is required that unknowns that depend on the <<independent>> variable have a partial derivative w.r.t this variable that is different from identically zero._
```
klausschuch, line 1714: # #`_As only explicit dependencies are listed, it is required that unknowns that depend on the <<independent>> variable have a partial derivative w.r.t. this variable that is different from identically zero._`

docs/2_4_common_schema.adoc
chrbertsch: "on the unknown" --> of the unknown

I think for a given simulation we cannot guarantee that such a partial derivative is not constant zero; the presence of unknows and knowns in the above ist only means that _structurally_ it could be nonzero.

KarlWernersson: @chrbertsch That is true, what we should say is that not listing a known implies that is has a partial derivative with respect to this known that is identically zero.

I will try to reformulate

chrbertsch: Does an FMU have to expose all partial derivatives if it sets the corresponding capability flag? Or can it be that it just exposes some, and the others are not exposed but could be nonzero?

KarlWernersson: If an FMU supports getDirectionalDirectional deriavtives, it must be able to compute them for all continuous outputs and all continuousStateDerivatives if they are exposed. An ME FMU must expose all continuousStateDerivatives, while a CS FMU is allowed to hide them. All outputs must be listed in ModelStructure and then their partial derivatives of the continuous real ones should be able to be computed if partial derivatives is supported.
It change an output to a local variable then there is no guarantee that partial derivative functions will work etc.

KarlWernersson:
new try

andreas-junghanns (approved):

chrbertsch (approved):

andreas-junghanns:
@klausschuch : do you approve? Please do so until end of business today. We want to tag RC.1 and this is the last issue open here.

klausschuch:
@andreas-junghanns sorry, i read this comment here a bit too late. I agree with the content of this merged pull-request.
(However, IMHO there are some minor language glitches that we should fix for the final release.)

chrbertsch:
@klausschuch could you provide a quick fix for the language glitches (that we perhaps even could merge before tagging RC1)  or create a ticket so that we do not forget

PR #1735, opened 2022-04-04
Discourage static linking

t-sommer:
on platforms that support shared libraries

see also https://github.com/modelica/Reference-FMUs/issues/215


andreas-junghanns (approved):
Maybe better in the IG?

t-sommer:
I've seen this done wrong quite often, so we should have it in a prominent place.

t-sommer:
...and I have to admit that examples like [FMU-Standalone-App](https://github.com/CATIA-Systems/FMU-Standalone-App) encourage people to do it wrong.

PR #1746, opened 2022-04-15
Remove model structure examples due to move to IG

pmai:
Removed the examples, since they are being moved to the IG (see modelica/fmi-guides#36).


andreas-junghanns (approved):
Maybe, just maybe, we can have a reference to the IG here to point people to these examples?

pmai:
sure could do that, however do we have an official reference already for the IG that we can place in the standard?

chrbertsch:
> sure could do that, however do we have an official reference already for the IG that we can place in the standard?

Not sure if I understood you question well, but we a have reference to the IG in section 1.4 https://fmi-standard.org/docs/3.0-dev/#_how_to_read_this_document

andreas-junghanns:
> sure could do that, however do we have an official reference already for the IG that we can place in the standard?

Hyperlinks are overrated: We could just say: "Please see the FMI Implementer´s Guide!" - without link, google will guide the interested.

TorstenBlochwitz (approved):
OK, but "Implementor's Guide through the Universe" should be published as soon as possible!

PR #1757, opened 2022-04-25
Fix contrib

chrbertsch:
resolves #1758  (tools participating in XC): Please check

Fix Comments by @MBlesken  in https://github.com/modelica/fmi-standard/issues/1737#issuecomment-1108361917 .
@MBlesken : "Maybe separate country followed by next name by more than just a blank": I do not find this. Wasn't this fixed already? Or could you please create a PR?
Company names are still not uniform (with or w/o "GmbH")

Issue #1745, opened 2022-04-14
Move layered Standards to a different Github repo


chrbertsch:
Layered standards such as 
https://github.com/modelica/fmi-standard/blob/master/docs/X_layered_XCP.adoc or https://github.com/modelica/fmi-standard/pull/1177
should not be bundled with the main FMI Standard document.

Reason: they have other release cycles and  approval processes - and the FMI standard itself  do not depend on them

Currently, e.g. the "assets" zip file of a release contains also the XCP layered standard (see https://github.com/modelica/fmi-standard/releases)

andreas-junghanns:
Should we have a single repo with all layered standards currently proposed?

chrbertsch:
> Should we have a single repo with all layered standards currently proposed?

As we will have different releases, perhaps its better to have a separate repos?
@pmai , @t-sommer : your opinion?

pmai:
I would also move to separate repos, i.e. one for each layered standard, e.g. fmi-ls-xcp fmi-ls-bus or whatever.

t-sommer:
So can we just remove it for now and create the repo later?

t-sommer:
Wouldn't this make sense for Ports & Icons, too?

andreas-junghanns:
> So can we just remove it for now and create the repo later?

... sure, as long as the text does not get lost...

t-sommer:
And maybe also the buildDescription.xml, so we can reuse it in FMI 2.0.

andreas-junghanns:
> And maybe also the buildDescription.xml, so we can reuse it in FMI 2.0.

We can reuse the description without removing it from the 3.0 standard.
This was discussed at the time and we decided against it.

PR #1760, opened 2022-04-26
Bump version to 3.0-rc.2

t-sommer:


chrbertsch:
Shall we merge #1757 first? So at least these corrections would be included.

chrbertsch (changes requested):
We should exclude X_layered_XCP.html from the distribution archive, see #1745

andreas-junghanns (approved):

PR #1762, opened 2022-04-26
Add notes to layered standards

t-sommer:


chrbertsch (changes requested):
Didn't we decide against making the BuildConfiguration and the Terminals layered standards?
See https://github.com/modelica/fmi-standard/issues/1745#issuecomment-1109708997

PR #1765, opened 2022-05-09
Change version from 3.0-rc.2 to 3.0

andreas-junghanns:


andreas-junghanns:
All: If we want to publish 3.0 today, we should get this PR approved by more than my eyes... please try to make the time soon today.

chrbertsch (commented):
Are the two "dots" in the change in the xsd files  correct?  <xs:pattern value="3[.](.*)"/>
(I am no expert in XML)

schema/fmi3BuildDescription.xsd
```diff
@@ -108,7 +108,7 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
             <xs:attribute name="fmiVersion" use="required">
                 <xs:simpleType>
                     <xs:restriction base="xs:normalizedString">
-                        <xs:pattern value="3[.](0-.*)"/>
+                        <xs:pattern value="3[.](.*)"/>
```
pmai, line 111:
```suggestion
                        <xs:pattern value="3[.](0|[1-9][0-9]*)([.](0|[1-9][0-9]*))?(-.+)?"/>
```

pmai: @andreas-junghanns This should cover the parts of semantic versioning we care about, while still allowing just "3.0", without any build-variants, which are not applicable... Still have to test this somewhat, though...

andreas-junghanns:
Also added to fmi3TerminalsAndIcons.xsd.

schema/fmi3BuildDescription.xsd
t-sommer: Shouldn't e.g. `3.0.1-beta.1` also be a valid version?

pmai: It is, or do I miss something?

andreas-junghanns:
> Shouldn't the modeDescription schema for fmiVersion also be changed accordingly?

Please check the proposal...

schema/fmi3BuildDescription.xsd
andreas-junghanns: > It is, or do I miss something?

We also allow now `3.0-gamma42Beta` - right?

chrbertsch (approved):

t-sommer:
I ran a quick test with FMPy.
The following versions match the `fmiVersion` regex: `3.0`, `3.1`, `3.0.1`, `3.1-alpha.2`, `3.0.1-rc.1`
These don't: `3.x.1-rc.1`, `31`

t-sommer (approved):

PR #1773, opened 2022-05-17
Back-port buildDescription.xml to FMI 2.0

t-sommer:
fixes #1769

PR #1772, opened 2022-05-16
Use FMI 3.0 platform tuples

t-sommer:
for platforms other than {darwin|linux|win}{32|64}

fixes #1764

PR #1783, opened 2022-06-07
backport #1777, #1773, #1772 to the Word document

chrbertsch:
backport #1777, #1773, #1772 to the Word document


chrbertsch:
@t-sommer : can we merge this? It only makes the word document consistent with the already merged changes.
As merging .docx documents is a mess, I would like to keep the .docx in sync with the asciidoc version, and we alredy merged #1785 which also has to be fixed in the word document

PR #1785, opened 2022-06-28
Fix description of forbidden special values for real values in FMI 2.0.x

### LINKED ISSUE #1784: Correct "+INF" to "INF" in FMI 2.0.3  2.2
in the sentence "The special values NAN, +INF, -INF for variables values are not allowed in the FMI xml files"
According to https://www.oreilly.com/library/view/xml-schema/0596002521/re63.html "+INF" is not correct XML.

The question came up, as there was an FMU with an "INF" value

t-sommer:
back-ports #1122 to FMI 2.0
fixes #1784

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -82,6 +82,8 @@ note the FMI calling interface requires that strings are encoded in UTF-8.
 Since the XML files are also required to be encoded in UTF-8,
 string variables need not to be transformed when reading from the XML files in to C string variables.]._
 
+The special values NAN, INF, +INF, and -INF for variables values are not allowed in the FMI XML files.
+
 _[Note that child information items,
 such as "elements" in a "sequence" are [underline]#ordered lists# according to document order,
 whereas attribute information items are [underline]#unordered sets# (see http://www.w3.org/TR/XML-infoset/#infoitem.element).
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -82,8 +82,6 @@ note the FMI calling interface requires that strings are encoded in UTF-8.
 Since the XML files are also required to be encoded in UTF-8,
 string variables need not to be transformed when reading from the XML files in to C string variables.]._
 
-The special values NAN, +INF, -INF for variables values are not allowed in the FMI XML files.
-
 _[Note that child information items,
 such as "elements" in a "sequence" are [underline]#ordered lists# according to document order,
 whereas attribute information items are [underline]#unordered sets# (see http://www.w3.org/TR/XML-infoset/#infoitem.element).


chrbertsch (changes requested):
I think as we forbid "INF", "-INF" "NAN" in FMI 2.0.3, we should not allow these values in FMI 2.0.4., as we otherwise confuse users whether an FMU e.g. with an "INF" parameter in the XML is a "legal" FMU or not (it would be a valid FMI 2.0.4 FMU but an invalid FMI 2.0.3 FMU). There are importing tools out in the wild, that to not treat "INF" correctly.

My point in #1784 was only to rename "+INF" to "INF" in FMI 2.0.4, as this is how it is called in XML (and which confused users)

(I now also understand why this restriction was removed 3.0 following the discussion in #644 )

andreas-junghanns (approved):

andreas-junghanns:
> > There are importing tools out in the wild, that to not treat "INF" correctly.
> 
> If a tool cannot handle NANs (or any other value for that matter), it must raise an error. Allowing NANs in the XML will not change anything for the tool or the user (except that they cannot blame it on the FMU anymore).

A tool must be able to assume that an FMU follows the standard. FMI 2.0 is clear on not allowing these and the reasons where clear: too many different ways of actually writing the special numbers in the different standard libs writing floats and parsing them back. The "blame" is an important issue here: All tools have to follow the standard, also exporters!

andreas-junghanns:
> Using non-finite numbers in the API was always legal and must be handled by the FMU.

FMI 2.0.3 restricts that, if I may remind you by stating: "The special values NAN, +INF, -INF for variables values are not allowed in the FMI xml files." With that, an importer's behavior is undefined if it finds such values.

docs/2_2_common_schema.adoc
```diff
@@ -82,8 +82,6 @@ note the FMI calling interface requires that strings are encoded in UTF-8.
 Since the XML files are also required to be encoded in UTF-8,
 string variables need not to be transformed when reading from the XML files in to C string variables.]._
```
pmai, line 84:
```suggestion

The special values NAN, INF, +INF, and -INF for variables values are not allowed in the FMI XML files.

```

pmai:
> > Using non-finite numbers in the API was always legal and must be handled by the FMU.
> 
> FMI 2.0.3 restricts that, if I may remind you by stating: "The special values NAN, +INF, -INF for variables values are not allowed in the FMI xml files." With that, an importer's behavior is undefined if it finds such values.

But as @t-sommer rightly points out, that only applies to the XML, not to the API.

I've provided my suggestion of how the wording should change in 2.0.x, to clarify that obviously INF without the redundant + also falls under the prohibition (this always was the intent, and since the language talks about values and not lexical items, +INF and INF are the same value anyways, so this technically also does not expand the prohibition).

It is right and proper that 3.0 drops this whole nonsense sharade, but for 2.0.x we cannot retroactively drop it.

chrbertsch (approved):

Issue #1778, opened 2022-05-25
Include Backports in FMI 2.0.4 Word document


chrbertsch:
#1777, #1773, #1772, ports and icons (PR to follow)

chrbertsch:
Attach PDFs
[FunctionalMockupInterface_2_0_4RC1.pdf](https://github.com/modelica/fmi-standard/files/9180441/FunctionalMockupInterface_2_0_4RC1.pdf)
[FunctionalMockupInterface_2_0_4RC1_ChangeMarks.pdf](https://github.com/modelica/fmi-standard/files/9180442/FunctionalMockupInterface_2_0_4RC1_ChangeMarks.pdf)
@t-sommer:  could you please create create the "complete package" file and attach the PDFs and complete the prerelease page? Thanks!

PR #1791, opened 2022-07-19
Fix pseudocode example according to Karl's suggestion

### LINKED ISSUE #1629: Model Exchange Pseudocode Example: Possible issue with comment about setting initial inputs
In the Model Exchange Psuedocode Example in Section 3.2.4 of the FMI-Specification-2.0.3 near the beginning are the comments:

// set all variable start values (of "ScalarVariable / <type> / start") and
// set the input values at time = Tstart

However in Figure 6. in Section 3.2.3 it has that fmi2SetSEA with the index information:
SEA is one of Real, Integer, Boolean, String
for a variable with variability != constant
that has initial = "exact" or "approx"

The issue is variables with causality="input" have no initial, this means that input values should not be set at the location mentioned in the psuedocode.

I suggest that the comment:
// set the input values at time = Tstart

be moved into between the lines:
M_fmi2EnterInitializationMode(m)
M_fmi2ExitInitializationMode(m)

chrbertsch:
Fixes #1629 

Has to be backported to Word --> #1778 

@GarronFish:  please check

### DIFF
--- a/docs/3_2_model_exchange_api.adoc
+++ b/docs/3_2_model_exchange_api.adoc
@@ -384,14 +384,17 @@ dt = 0.01        // fixed step size of 10 milli-seconds
 // set the start time
 time = TStart
 
-// set all variable start values (of "ScalarVariable / <type> / start") and
-// set the start values at time = Tstart
+// set variable start values (of "ScalarVariable / <type> / start")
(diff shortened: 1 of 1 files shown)


GarronFish (commented):
Looks correct to me.  Thanks

t-sommer (approved):

PR #1809, opened 2022-09-11
Explain matching of terminals with structured naming convention arrays in 2.0.4

chrbertsch:
As it is proposed to explain the matching in more detail in FMI 3.0 (https://github.com/modelica/fmi-standard/pull/1808) and the IG, this consist mainly of a reference to these two places to avoid redundancies.
Please comment or create an alternative PR if you have a better suggestion.

### DIFF
--- a/docs/2_3_fmu_distribution.adoc
+++ b/docs/2_3_fmu_distribution.adoc
@@ -208,3 +208,6 @@ _[New in FMI 2.0.4:]_
 ==== Terminals and Icons
 
 The optional directory `terminalsAndIcons` contains the definition of the Ports and Icons of the FMU as defined in the FMI 3.0 Specification, 2.4.9. Terminals and Icons.
+
+Please note that "matching" terminal variables includes matching of scalarized variables "based on structured naming convention" of with FMI 3.0 array variables.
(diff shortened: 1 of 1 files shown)


docs/2_3_fmu_distribution.adoc
```diff
@@ -208,3 +208,6 @@ _[New in FMI 2.0.4:]_
 ==== Terminals and Icons
 
 The optional directory `terminalsAndIcons` contains the definition of the Ports and Icons of the FMU as defined in the FMI 3.0 Specification, 2.4.9. Terminals and Icons.
+
```
t-sommer, line 211: This clarification doesn't help much.

t-sommer (changes requested):

docs/2_3_fmu_distribution.adoc
chrbertsch: As we will explain this in 3.0.1 and the Implementer's Guide: Do we need an additional explaination for FMI 2.0.4 at all or can we close this PR?

CSchulzeTLK:
In fact this remark is useless, once the maintenance release of FMI 3 is done.

We decided to update FMI 3. I think we don't need this PR (hoping that we are actually going to release 3.0.1).

chrbertsch:
Closing, as there seems to agreement now to clarify this in FMI 3.0.1 (PR #1808)
If someone has a different opinion, please (re-)open a PR.

Issue #1822, opened 2022-10-26
Relax content description of /binaries folder


andreas-junghanns:
Current 3.0 wording in 2.5.1. Structure of the ZIP file suggests that only binaries are allowed in /binaries - we should allow other runtime-relevant files inside their platform specific /binaries folder.

chrbertsch:
The content of the a specific binaries/platform folder should be platform specific.

chrbertsch:
Design Webmeeting: 

Irina: what is the difference between the /binaries folder and the /resources folder? Where does an exporter put what?
Andreas: certain DLLs cannot be in the resource folder. OS specific load routines have acess specific files during runtime.
Irina: for DLLs it is already clear.
Pierre: For DLLs it is clear. Everything that is needed during runtime and is under control of the FMU can be put in the resources 
Andreas: ... for A2L we have several copies
Pierre: no use case for using it during load time
Markus: Proposal: everything in /binaries must be available before your load the dll and first call to the FMU, in the resources folder only after instanciation. What is inside this directories is up to the FMU.
Timo: for binaries one makes assumption on OS "magic", for resources one can use the filepath.
Andreas: for binaries, one has to unpack only the specific subdirectory. For resources everything has to be unpacked.
Irina: This seems to be a workaround for not having platform specific resources folder.
Andreas: Layered standards could provide more information for platform-specific content of resources files.
Andreas: fix language at the two places 
Pierre: let's go with Markus' proposal, clarify processing expectations.

PR #1832, opened 2022-11-18
Improved content description of binaries and resources dirs (#1822)

friedrichatgc:
Clearly define which files should go the binaries and resources dir of the FMU which when these dirs must be unpacked by the FMU importer.

### DIFF
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -90,7 +90,7 @@ The header files `fmi3PlatformTypes.h`, `fmi3FunctionTypes.h` and `fmi3Functions
 A binary FMU must contain, for each supported platform, the shared library `<modelIdentifier>.{dll|dylib|so}` and all files which are needed to load this shared library
 in the corresponding platform-specific binary folder.
 To use the binaries of a specific platform, all items in that platform-specific binary folder must be unpacked at the same location as the binary `<modelIdentifier>.{dll|dylib|so}`.
-The FMU importer must unpack these files prior loading the shared library `<modelIdentifier>.{dll|dylib|so}` and these files must stay there until unloading the shared library is finished.
+The FMU importer must unpack these files prior to loading the shared library `<modelIdentifier>.{dll|dylib|so}` and these files must stay there until unloading the shared library is finished.
 
 ====== Platform Tuple Definition [[platform-tupe-definition]]
 
@@ -196,7 +196,7 @@ The requirements and the expected processing must be documented in the `document
 In the optional directory `resources`, additional data can be provided in FMU specific formats, typically for tables and maps used in the FMU.
 More folders can be added under `resources` (tool/model specific).
 For the FMU to access these resource files, the resources directory shall be available in extracted form and the absolute path to this directory is provided via argument <<resourcePath>> of <<fmi3Instantiate>>.
-The `resources` directory must be available for the lifetime of the FMU instance, prior calling <<fmi3Instantiate>> and until <<fmi3FreeInstance>> has finished.
+The `resources` directory must be available for the lifetime of the FMU instance, from before calling <<fmi3Instantiate>> and until after <<fmi3FreeInstance>> has finished.
 The FMU instance must not write to or delete the directory or parts of it.
 
 ===== Directory `extra` [[extra-directory]]
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -50,13 +50,16 @@ sources                             // directory containing the C sources (optio
 binaries                            // directory containing the binaries (optional)
    x86_64-windows                   // binaries for Windows on Intel 64-bit
       <modelIdentifier>.dll         // shared library of the FMI implementation
-      <other DLLs>                  // the DLL can include other DLLs
+      <other files>                 // the shared library may depend on other files
    x86-linux                        // binaries for Linux on Intel 32-bit
       <modelIdentifier>.so          // shared library of the FMI implementation
+      <other files>                 // the shared library may depend on other files
    aarch32-linux                    // binaries for Linux on ARM 32-bit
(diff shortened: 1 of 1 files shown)


andreas-junghanns (approved):

docs/2_5_fmu_distribution.adoc
```diff
@@ -84,19 +87,20 @@ The header files `fmi3PlatformTypes.h`, `fmi3FunctionTypes.h` and `fmi3Functions
 
 ===== Directory `binaries` [[binaries-directory]]
 
-A binary FMU must contain the binary files for all supported platforms in this folder.
+A binary FMU must contain, for each supported platform, the shared library `<modelIdentifier>.{dll|dylib|so}` and all files which are needed to load this shared library
```
t-sommer, line 90: An FMU may  also ship with static libraries.

t-sommer (changes requested):

chrbertsch:
Webmeeting: @t-sommer : I will come up with proposals to simplify the wording. Could also be backported to FMI 2.0.x

docs/2_5_fmu_distribution.adoc
```diff
@@ -87,11 +88,23 @@ The header files `fmi3PlatformTypes.h`, `fmi3FunctionTypes.h` and `fmi3Functions
...
 The FMU importer must unpack these files prior to loading the shared library `<modelIdentifier>.{dll|dylib|so}` and these files must stay there until unloading the shared library is finished.
 
+====== FMUs with static libraries [[binaries-directory-static]]
+
+A static library binary FMU must contain, for each supported platform and ABI, the static library `<modelIdentifier>.{a|lib}` and all files which are needed to link with this static library
+in the corresponding platform-specific binary folder.
+These binary files may go into an ABI specific subdirectory, see <<platform-tupe-definition>> also for linking instructions.
+For static libraries the `binaries` directory can only contain files needed during link time since, in contrast to shared libraries, it is not possible to determine the file location
+of the static library at load time.
+Hence, the FMU importer is not needed to unpack these files during execution of the FMU.
```
pmai, line 106:
```suggestion
Hence, the FMU importer is not required to unpack these files during execution of the FMU.
```

pmai (commented):
Content-wise seems the right thing to me, tightened wording would be OK by me.

docs/2_5_fmu_distribution.adoc
```diff
@@ -84,19 +88,32 @@ The header files `fmi3PlatformTypes.h`, `fmi3FunctionTypes.h` and `fmi3Functions
 
 ===== Directory `binaries` [[binaries-directory]]
 
-A binary FMU must contain the binary files for all supported platforms in this folder.
+====== FMUs with shared libraries [[binaries-directory-shared]]
+
+A shared library binary FMU must contain, for each supported platform, the shared library `<modelIdentifier>.{dll|dylib|so}` and all files which are needed to load this shared library
+in the corresponding platform-specific binary folder.
 To use the binaries of a specific platform, all items in that platform-specific binary folder must be unpacked at the same location as the binary `<modelIdentifier>.{dll|dylib|so}`.
+This is needed since the shared library may search for other files, needed during library load time, relative to its own file location.
```
t-sommer, line 96: This reasoning should not be normative.

docs/2_5_fmu_distribution.adoc
```diff
@@ -84,19 +88,32 @@ The header files `fmi3PlatformTypes.h`, `fmi3FunctionTypes.h` and `fmi3Functions
...
+The FMU importer must unpack these files prior to loading the shared library `<modelIdentifier>.{dll|dylib|so}` and these files must stay there until unloading the shared library is finished.
+
+====== FMUs with static libraries [[binaries-directory-static]]
+
+A static library binary FMU must contain, for each supported platform and ABI, the static library `<modelIdentifier>.{a|lib}` and all files which are needed to link with this static library
```
t-sommer, line 101: Requiring "all files which are needed to link with this static library" to be contained in the FMU would render any FMU illegal that has external dependencies (caused by the static library).

docs/2_5_fmu_distribution.adoc
```diff
@@ -84,19 +88,32 @@ The header files `fmi3PlatformTypes.h`, `fmi3FunctionTypes.h` and `fmi3Functions
...
+in the corresponding platform-specific binary folder.
+These binary files may go into an ABI specific subdirectory, see <<platform-tupe-definition>> also for linking instructions.
```
t-sommer, line 103: This folder must not be a subdirectory of the platform binary directory but the `binaries` directory. Also, we should not allow spreading libraries and dependencies across multiple folders.

t-sommer (changes requested):

docs/2_5_fmu_distribution.adoc
t-sommer: IMHO, we should have a concise set of rules for the naming and availability at runtime in the spec and move all the reasoning and explanations to the implementers guide.

andreas-junghanns:
All: Markus dropped off our planet. Any change we want to be made, we must now make. @t-sommer : will you take over? You seem most interested...

chrbertsch:
FMI Design Meeting: 
Pierre: the original intent was to clarify that in binaries only things are put that are needed at link or load time.
For things needed at runtime, they have to be put in resources
Torsten: Give a rational that the FMU may depend on files that must be present during load time. 
Pierre: Link time is different for statically or dynamically linked library.
Pierre: There is a whole section in the IG! The reason for this PR was, that people were unclear that platform specific files can be in resources.
Pierre: Similar questions currently arise in the XCP LS, so a clarification is welcome. E.g. would not expect A2L files in resources or binaries, as they are not needed during runtime. Only for specific use cases the A2L file would be needed during runtime ...

Christian: I will create a new PR

chrbertsch:
Superseded by #1849

PR #1836, opened 2022-11-23
Define default variability

### LINKED ISSUE #1835: Default variability is missing for non-float non-clock non-clocked variables
The variability _attribute_ of the <fmi3Variable> elements is optional (see section 2.4.7.4).
When not present, table 17 specifies a default value for:
- variables of type `Float32` and `Float64`
- variables of type `Clock`
- clocked variables
In all other cases the default value is not defined.

chrbertsch:
resolves #1835

by deifing default variabiliy as "discrete" for varaibles of type other than float32 or float64

### DIFF
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -999,6 +999,7 @@ During <<IntermediateUpdateMode>>, <<discrete>> variables are not allowed to cha
 
 The <<variability>> of <<Clock,Clocks>> must be <<discrete>>.
 The <<variability>> of <<clocked-variable,`clocked variables`>> must be <<discrete>> or <<tunable>>.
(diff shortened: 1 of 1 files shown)


chrbertsch:
@antvl , @pmai  and others: could you please review this, so that we can merge it?

andreas-junghanns (approved):
Looks ok, not sure it is complete.

pmai (approved):
Makes sense to me

chrbertsch:
Design Webmeeting: 

Andreas: is this complete?
Pierre: It seemed to me to be complete. It covers all of he causalities. Clocks are not relevant here. 
Pierre: I checked it with table 18. This differs from the default value in FMI 2.0, where continuous did not make sense in many cases.
Andreas: OK

Issue #1839, opened 2022-11-30
Clarification on 'nominal' attribute for fmi3Float64 arrays


Maplesoft-fmigroup:
It is not clear if the 'nominal' attribute is allowed in the specification of fmi3Float64 type arrays. If present in the modelDescription.xml, how should it be treated? It is defined as a 'double' so does that mean a single value should be applied to all array elements? This could be problematic...

chrbertsch:
Design Meeting: 

Andreas: why is this different from other start values?
Kaska: it is a single value
Andreas: Elements of an array should be similar. Otherwise should they be in an array?
Pierre: otherwise just not provide a nominal.
Kaska: we should an explanation. This would be good enough for now.
Andreas: Non-Backward compatible changes should be in a different issue.
Pierre: Nominal values have to be treated with suspicion anyways.

PR #1841, opened 2022-12-07
Clarification of 'nominal' for array variables

Maplesoft-fmigroup:
Addresses issue #1839


andreas-junghanns (approved):
@Maplesoft-fmigroup : Please check also my latest changes before merging.

PR #1845, opened 2022-12-14
Clarify earlyReturn from fmi3DoStep

### LINKED ISSUE #1844: Return status when earlyReturn=fmi3True
What status value should be returned by fmi3DoStep when earlyReturn=fmi3True, is fmi3OK still acceptable? What about when terminateSimulation is also set to fmi3True?

Maplesoft-fmigroup:
Added a comment that this should not be used if the step terminated early due to an error to resolve #1844

### DIFF
--- a/docs/4_2_co-simulation_api.adoc
+++ b/docs/4_2_co-simulation_api.adoc
@@ -116,7 +116,8 @@ Reasons for <<earlyReturn, `earlyReturn = fmi3True`>> are
 ** the FMU's reaction to the importers request for <<earlyReturn>> by the return argument <<earlyReturnRequested, `earlyReturnRequested = fmi3True`>> of <<fmi3IntermediateUpdateCallback>>, or
 ** the FMU's request for a communication point at <<lastSuccessfulTime>> for any other reason.
 + 
-<<earlyReturn,`earlyReturn = fmi3True`>> must not be used to signal that the requested step was not completed due to an error (e.g. integrator encountered division by zero).
+<<earlyReturn,`earlyReturn = fmi3True`>> must not be used to signal that the requested step was not completed due to an error _[(e.g. integrator encountered division by zero)]_.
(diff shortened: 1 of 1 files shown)


andreas-junghanns (approved):
@pmai : Intend and wording ok?

docs/4_2_co-simulation_api.adoc
```diff
@@ -115,6 +115,9 @@ Reasons for <<earlyReturn, `earlyReturn = fmi3True`>> are
 ** expressed by the return arguments <<eventHandlingNeeded>> or <<terminateSimulationDS>> of <<fmi3DoStep>>, or
 ** the FMU's reaction to the importers request for <<earlyReturn>> by the return argument <<earlyReturnRequested, `earlyReturnRequested = fmi3True`>> of <<fmi3IntermediateUpdateCallback>>, or
 ** the FMU's request for a communication point at <<lastSuccessfulTime>> for any other reason.
++ 
+<<earlyReturn,`earlyReturn = fmi3True`>> must not be used to signal that the requested step was not completed due to an error _[(e.g. integrator encountered division by zero)]_.
+Instead, the function must return with an error code (<<fmi3Error>> or <<fmi3Fatal>>).
```
pmai, line 120: I think we also must point to fmi3Discard, which is another way the function can return immediately (without earlyReturn being true), e.g.:

```suggestion
Instead, the function must return with an error code (<<fmi3Error>> or <<fmi3Fatal>>).
Alternatively, the function can return with <<fmi3Discard>>, if the FMU is in the same state as before the call, i.e. the output argument values are undefined, but the computation may continue.
```

pmai (changes requested):
I think the wording is fine for me, but given that fmi3Discard also still exists, we should point to that as well, as alternatives to earlyReturn being true.

chrbertsch (approved):

andreas-junghanns (approved):

pmai (approved):

PR #1849, opened 2023-01-31
Clarify use of binaries and resources folder

### LINKED ISSUE #1822: Relax content description of /binaries folder

chrbertsch:
New apptempt to fix #1822 as discussed in todays design meeting. See https://github.com/modelica/fmi-standard/pull/1832#issuecomment-1410453308

### DIFF
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -88,7 +88,6 @@ The header files `fmi3PlatformTypes.h`, `fmi3FunctionTypes.h` and `fmi3Functions
 A binary FMU must contain the binary files - that are needed at linking or loading time - of the FMU for all supported platforms in this folder.
 To use the binaries of a specific platform, all items in that platform-specific binary folder must be unpacked at the same location as the binary `<modelIdentifier>.{dll|dylib|so}`.
 
-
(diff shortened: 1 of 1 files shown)


pmai (approved):
Looks good to me.

Stylistically I might prefer (en-)dashes rather than parens for the parenthetical expressions, but is just my typographic preference.

t-sommer (approved):

andreas-junghanns (approved):

docs/2_5_fmu_distribution.adoc
```diff
@@ -50,7 +50,8 @@ sources                             // directory containing the C sources (optio
 binaries                            // directory containing the binaries (optional)
    x86_64-windows                   // binaries for Windows on Intel 64-bit
       <modelIdentifier>.dll         // shared library of the FMI implementation
-      <other DLLs>                  // the DLL can include other DLLs
+      <other files>                 // other platform dependend files
```
MBlesken, line 53: @chrbertsch is it clear enough that <other files> are allowed in parallel subfolders as well?

PR #1850, opened 2023-01-31
clarify usage of structural parameters in glossary

### LINKED ISSUE #1821: Use cases for structural parameters
During the FMI Design Meeting (2022-10-26) we had a discussion on use-cases for structural parameters and Configuration Mode. Some of us thought that structural parameters are **only** intended for dimensions of variables of the FMU. But they can also be used to mark variables, which must be set **before** Initialization Mode.

I went through the FMI 3.0 Specification: we mention this fact only once: https://fmi-standard.org/docs/3.0/#ConfigurationMode. 
Most of the other occurances of structural parameters refer to changeable dimensions of variables.

We should mention at least this other use-case too. At least in the Glossary.

chrbertsch:
fixes #1821

### DIFF
--- a/docs/7___appendix.adoc
+++ b/docs/7___appendix.adoc
@@ -217,7 +217,8 @@ The discrete-time states of a model are time-discrete variables that have two va
 Step events are, for example, used to change the mapping of the continuous states to variables (<<dynamic-state-selection>>).
 
 |_structural parameter_
-|A parameter that can be set in <<ConfigurationMode>> before <<InitializationMode>>, e.g.,  for influencing the size and/or dimensionality of an array variable of an FMU.
+|A parameter that can be changed in <<ConfigurationMode>>, i.e. prior to <<InitializationMode>>, and, if tuneable, also in <<ReconfigurationMode>>.
+Structural parameters can be used to influence the size and/or dimensionality of array variables of an FMU, for example.
 
 |[[super-dense-time,super-dense time]]_super-dense time_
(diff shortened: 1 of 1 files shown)


docs/7___appendix.adoc
```diff
@@ -217,7 +217,7 @@ The discrete-time states of a model are time-discrete variables that have two va
 Step events are, for example, used to change the mapping of the continuous states to variables (<<dynamic-state-selection>>).
 
 |_structural parameter_
-|A parameter influencing the size and/or dimensionality of an array variable of an FMU.
+|A parameter that can be set in <<ConfigurationMode>> before <<InitializationMode>>, e.g.,  for influencing the size and/or dimensionality of an array variable of an FMU.
```
pmai, line 220:
```suggestion
|A parameter that can be changed in <<ConfigurationMode>>, i.e. prior to <<InitializationMode>>, and, if tuneable, also in <<ReconfigurationMode>>.
Structural parameters can be used to influence the size and/or dimensionality of array variables of an FMU, for example.
```

pmai (changes requested):
I think this should mention reconfiguration mode, and set of the example more clearly.

andreas-junghanns (approved):
The end of the second sentence leaves the door open to other uses of structural parameters (as @TorstenBlochwitz keeps reminding us). Maybe, just maybe, this could be more explicit?

pmai (approved):

PR #1851, opened 2023-02-12
Initial proposal for FMI layered standard manifest

pmai:
This provides the initial proposal for the FMI layered standard manifest information that is suggested for layered standards to support.

Modifications might be needed to make the whole added part non-normative, and we might want to also provide a schema that allows checking of those attributes.

This is against the 3.0.x branch to ease discussion/merger, backmerge to main can be performed afterwards.


chrbertsch (commented):
Looks good to me. 
Would it make sense to define an XSD for the minimal manifest file, which could also be a template for more complex XSD files if more elements are added?

andreas-junghanns (approved):
Otherwise I think this proposal is as discussed.

chrbertsch:
Discussion: 
Pierre: an xsd does not privide a huge benefit. 
Pierre: currently it is normative text, not in brackets.
Christian: I would leave it as it is.
Matthias: I have no strong opinion on it.

Issue #1837, opened 2022-11-28
fmi3ModelDescription.xsd: Missing Dimension and Annotations elements for Enumerations


IZacharias:
According to Figure 28 in FMI 3.0 specification, array variables of type enumeration should be possible:
![image](https://user-images.githubusercontent.com/33029000/204255996-2d0e97f6-cb44-4e70-bf76-bde2310a973e.png)
However, the fmi3ModelDescription schema does not allow this. The Enumeration element does not contain a Dimension element:
![image](https://user-images.githubusercontent.com/33029000/204255454-bed1b6a6-16e9-4204-af9d-6b5881c23a4b.png)
Furthermore, the Annotations element is also missing.

I don’t see any reason to forbid Enumeration arrays. Therefore, I assume that this is a bug in the schema and not in the specification document.

pmai:
Hi Irina, the schema is actually correct, as far as I can see, in that `fmi3Enumeration` is an extension of `fmi3MandatorilyTypedInitializableVariable`, which is an extension of `fmi3MandatorilyTypedArrayableVariable`, which is an extension of `fmi3TypedArrayableVariable`, which is an extension of `fmi3ArrayableVariable` which has the `Dimension` elements, which is again an extension of `fmi3AbstractVariable`, which has the `Annotations` element.

I don't know where the second diagram is coming from, but it seems that that program does not show the inherited/extension base content, but only the added content, leading to the confusion?

IZacharias:
Hi Pierre, I think there is a bug in `fmi3MandatorilyTypedArrayableVariable`. It is not an extension of `fmi3TypedArrayableVariable` but `fmi3TypedArrayableVariable `is specified as restriction:

```
<xs:complexType name="fmi3MandatorilyTypedArrayableVariable" abstract="true">
        <xs:complexContent>
            <xs:restriction base="fmi3TypedArrayableVariable">
                <xs:attribute name="declaredType" type="xs:normalizedString" use="required"/>
            </xs:restriction>
        </xs:complexContent>
    </xs:complexType>
```
I guess that this would be correct:
```
<xs:complexType name="fmi3MandatorilyTypedArrayableVariable" abstract="true">
	<xs:complexContent>
		<xs:extension base="fmi3TypedArrayableVariable">
			<xs:attribute name="declaredType" type="xs:normalizedString" use="required"/>
		</xs:extension>
	</xs:complexContent>
</xs:complexType>
```

But then the declaredType attribute must be treated in a different way.

[...]

PR #1853, opened 2023-02-20
Merge branch 'main' into v3.0.x

t-sommer:


chrbertsch (approved):
I have not seen anything that should not go into FMI 3.0.1

andreas-junghanns (approved):
I can approve, with two open points:
 - the .png´s got regenerated and I did not see the differences
 - The fmi3Variable.xsd was changed and while I can convince myself the changes have no effect on the xml, I am not sure, an expert should agree here

pmai (approved):
Seems sensible; the schema changes were due to #1838, which fixes #1837. The change should keep all valid 3.0 XML files valid in 3.0.1, but allows enums to have annotations and dimensions, which was the documented state in the spec itself.

PR #1854, opened 2023-03-01
Enhance ls manifest description with schema

pmai:
This adds a schema for the manifest attributes and the default minimal manifest, and enhances the examples to showcase that to some extent.

### DIFF
--- a/docs/2_6_versioning_layered_standards.adoc
+++ b/docs/2_6_versioning_layered_standards.adoc
@@ -67,7 +67,7 @@ A minimal example of a manifest file, for layered standards that have no need fo
 include::examples/fmi_ls_manifest_minimal_example.xml[]
 ----
 
-The standard includes a schema for the attribute definitions and the minimal example element as `fmi3LSManifest.xsd`.
+The standard includes a schema for the attribute definitions and the minimal example element `fmiLayeredStandardManifest` shown above as `fmi3LayeredStandardManifest.xsd`.
 An example schema that describes the simple example of a manifest above is shown below:
 
 [source, xsd]
--- a/docs/2_6_versioning_layered_standards.adoc
+++ b/docs/2_6_versioning_layered_standards.adoc
@@ -66,3 +66,12 @@ A minimal example of a manifest file, for layered standards that have no need fo
 ----
 include::examples/fmi_ls_manifest_minimal_example.xml[]
 ----
+
+The standard includes a schema for the attribute definitions and the minimal example element as `fmi3LSManifest.xsd`.
+An example schema that describes the simple example of a manifest above is shown below:
+
+[source, xsd]
+----
+include::examples/fmi_ls_manifest_example_schema.xsd[]
+----
+
--- a/docs/examples/fmi_ls_manifest_example.xml
+++ b/docs/examples/fmi_ls_manifest_example.xml
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<MyRootElement 
+<MyRootElement
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:noNamespaceSchemaLocation="fmi_ls_manifest_example_schema.xsd"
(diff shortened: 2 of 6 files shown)


docs/2_6_versioning_layered_standards.adoc
```diff
@@ -66,3 +66,12 @@ A minimal example of a manifest file, for layered standards that have no need fo
 ----
 include::examples/fmi_ls_manifest_minimal_example.xml[]
 ----
+
+The standard includes a schema for the attribute definitions and the minimal example element as `fmi3LSManifest.xsd`.
```
chrbertsch, line 70: What is the "minimal example element"? The default root element `FMILSManifest`?

chrbertsch (commented):
Shall we include a graphical illustration of the new schema as for the other schemas?

schema/fmi3LSManifest.xsd
```diff
@@ -0,0 +1,93 @@
...
+                of the fmi-ls-manifest.xml file in the sub-directory of extra
+                mandated by a layered standard.
+            </xs:documentation>
+        </xs:annotation>
+        <xs:attribute ref="fmi-ls:fmi-ls-name" use="required"/>
+        <xs:attribute ref="fmi-ls:fmi-ls-version" use="required"/>
+        <xs:attribute ref="fmi-ls:fmi-ls-description" use="required"/>
+    </xs:attributeGroup>
+
+    <xs:element name="FMILSManifest">
```
PTaeuberDS, line 81: Not sure if I like the casing. Maybe we should stick to simple camel case as in "fmiModelDescription"?

docs/examples/fmi_ls_manifest_example.xml
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<MyRootElement 
+<MyRootElement
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:noNamespaceSchemaLocation="fmi_ls_manifest_example_schema.xsd"
   xmlns:fmi-ls="http://fmi-standard.org/fmi-ls-manifest"
   fmi-ls:fmi-ls-name="org.fmi-standard.demo-ls-name"
```
PTaeuberDS, line 6: Do we need the attribute prefix in addition to the namespace? Wouldn't "fmi-ls:name" be sufficient?

docs/examples/fmi_ls_manifest_example_schema.xsd
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+    xmlns:vc="http://www.w3.org/2007/XMLSchema-versioning" elementFormDefault="qualified"
+    vc:minVersion="1.1"
```
PTaeuberDS, line 4: Is version=1.1 necessary for this example? Otherwise we could remove it.

PTaeuberDS (commented):
This is basically the proposal that I made in fmi-ls-xcp so it's fine for me.

chrbertsch:
It would be good to finalize and merge this. 
Let's discuss this at the beginning of tomorrow's design webmeeting.

docs/examples/fmi_ls_manifest_example.xml
pmai: It technically would, however it is probably still a good idea to have the more explicit names, given that in some circumstances the namespace is not readily apparent.

schema/fmi3LSManifest.xsd
pmai: I've regularized and spelt out the naming some more.

docs/2_6_versioning_layered_standards.adoc
pmai: Adding a diagram would not help much, since it would only show one box given our diagram capabilities. Additionally the schema is for other schema authors, hence they should be capable of reading or viewing it in their own tools...

PR #1857, opened 2023-03-16
Mention FMI 3.0.1, change copyright year

chrbertsch:


andreas-junghanns (approved):

chrbertsch:
Updating the copyright year w/o changes in the files is not necessary (see https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated). So we could revert most of of the changes.

chrbertsch:
Copyright year was already changed in a different PR

PR #1875, opened 2023-05-16
Add pointer to IG for language and compiler names

pmai:


chrbertsch (commented):
Should we explicitly point to section 3.5? (https://modelica.github.io/fmi-guides/main/fmi-guide/#_compiler_dependencies)

pmai:
> Should we explicitly point to section 3.5? (https://modelica.github.io/fmi-guides/main/fmi-guide/#_compiler_dependencies)

I think this might be a bit problematic, as we might restructure or otherwise change the anchors, and then the link would be dangling; hence currently I think all links are only to the root of the IG, which I've adhered to in this PR.

But we might want to change this, however then we probably should define more durable anchors first in the adoc...

chrbertsch (approved):

PR #1891, opened 2023-07-13
Add Direct Memory Access

### LINKED ISSUE #515: Re-use memory in fmi3SetX / fmi3GetX functions
Currently all requested values are copied twice for every call of fmi3GetX / fmi3SetX which is inefficient in terms of bandwidth and memory usage. This can be avoided by re-using the memory for subsequent calls of the getters and setters.
A possible implementation is outlined below. (It is the result of a discussion with @pmai)

## Instantiation

The FMU allocates the memory for all variables during `fmi3Instantiate` using the callbacks provided by the environment.
For FMUs that don't support variable array sizes this memory may be static. The FMU may even use this memory directly for it's calculations.
The Memory layout is defined by the model description: it is the same as for a get / set call where all variables of a respective type are set / retrieved in the same order as they appear in the model description. 

## Get variables

```
fmi3Status fmi3GetX(const fmi3ValueReference vr[], size_t nvr, fmi3X* values[], size_t *nValues);
```

- `vr`: all variables the caller is interested in (NULL = all)
- `nvr`: size of vr
- `values`: the memory that contains the variables
- `nValues`: size of `values`

## Set variables

```
fmi3Status fmi3SetX(const fmi3ValueReference vr[], size_t nvr, fmi3X values[], size_t nValues);
```

- `vr`: all variables that have changed since the last call to fmi3SetX (NULL = all)
- `nvr`: size of vr
- `values`: memory retrieved from `fmi3GetX`
- `nValues`: size of `values`

The value references are provided to allow optimizations if only a small number of variables are set / retrieved.

## Structural changes

The FMU re-allocates the memory using the callbacks provided by the environment. The memory layout changes according to the structural parameters.

## Termination 

In fmi3Terminate() the FMU frees the memory using fmi3Free()

Comments welcome!

t-sommer:
fixes #515

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -1330,3 +1330,36 @@ include::Reference-FMUs/examples/jacobian.c[tags=GetJacobianAdjoint]
 ----
 
 _]_
+
+==== Direct Memory Access [[direct-memory]]
+
+The optional Direct Memory Access (DMA) API allows importers to access memory allocated by the FMU and vice versa.
+_[This eliminates the need to copy the values upon every call to `fmi3{Get|Set}{VariableType}()`]_.
+The variables that support DMA have to be listed in the `directMemory.xml` at the root of the FMU.
+
+[source, XML]
+----
+include::examples/directMemory.xml[]
+----
+
+An FMU that contains the `directMemory.xml` must also implement all functions declared in `fmi3DirectMemory.h`.
+These functions allow direct access to the variables memory.
+
+[source, C]
+----
+include::../headers/fmi3DirectMemory.h[tags=GetFloat64Pointer]
+
+include::../headers/fmi3DirectMemory.h[tags=SetFloat64Pointer]
+----
+
+The `fmi3{Get|Set}{VariableType}Pointer()` functions follow the same rules as the respective `fmi3{Get|Set}{VariableType}()` functions with the following exceptions.
+
+===== Providing Memory for Outputs
+
+Memory for variables with `causality = output` can only be set before entering <<Initialization Mode>>.
+_[This way the FMU does not have to allocate memory for these variables.]_
+
+===== Lifetime of the Memory
+
+The memory shared between the importer and the FMU may not be changed by the importer during a function call and any change must be communicated to the FMU by a call to `fmi3Set{VariableType}Pointer()`.
+Memory of variables with `causality = output` may change during function calls and have to be retrieved using `fmi3Get{VariableType}Pointer()`.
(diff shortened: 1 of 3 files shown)


chrbertsch:
Design Meeting:
Pierre: For efficient Co-Simulation we need double buffering. You want to run the FMUs in parallel. Thus we need double buffers, otherwise you do not get a benefit. 
Torsten: you still get a benefits
Pierre: Single buffers are not sufficient
Torsten: The pointers might be different
Klaus: We should mention this in the layered standard. 
Pierre: Switching the pointers for individual variables gets cost intensive. In order to make this efficient either pass the pointers once .... or we need a much better API
Torsten: My intention is for a variable representing 1GByte of data, not for many "small" variables
Pierre: We already have a memory order / area in the get/set call for multiple variables.
We should do just a setReal, which only affects the pointers.
We should use one pointer for multiple variables. Currently we pass a pointer, but the memory is "away" when you leave the set-call
Torsten: ... we could have an array of pointers 
Pierre: this kind of indirections is expensive due to address calculation. You must make this "apparent" to the CPU. One has to dereference the memory for each individual variable. If you do this only once, you can cache it.
Torsten: You talk about multiple variables, I about few of large size.
Pierre: for resizable variable this introduces a new error possibility.
Why can we not do just a "setReal", just with a guarantee to not take the memory away after the end of the setReal call? ... to be discussed : valid until the when ...
That might be sufficient all of our needs.
Torsten: There is  a tradeoff .... depending if the memory of the individual variable is distributed
Torsten: Alternative: one could provide an opaque pointer  and define the internal memory structure
Pierre: This would be very complicated with few benefit.
Pierre: we have to figure out the different use cases. Our intuition fails on today's hardware. We should make benchmark experiments. E.g. multi-threaded FMU calculation, looking into performance counters.
Torsten: My goal was to keep this out of FMI and let the importer handle it.
Pierre: Memory Copy performance is very nonlinear ...
Torsten: I could create examples based on the reference FMUs
Christian: We should start an FCP listing use cases
Pierre: There is experience, e.g. with OSI (with double buffering)
Torsten: for the use cases, just commit to the documentation in this branch
Pierre: listing the uses cases is important to understand this concept
Torsten: This does not enable new functionality, but is intended to decrease CPU-time
Torsten: I will set up a meeting.

timrulebosch:
I'm leaning towards a method of allowing direct memory access using flatbuffers to represent the FMU (instance data). You might find the basic idea interesting.
https://github.com/boschglobal/dse.fmi/issues/47

My use case is packaging vECU into FMU container, simple Co-Simulation. I want to bypass the FMI C API for all data marshalling, and instead exchange scalar variables directly with the vECU (bypass using memcpy), and "stream" binary data from a network buffer directly into the FMU by simply adjusting a pointer (zero copy).

PR #1892, opened 2023-07-14
Clarify default variability for discrete variables

### LINKED ISSUE #1890: Clarification on fmi2SetXXX
I need a clarification on when a simulation tool is allowed to call the fmi2SetXXX functions.

In version 2.0.4, page 50 is stated:

> fmi2SetXXX" can be called on **any** variable for ModelExchange in Continuous-Time Mode

In version 2.0.4, page 90 the table states:

> fmi2SetReal can be called in "instantiated", "initilization mode", "event mode", "continuous-time mode"
> fmi2SetInteger can be called in "instantiated", "initilization mode", "event mode"
> fmi2SetBoolean  can be called in "instantiated", "initilization mode", "event mode"
> fmi2SetString can be called in "instantiated", "initilization mode", "event mode"

1. Which page is correct?
2. If page 90 is correct: Why is it not allowed to call fmi2SetInteger, fmi2SetBoolean and fmi2SetString in continuous time?
3. If page 90 is correct: Why was this restriction dropped in fmi3?
4. If page 90 is correct: Since most simulations tools call fmi2SetInteger, fmi2SetBoolean and fmi2SetString in continuous time mode anyway. Why not change the standard and allow these functions to be called in continuous time mode? (For example Simulink, MapleSim and Dymola call these functions in continuous time mode.)

t-sommer:
fixes #1890

### DIFF
--- a/docs/2_2_common_schema.adoc
+++ b/docs/2_2_common_schema.adoc
@@ -823,7 +823,7 @@ Co-Simulation: By convention, the variable is from a "real" sampled data system 
 ModelExchange: No restrictions on value changes. +
 Co-Simulation: By convention, the variable is from a differential
 
-The default is `"continuous"`.
+The default is `continuous` for variables of type `Real` and `discrete` for all other types.
 
 _[Note that the information about continuous states is defined with element
 `fmiModelDescription.ModelStructure.Derivatives`.]_


nik5346:
@t-sommer 
The change needs to be applied to the XSD as well (Documentation page 47):
image
Make it consistent with the comment for the "initial" default.

pmai:
I don't think this change can be done in 2.0, as it is not backwards compatible nor any kind of bug fix. The default is clearly defined in both text and schema. Changing it now just makes this problematic. And exporters really cannot rely on this change, so they have to do what they do now, explicitly setting the variability, anyways.

Other than maybe putting in a non-normative comment to point out that non-Real variables cannot have the default variability and hence need to explicitly specify their variability. I don't see what can or should be done here.

nik5346:
yes it is clearly defined in the schema. The schema allows you not set the variability.
If the "variability" it is not set, the default is "continuous".

Since a integer can not be continuous as stated in the PDF, xsd and pdf are contradictory. Therefor this case is undefined and surely a bug.

pmai:
It is not undefined in the sense that it is clearly defined as invalid, and it cannot be "fixed" in 2.0.

nik5346:
That's correct, but why is the default for "initial" handled differently then?

nik5346:
> ... so they have to do what they do now, explicitly setting the variability, anyways.

Here is an example with an undefined variability in the reference FMUs:
https://github.com/modelica/Reference-FMUs/blob/main/Stair/FMI2.xml

pmai:
Because it is? The default for variability is handled differently in FMI 3.0, too. It could have been handled like this in 2.0. But it wasn't. Some considerations like I pointed out in the issue might have gone into it, and just were decided differently for initial, e.g. because the rules are more complex, or because initial was only introduced in 2.0, so different people at a different time discussed and designed it differently. Or it might just have been an oversight, though a long standing one, as already FMI 1.0 handled it like this.

While there might be (limited) historic interest in finding the reasons, from a standardisation perspective none of this matters. The standard is clear enough (default is continuous, only reals are allowed to be continuous, hence non-Reals must have explicit variability) and non-contradictory, so that any further interpretation by intent or precedent is ruled out.

We might not like what the standard says (and the majority of us obviously didn't, hence the change in 3.0), but it says what it says ... 

And since this does not break any functionality, it is on the benign side of unlikeable aspects of a standard, IMHO.

nik5346:
> In summary: It is not that I don't get what you are saying, I'm just saying you are wrong and have no leg to stand on, because you are approaching interpreting a standard in the wrong way.

pmai:
> > ... so they have to do what they do now, explicitly setting the variability, anyways.
> 
> Here is an example with an undefined variability for a integer variable in the reference FMUs: https://github.com/modelica/Reference-FMUs/blob/main/Stair/FMI2.xml

Which is clearly a bug in the reference FMUs and should be fixed (especially since it does explicitly specify the variability for the time variable, where the default would suffice). PR is likely appreciated by @t-sommer.

> The PDF documentation does not state, that the variability attribute must be set.

It does not have to state this, as it is implied by the things it does state: (Rule 1) If you do not specify variability, it defaults to continuous, and (Rule 2) only variables of type Real are allowed to have a continuous variability. Implication: Non-reals cannot use the default, because that would violate rule 2. And it's not the PDF documentation, it is the standard, which consists of the PDF and the schema (and the header files).

nik5346:
> It does not have to state this, as it is implied by the things it does state: (Rule 1) If you do not specify variability, it defaults to continuous, and (Rule 2) only variables of type Real are allowed to have a continuous variability. Implication: Non-reals cannot use the default, because that would violate rule 2. And it's not the PDF documentation, it is the standard, which consists of the PDF and the schema (and the header files).

That correct, but it would be much clearer, if it was handled like the "default" for "initial" is.

pmai:
> > It does not have to state this, as it is implied by the things it does state: (Rule 1) If you do not specify variability, it defaults to continuous, and (Rule 2) only variables of type Real are allowed to have a continuous variability. Implication: Non-reals cannot use the default, because that would violate rule 2. And it's not the PDF documentation, it is the standard, which consists of the PDF and the schema (and the header files).
> 
> That's correct, but it would be much clearer, if it was handled like the "default" for "initial" is.

It might well have been - personally I would have preferred if variability had been mandatory from the get go without any default, even at the cost of increased XML sizes for large models - but that is not what happened, and for 2.0 this is non-changeable. For 3.0 it was changed - though this also caused some trouble, IIRC - so there.

nik5346:
I don't get, why changing the XSD would break something.

Just remove the default from the XSD file and set the "required" tag. Hence all tools should state it anyway from the very first fmi2 release.

pmai:
> I don't get, why changing the XSD would break something.
> 
> Just remove the default from the XSD file and set the "required" tag. Hence all tools should state it anyway from the very first fmi2 release.

This would suddenly make all FMI 2.0 FMUs that correctly relied on the default value invalid. And changing the default to depend on the type of the variable (like 3.0 does) does not work either, because it would allow FMUs to rely on this new default, however all FMI 2.0 tools will still have the continuous default, breaking the new FMUs. And implementations would not even have a way of detecting "new" FMUs, as the patch version is not carried in the version attribute of an FMU (only major and minor version are carried, because patch releases are not supposed to change semantics like this).

Changing a standard is very much unlike changing a piece of software, as no one is forced to adjust their implementations of the standard, just because a new patch release is released. And for FMI, this is especially so, as FMUs tend to stay around forever, without being regenerated.

nik5346:
what about fmi 2.1 then?

pmai:
> what about fmi 2.1 then?

Depending on our rules for backward/forward compatibility, even this might be difficult, though it might be possible (would need further thought).  But that is a theoretical discussion, as no one will want to do a 2.1, given that 3.0 has been out for a year. Even if we published a 2.1, it is dubious that implementors will want to spend effort on implementing a 2.1, given that 3.0 has many more features, and is the way to go. We would need compelling new features for a 2.1 to gain traction, and this would detract from support for 3.0, so is not in anyone's interest.

> At very least the comment in the XSD should then change.

Why? I'm not seeing anything that is wrong in there...

nik5346:
```adoc
The default is `"continuous"`.
The default is `continuous` for variables of type `Real` and `discrete` for all other types.
```

I don't think this change is verbose enough to comprehend two days worth of discussion.

pmai:
> ```
> The default is `"continuous"`.
> The default is `continuous` for variables of type `Real` and `discrete` for all other types.
> ```
> 
> I don't think this change is verbose enough to comprehend two days worth of discussion. It is even incorrect, since there is no default for other types, but it must be stated explicitly.

Indeed the PR is incorrect, and should not be merged as is, as it is wrong, as you say: The default is continuous regardless of type. It is just not allowed to use continuous (whether implicitly by not providing a value and thus getting the default value or explicitly by providing continuous) for non-real type variables.

I've changed the PR to just add a non-normative comment that spells this out explicitly.

> Anyway. Thanks for your support. I think I understood how to handle the fmi2Integer now :).

Glad to be of service :).

t-sommer:
> _[Note that therefore, for all variables that are not of `type = "Real"`, a variability value other than `"continuous"` has to be supplied with an explicit `variability` attribute.]_

But wouldn't that render all (existing) FMI 2.0 FMUs illegal that do not explicitly specify the variability for non-Reals?

pmai:
> > _[Note that therefore, for all variables that are not of `type = "Real"`, a variability value other than `"continuous"` has to be supplied with an explicit `variability` attribute.]_
> 
> But wouldn't that render all (existing) FMI 2.0 FMUs illegal that do not explicitly specify the variability for non-Reals?

They are already illegal, this just points this fact out more concisely: If you do not specify a variability, the default since 1.0 has been continuous, and since 1.0 it has been forbidden to have non-real variables with continuous variability.

There is no difference between specifying an illegal variability via the default or explicitly - indeed for most validating parsers, with the current schema, the default would have been substituted in, so any processing code cannot really distinguish between the absence and an explicit value (non-validating parsers of course can do so, and you probably can interrogate some of the validating parsers).

That being said, since such an FMU is clearly illegal, an implementation can do as it pleases, and substituting discrete for continuous is as valid a course of action as any other. But an FMU cannot and should not rely on this non-standard behavior.

nik5346:
I think this is the case. 

According to pmai this has always been illegal, since the default is "continuous" and an integer can't be "continuous". Therefor this was always an implicit rule. Now it is explicit and the standard did not change, but was only clarified.

t-sommer:
Okay, fair enough.

chrbertsch (approved):

PR #1893, opened 2023-07-18
Backport #1892 and #1889 to Word

chrbertsch:


t-sommer (changes requested):
p. 49: `variability` should be formatted as monospace.

t-sommer (approved):

PR #1913, opened 2023-11-21
Bump fmi3Version to "3.0.1" in fmi3Functions.h

t-sommer:


chrbertsch (changes requested):
Hm. So we missed this in the FMI 3.0.1 release.
But when we have a new release, it will be 3.0.2, so shouldn't we change this accordingly already?
And 2.2.5 still states "The standard header file as documented in this specification has version 3.0, so this function returns 3.0." and should also be changed.

pmai:
Should this change at all? The standard states:

> Patch versions will only change explanatory text of the standard, make formerly defined content clearer, without any other changes to the XML schemas or other content definitions. For this reason, the version number attribute of all FMI files will only contain major and minor version numbers and not the patch version number: It should never be necessary for an importer to know the patch version number of the standard that the generating tool implemented.

Given this rationale, that the XML attribute is supposed to say 3.0 for 3.0.1, and that 3.0.1 did not change the version in the header, I think this ship has sailed and fmi3Version as well as fmi3GetVersion should likewise only report the major.minor version...

PR #1914, opened 2023-11-22
Add "universal2" to architectures

t-sommer:
to support universal binaries on macOS

### DIFF
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -119,8 +119,8 @@ Architecture `<arch>`::
 |x86_64
 |Intel/AMD x86 64-bit
 
-|universal
-|Binaries for multiple architectures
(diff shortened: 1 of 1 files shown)


pmai:
Given that this is potentially not limited to macOS, and the description, maybe an arch name of `multi` or similar is a more neutral approach.

chrbertsch:
FMI Design Web meeting
- Jorge :  how is this related to emulation?
- Pierre: here we are not talking about emulation, the binaries are mutli-binary. I would avoid "universal".  Universal 2 is a marketing name. Technically it is "fat binary". The important thing is, that one has to look into the file format. We have the operating system anyway.
- Torsten S: We could also recommend to separate the binaries out. 
- Pierre: you still have to see if the binary runs on your hardware. You could also put fat binaries into the individual binary folders.

PR #1937, opened 2024-04-17
Clarify multiple Clock ticks per Event Mode for different Clock types

### LINKED ISSUE #1916: Handling of time-based Clocks with interval=0
In section [2.3.5. State: Event Mode](https://fmi-standard.org/docs/3.0.1/#EventMode) in the description of the fmi3SetClock function the following restriction is made for time-based clocks:

"Only [time-based Clocks](https://fmi-standard.org/docs/3.0.1/#time-based-clock) must not be active for more than one call of [fmi3UpdateDiscreteStates](https://fmi-standard.org/docs/3.0.1/#fmi3UpdateDiscreteStates) per [Event Mode](https://fmi-standard.org/docs/3.0.1/#EventMode)."

This restriction contradicts the possibility for aperiodic time-based clocks to return an interval=0, which is possible according to the interval definition for, e.g., changing clocks:
image

The restriction should only be made for periodic clocks:
"Only ~[time-based Clocks](https://fmi-standard.org/docs/3.0.1/#time-based-clock)~ periodic Clocks must not be active for more ..."

PTaeuberDS:
Fixes #1916 
Removes the contradicting statement that aperiodic Clocks are not allowed to be activated more than once per Event Mode.

### DIFF
--- a/docs/2_3_common_states.adoc
+++ b/docs/2_3_common_states.adoc
@@ -573,7 +573,8 @@ _As part of the iterations to solve the algebraic loop, latexmath:[v] acquires a
 _If the final (or some intermediate) value of latexmath:[v] no longer activates the Clock, then this Clock must be deactivated by the Importer during the same super-dense time instant.]_ +
 The importer can set a Clock to `fmi3ClockInactive` only if the Clock has the attribute <<canBeDeactivated,`canBeDeactivated = true`>>.
 Any Clock active during <<fmi3UpdateDiscreteStates>> must be deactivated by the FMU itself. +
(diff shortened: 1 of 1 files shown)


masoud-najafi (approved):

chrbertsch (approved):
FMI Design Webmeeting: 

Klaus: we have  contratiction for aperiodic clocks. Either we disallow an interval 0, or allow triggering more than once.
Markus Suevern: There could be a use case for interval 0, e.g. in LS-BUS. Interval 0 means: "stay in event mode".
Torsten: We should merge it, it is a contradiction. We overlooked this case.
Klaus: We should clearly mention this in the release notes.

PR #1944, opened 2024-05-15
Clarify the information flow of clocks

### LINKED ISSUE #1917: Standard is not clear about the information flow of Clocks
Apparently it is common understanding that input Clocks are able to provide information to other input Clocks (of another FMU).

This is not clear by reading the standard document.

In [2.2.7. Variables](https://fmi-standard.org/docs/3.0.1/#fmu-variables), causality is described as follows:
"The attribute causality defines the direction of the information flow with respect to the FMU (e.g. input, output, parameter)."

Table 17 says for causality:
= input: The variable value may be provided by the importer.
[For example, the importer could forward the output of another FMU into this input.]

= output: The variable value may be used by the importer.
[For example, this value can be forwarded to an input of another FMU.]
The algebraic relationships to the [inputs](https://fmi-standard.org/docs/3.0.1/#input) may be defined via the [dependencies](https://fmi-standard.org/docs/3.0.1/#dependencies) attribute of [<fmiModelDescription><ModelStructure><Output>](https://fmi-standard.org/docs/3.0.1/#Output).

If information flows from one input clock to another input Clock, the direction is not defined, which contradicts the general description of causality.

Example:
image

If Clocks are handled differently regarding the meaning of causality, it should be pointed out in the document.

klausschuch:
this should fix #1917

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -284,7 +284,9 @@ All variables are listed in <<modelDescription.xml>> as elements of `<fmiModelDe
 
 They are identified with a unique handle called <<valueReference, value reference>>.
 
-The attribute <<causality>> defines the direction of the information flow with respect to the FMU (e.g. <<input>>, <<output>>, <<parameter>>).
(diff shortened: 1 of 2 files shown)


andreas-junghanns (approved):

klausschuch:
confirmend by the whole group in the F2F meeting

PR #1955, opened 2024-06-06
Allow setting continuous states in Event Mode if reinit = false

### LINKED ISSUE #1950: Allow setFloat64() for continuous states in Event Mode if reinit = false
[2.3.5. State: Event Mode](https://fmi-standard.org/docs/3.0.1/#EventMode):

> Allowed Function Calls
Function [fmi3Set{VariableType}](https://fmi-standard.org/docs/3.0.1/#get-and-set-variable-values)
This function can be called for variables with [causality = input](https://fmi-standard.org/docs/3.0.1/#causality) and for variables with [causality = parameter](https://fmi-standard.org/docs/3.0.1/#causality) and [variability = tunable](https://fmi-standard.org/docs/3.0.1/#variability).

It should also be allowed to set continuous states in Event Mode by calling `fmi3Set{VariableType}` if [`reinit`](https://fmi-standard.org/docs/3.0.1/#reinit) = `false`.

t-sommer:
fixes #1950

### DIFF
--- a/docs/2_3_common_states.adoc
+++ b/docs/2_3_common_states.adoc
@@ -553,7 +553,7 @@ a|
 Allowed Function Calls::
 
 Function <<get-and-set-variable-values,`fmi3Set{VariableType}`>>::
-This function can be called for variables with <<causality,`causality = input`>> and for variables with <<causality,`causality = parameter`>> and <<variability, `variability = tunable`>>.
+This function can be called for variables with <<causality>> = <<input>>, <<causality>> = <<parameter>> and <<variability>> = <<tunable>>, and <<reinit>> = `false`.
 
 Functions <<get-and-set-variable-values,`fmi3Get{VariableType}`>>::
 Getting variables might trigger <<selective-computation,computations>>.
--- a/docs/2_3_common_states.adoc
+++ b/docs/2_3_common_states.adoc
@@ -553,7 +553,7 @@ a|
 Allowed Function Calls::
 
 Function <<get-and-set-variable-values,`fmi3Set{VariableType}`>>::
-This function can be called for variables with <<causality>> = <<input>>, <<causality>> = <<parameter>>, <<variability>> = <<tunable>> and <<state,continuous-time states>> with <<reinit>> = `false`.
+This function can be called for variables with <<causality>> = <<input>>, <<causality>> = <<parameter>>, <<variability>> = <<tunable>>, and <<state,continuous-time states>> with <<reinit>> = `false`.
 
 Functions <<get-and-set-variable-values,`fmi3Get{VariableType}`>>::
 Getting variables might trigger <<selective-computation,computations>>.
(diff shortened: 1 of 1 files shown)


KarlWernersson (commented):
did some changes

docs/2_3_common_states.adoc
```diff
@@ -553,7 +553,7 @@ a|
 Allowed Function Calls::
 
 Function <<get-and-set-variable-values,`fmi3Set{VariableType}`>>::
-This function can be called for variables with <<causality,`causality = input`>> and for variables with <<causality,`causality = parameter`>> and <<variability, `variability = tunable`>>.
+This function can be called for variables with <<causality>> = <<input>>, <<causality>> = <<parameter>>, <<variability>> = <<tunable>> and <<state,continuous-time states>> with <<reinit>> = `false`.
```
chrbertsch, line 556:
```suggestion
This function can be called for variables with <<causality>> = <<input>>, <<causality>> = <<parameter>>, <<variability>> = <<tunable>>, and <<state,continuous-time states>> with <<reinit>> = `false`.
```

chrbertsch (approved):

t-sommer:
Should setting (all) variables with `<<causality>> = <<parameter>>` really be allowed? I suggest to change

```
This function can be called for variables with <<causality>> = <<input>>, <<causality>> = <<parameter>>, <<variability>> = <<tunable>>, and <<state,continuous-time states>> with <<reinit>> = `false`.
```

to 

```
This function can be called for variables with <<causality>> = <<input>>, <<variability>> = <<tunable>>, and <<state,continuous-time states>> with <<reinit>> = `false`.
```

chrbertsch:
Torsten S:  I want to change the second comma do "and" .
Masoud: I will look into this regarding clocks.

masoud-najafi (commented):
May be I'm missing something here. Acccording to the definition of "reinit", if reinit=false, you are not allowed to reinitialize the state variable. So you cannot call fmi3SetXX.  Is it a typo or I'm missing something?

t-sommer:
[reinit](https://fmi-standard.org/docs/3.0.1/#reinit):

> If true, [state](https://fmi-standard.org/docs/3.0.1/#state) may be reinitialized by the FMU in [Event Mode](https://fmi-standard.org/docs/3.0.1/#EventMode).
If false, [state](https://fmi-standard.org/docs/3.0.1/#state) will not be reinitialized.

I think this is correct, since an importer may only set a continuous state in Event Mode, if the FMU does not recalculate it.

masoud-najafi:
OK, I got your point. Then let's first clarify the reinit attribute. Is it just for information purposes? What's useful for?
An FMU may go to the Event mode for several reasons. For a particular event, the FMU may reinitialize the state with reinit==true). For other events, the "importer" may reinitialize the same state.  Let's take the BouncingBall example.  The FMU is supposed to change the height and the velocity at zero-crossing event instants, so reinit for height and velocity is true. The importer may also be able to change these state variables at other events (other that zero-crossing events), for example if t>10, the ball change the velocity due to a sudden external wind. 
By saying "reinit=true", should the importer be forbidden from updating the state? I'm not really in favor of this.

PR #1971, opened 2024-09-04
Add missing architecture to ZIP file image

guilhermeAlmeida1:
Checklist

* [ ] ~~[Re-generated schema figures](https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#changing-the-xsd-schemas).~~ unnecessary
* [ ] [Linted the documents](https://github.com/modelica/fmi-standard/blob/master/CONTRIBUTING.adoc#building-the-specification-document). -> this seems to be a leftover check or the Contributing doc needs updating. I can't seem to find how to do this but I strongly doubt this will fail linting.

The current `Structure of the ZIP file` image is missing the alternative `x86-windows`.

### DIFF
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -49,6 +49,8 @@ terminalsAndIcons                   // FMU and terminal icons (optional)
 sources                             // directory containing the C sources (optional)
    buildDescription.xml
 binaries                            // directory containing the binaries (optional)
+   x86-windows                      // binaries for Windows on Intel 32-bit
+      <modelIdentifier>.so          // shared library of the FMI implementation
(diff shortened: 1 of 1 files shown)


chrbertsch:
The list of platforms is not complete.
Perhaps we should change the headline "Structure of the ZIP file of an FMU" to "Structure of the ZIP file of an FMU (with exemplary binary platform support)" instead of adding more and more specific platforms.
grafik

guilhermeAlmeida1:
Reading through this again I see that this it is actually well-written. The title above the image is a bit misleading and the options in the image led me to believe you wanted to list all platforms that were present in both FMI2 and FMI3, not just an example of some. Closing as resolved

Issue #1942, opened 2024-05-14
Define matching rule "none" in the base FMI standard


chrbertsch:
providing a matching rule is mandatory, see 
There might be use cases where one does not want one of the predefined matching rules to be applied, see https://github.com/modelica/fmi-ls-struct/issues/8

chrbertsch:
FMI Webmeeting:

Christian: Could this be defined in 3.0.2?
KLaus: could be done technically, but does not sound like a bugfix.

chrbertsch:
Pierre: I could live with this in 3.0.2. We are not really defining something new.
Klaus: It is a non-breaking change to make matching rule optional.
Pierrre: "none" is more than "do whatever you want", because it discourages doing something.
Klaus: the FMU tells "I do not know anything about matching".
Pierre: Could be the FMU/ layered standard author.
Klaus: "none" is undefined, "don't" would tell not to match.
Pierre: "none": for me means: do no automatic matching, "undefined": do whatever you want. 
Klaus: in FMI-LS-STRUCT we defined a specific matching rule

Poll: 
Do we want to add a new matching rule at all?
Yes: Pierre, Christian, Otto, Garron, Jan, Daniel
No: - 
Abstain: Rest


Do we want as new matching rules:

* "none" only  ("the creator of the FMU does not intend automatic matching", under user direction the tool can do whatever it pleases): Nikolai, Kharamon, Pierre, Garron, Jan, Klaus
* "none" and "undefined": Christian
* Abstain: Rest

Pierre: "none" would also be a good default

PR #1976, opened 2024-09-12
Introduce matching rule 'none'

snps-ribbe:
Fixing https://github.com/modelica/fmi-standard/issues/1942

### DIFF
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -1921,7 +1921,7 @@ An importer should connect terminals only if the number of member variables matc
 
 [[none,`none`]]
 |`none`
-|Matching of the variables is not intended by the creator. 
+|Matching of the variables is not intended by the creator of the FMU. 
 An importer may connect terminals based on the user's additional input.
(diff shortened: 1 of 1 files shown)


chrbertsch (approved):

docs/2_4_common_schema.adoc
```diff
@@ -1919,6 +1919,11 @@ An importer may connect terminals if some or no terminal member variables are pr
 |Matching of the variables is based on the order of the terminal member variables.
 An importer should connect terminals only if the number of member variables matches.
 
+[[none,`none`]]
+|`none`
+|Matching of the variables is not intended by the creator of the FMU. 
+An importer may connect terminals based on the user's additional input.
```
klausschuch, line 1925:
```suggestion
However, an importer may still connect terminals based on additional information.
```
The additional information might come from anywhere. I think we don't mention "the user" anywhere else and should avoid it here too.

klausschuch (approved):

pmai (approved):

PR #1979, opened 2024-09-24
Clarify when to use Scheduled Execution

### LINKED ISSUE #1948: Add section on when (not) to use Scheduled Execution
as discussed at the 2024-05-21 FMI-Design-Webmeeting

- when the importer needs control over the tasks to avoid overruns (typically only in hard real time applications)
- for synchronization use Clocks + CS instead

t-sommer:
fixes #1948

### DIFF
--- a/docs/1___overview.adoc
+++ b/docs/1___overview.adoc
@@ -119,13 +119,11 @@ image::images/co-simulation-data-flow.svg[width=40%, align="center"]
 The Scheduled Execution interface exposes individual <<model-partition, model partitions>>.
 A scheduler provided by the <<importer>> can control the execution of each model partition separately.
 In some ways the Scheduled Execution interface has similarities to the Model Exchange interface: the first externalizes a scheduling algorithm usually found in a controller algorithm (see <<fmi-for-scheduled-execution>>) and the second interface externalizes the ODE/DAE solver.
+_[See also https://modelica.github.io/fmi-guides/main/fmi-guide/#_use_cases_for_scheduled_execution[Use-cases for Scheduled Execution] in the FMI 3.0 Implementers' Guide.]_
 
 .Schematic view of data flow between user, the scheduler of the importer and model partitions of the FMU for Scheduled Execution
 [#figure-scheduled-execution-data-flow]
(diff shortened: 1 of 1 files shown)


docs/1___overview.adoc
```diff
@@ -124,6 +124,9 @@ In some ways the Scheduled Execution interface has similarities to the Model Exc
 [#figure-scheduled-execution-data-flow]
 image::images/scheduled-execution-data-flow.svg[width=40%, align="center"]
 
+_[An external scheduler is only required if the importer needs direct control over the tasks, e.g. to avoid overruns in a real-time application.
```
MBlesken, line 127: I don't like the "is only required" because this is not relevant perspective. Better "An external scheduler allows to control scheduling of tasks and thus model partitions. Such an environment e.g. is required to execute real-time simulations." 

If you really need to reference _overruns_ I suggest "Only in a real-time execution overruns can occur and thus can be stressed." Actually mentioning _preemption_ and _priorities_ would be just as relevant to mention. So I rather suggest not to mention any since it is explained in more detail in the SE section.

docs/1___overview.adoc
```diff
@@ -124,6 +124,9 @@ In some ways the Scheduled Execution interface has similarities to the Model Exc
...
+To synchronize events between the importer and across FMUs, <<Clock, Clocks>> should be used that are available in all interface types.]_
```
MBlesken, line 128: " The FMU provides model partitions defined by clocks in the FMU interface description. This way the importer can allow to configure the scheduling of each model partition, e.g. it enables to synchronize the execution across FMUs."

MBlesken (changes requested):

pmai:
Specific places to link to:

Section 8 generally on SE: https://modelica.github.io/fmi-guides/main/fmi-guide/#_scheduled_execution_specific_issues

Section 8.1 on Use Cases for SE: https://modelica.github.io/fmi-guides/main/fmi-guide/#_use_cases_for_scheduled_execution

chrbertsch (approved):
This was suggested in the FMI Steering Committee.

MBlesken (approved):

PR #1980, opened 2024-09-24
Clarify meaning of DefaultExperiment.stepSize for Model Exchange

### LINKED ISSUE #1939: Clarify meaning of DefaultExperiment.stepSize for Model Exchange
In FMI 2.0 the meaning of `DefaultExperiment.stepSize` for Model Exchange is defined in the XSD diagram (FMI Spec 2.0.4, p. 45)

image

In FMI 3.0 this description has been replaced with

> Furthermore, for Co-Simulation FMUs the [stepSize](https://fmi-standard.org/docs/3.0.1/#stepSize) defines the preferred `communicationStepSize`.

@casella suggested in https://github.com/CATIA-Systems/FMPy/issues/665, to use the `stepSize` in Model Exchange as the preferred output interval.

t-sommer:
fixes #1939

### DIFF
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -794,9 +794,9 @@ image::images/schema/DefaultExperiment.png[width=60%, align="center"]
 [[stepSize,`stepSize`]]`<DefaultExperiment>` consists of the optional default start time, stop time, relative tolerance, and step size for a simulation run.
 A tool may ignore this information. +
 _[However, it is convenient for a user that <<startTime>>, <<stopTime>>, <<tolerance>> and <<stepSize>> have already a meaningful default value for the model at hand.]_ +
-Furthermore, for Co-Simulation FMUs the <<stepSize>> defines the preferred <<communicationStepSize>>.
 <<startTime>>, <<stopTime>> and <<stepSize>> refer to the unit of the <<independent>> variable.
-For Model Exchange <<stepSize>> can be used as a default sample interval.
+<<stepSize>> defines the preferred <<communicationStepSize>> for Co-Simulation.
+For Model Exchange and Scheduled Execution <<stepSize>> has no defined meaning.
 
 [[ModelVariables,`<ModelVariables>`]]
 ==== Model Variables
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -796,6 +796,7 @@ A tool may ignore this information. +
 _[However, it is convenient for a user that <<startTime>>, <<stopTime>>, <<tolerance>> and <<stepSize>> have already a meaningful default value for the model at hand.]_ +
 Furthermore, for Co-Simulation FMUs the <<stepSize>> defines the preferred <<communicationStepSize>>.
 <<startTime>>, <<stopTime>> and <<stepSize>> refer to the unit of the <<independent>> variable.
+For Model Exchange <<stepSize>> can be used as a default sample interval.
 
 [[ModelVariables,`<ModelVariables>`]]
 ==== Model Variables


t-sommer:
Proposal by @KarlWernersson:

> For Model Exchange <<stepSize>> is the preferred step size for fixed step solvers.

t-sommer:
Poll at the Design Meeting:

- State that it has no meaning for ME: Nikolai, Otto, Peter, Christian W., Mohammed, Karl, Kashka
- "Keep" the old definition from 2.0: Torsten S., Karl, Khoa
- Not say anything: Klaus

KarlWernersson (changes requested):
I guess not updated,
Did we decide to go with the poll?

In that case
chachge to
For Model Exchange and scheduled execution, step size do not have a defined meaning.

chrbertsch (approved):

KarlWernersson (approved):

casella:
I'm sorry I could not participate to the phone meeting and the poll. I still see an open issue with the taken decision. Assume that you build some system model in Modelica and simulate it satisfactorily with a variable step-size solver and a certain communication interval. You then make an FMU and ship it to a third party, that will simulate it in its own simulation environment using FMI-ME. If we say that stepSize has no definite meaning, this means that the information about what could be a meaningful default communication interval is lost, and has to be guessed. What's the point of doing that?

We are currently testing OpenModelica's FMI generation by simulating all runnable models in open-source libraries using FMI-CS and FMPy. We observed that picking the right communication interval is sometimes crucial for simulation success or failure, so we are currently setting the stepSize attribute to the Interval experiment annotation value, so we can retrieve it when simulating the FMUs. With this new definition, this will no longer be possible. 

Again, what is the advantage of losing this information?

Thanks

casella:
Of course I would be perfectly fine if another attribute was added to specify the communication interval in FMI-CS mode. If that was the intention of this PR, I'm 100% happy about it.

chrbertsch:
> Assume that you build some system model in Modelica and simulate it satisfactorily with a variable step-size solver and a certain communication interval. You then make an FMU and ship it to a third party, that will simulate it in its own simulation environment using **FMI-ME**. If we say that stepSize has no definite meaning, this means that the information about what could be a meaningful default communication interval is lost, and has to be guessed. What's the point of doing that?
> We are currently testing OpenModelica's FMI generation by simulating all runnable models in open-source libraries using **FMI-CS** and FMPy. We observed that picking the right **communication interval** is sometimes crucial for simulation success or failure, so we are currently setting the stepSize attribute to the Interval experiment annotation valuen when generating the FMU, so we can later retrieve it when simulating the FMUs. With this new definition, this will no longer work for FMUs generated by other tools.
> Again, what is the advantage of losing this information in this scenario? I only see dis-advantages.

Hi @casella : I do not get your point. Do you talk about ME or CS FMUs? For CS FMUs nothing has changed, the FMI 3.0.2 standard still will tell "*stepSize* defines the preferred *communicationStepSize* for Co-Simulation."

casella:
Sorry, I meant ME where I wrote CS.

chrbertsch:
@casella 
>Again, what is the advantage of losing this information in this scenario? I only see dis-advantages.

The problem is, that this was never defined.
Is a communication interval even defined in the Modelica language, or is it just a concept that some Modelica tools have)?
In ME FMI this concept is at least not defined.

Perhaps FMI-LS-REF is the right place to clearly define what you want, perhaps add it the "experiment section" https://modelica.github.io/fmi-ls-ref/main/#_experiment_setup
If you agree, please open an issue there https://modelica.github.io/fmi-ls-ref/

chrbertsch:
> If you think this is worth discussing, I can open a separate issue for that.

@casella : Yes please to so.
The concept of a "time resolution for the result grid" does currently not exist in FMI for ME.
I still think, the right place is a layered standard, and not the core standard.

> I would just add an Interval field to the DefaultExperiment tag so we can preserve that information from Modelica models to FMI-MEs

Adding an additional attribute is not possible in a maintenance release FMI 3.0.x, and a minor relesae FMI 3.1 is not scheduled currently.  
This is why IMHO the rights short-term solution is a layered standard such as FMI-LS-BUS (https://github.com/modelica/fmi-ls-ref). There you can also ship the original Modelica model, from which the FMU was created, so the connection is obvious. And we can without dependence on the FMI release cycle add as many new arguments to the default experiment (https://modelica.github.io/fmi-ls-ref/main/#_experiment_setup) as we want.

Tc-Fast:
@casella: I voted to state, that the stepSize has no meaning in ME.

If you consider an ode with complex eigenvalues at $\lambda_{1,2} = 25i$. Then you can not find a stepSize $h$, such that the simulation is stable with an euler or heun integrator. For an RK4 integrator on the other hand you can use a stepSiez $h=0.1$.

In my opinion specifying a stepSize $h$ without specifying an integrator method is therefor meaningless for ME.

PR #1986, opened 2024-10-22
Fixed get/set fractional variable naming for clocks

### LINKED ISSUE #1985: Incorrect variable naming for fmi3GetIntervalFraction, fmi3SetIntervalFraction, fmi3GetShiftFraction and fmi3SetShiftFraction
- The functions fmi3GetIntervalFraction, fmi3SetIntervalFraction, fmi3GetShiftFraction and fmi3SetShiftFraction use arguments named "counter" and "resolution". This is incomprehensible. The variable names should be "intervalNumerators", "intervalDenominators", "shiftNumerators" and "shiftDenominators", because these names are commonly used in the context of fractions.
- The variable "counter" is not a "counter". I guess this name is the result of a translation from german to english ("Zähler" -> "Counter"), but in the context of fraction it should translate to "numerator".
- The documentation states about the "counter" variable: "Note: This variable may increment by values other than 1."
This is wrong, since the "counter" is not a "counter". It is never incremented or decremented.
- The variable "resolution" is misleading as well. A fraction does not have a "resolution".
- Some subscripts in the mathematical notation use numbers in italic font. In mathematical notation numbers should always use a "normal" font.

In order to maintain backwards compatibility the naming in the xsd files can not be changed. I suggest to leaf the names in the xsd therefor as is and add a comment for clearification.

Tc-Fast:
closes #1985

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -657,7 +657,7 @@ latexmath:[T_{\mathit{interval, i}}] can be set with <<fmi3SetIntervalDecimal>> 
 
 Periodic Clock::
 is a time-based Clock with a constant interval, except when <<intervalVariability>> = <<tunable>>, which indicates that the interval can change when tunable <<parameter, parameters>> change.
-The time instant of the first Clock tick is defined by a <<shiftCounter>> or <<shiftDecimal>>.
+The time instant of the first Clock tick is defined by a decimal (<<shiftDecimal>>) and optionally by a fraction (<<shiftNumerator>>, <<shiftDenominator>>).
 +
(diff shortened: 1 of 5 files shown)


chrbertsch (commented):
I am all for better explanations but I am reluctant changing variable names in a maintenance release if it not absolutely necessary, as this will create confusion.

otronarp:
Also the currently used terminology (of `counter` and `resolution`) aligns well with how clocks are described in the Modelica Language Specification.

chrbertsch:
See discussion and comments at FMI Design Meeting https://github.com/modelica/fmi-standard/issues/1985#issuecomment-2429560954

PR #1993, opened 2024-11-04
Added explanation to fmi3{Get|Set}{Interval|Shift}Fraction arguments

### LINKED ISSUE #1985: Incorrect variable naming for fmi3GetIntervalFraction, fmi3SetIntervalFraction, fmi3GetShiftFraction and fmi3SetShiftFraction

Tc-Fast:
Closes #1985

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -716,7 +716,7 @@ is a time-based Clock whose next interval is unchangeably known right after the 
 +
 The next Clock tick at time instant latexmath:[t_i] is defined as: +
 latexmath:[\begin{align*}
-t_0 &:= t_{\mathit{start}} + T_{\mathit{interval, 0}} \\
+t_0 &:= t_{\mathit{start}} + T_{\mathit{interval}, 0} \\
 t_i &:= t_{i-1} + T_{\mathit{interval, i}} \qquad i = 1,2,3,{...}
 \end{align*}]
 
@@ -726,7 +726,7 @@ t_i &:= t_{i-1} + T_{\mathit{interval, i}} \qquad i = 1,2,3,{...}
 [cols="1,2"]
 |===
 
-|latexmath:[T_{\mathit{interval, 0}} \geq 0]
+|latexmath:[T_{\mathit{interval}, 0} \geq 0]
 |The time interval from latexmath:[t_{\mathit{start}}] to first Clock tick.
 Must be retrieved using <<fmi3GetInterval>> in <<InitializationMode>>.
 
@@ -742,7 +742,7 @@ The return argument <<qualifiers>> of <<fmi3GetInterval>> is used to indicate if
 +
 The next Clock tick at time instant latexmath:[t_i] is defined as: +
 latexmath:[\begin{align*}
-t_0 &:= t_{\mathit{start}} + T_{\mathit{interval, 0}} \\
+t_0 &:= t_{\mathit{start}} + T_{\mathit{interval}, 0} \\
 t_i &:= t_{\mathit{event}} + T_{\mathit{interval, i}} \qquad i = 1,2,3,{...}
 \end{align*}]
 
@@ -752,7 +752,7 @@ t_i &:= t_{\mathit{event}} + T_{\mathit{interval, i}} \qquad i = 1,2,3,{...}
 [cols="1,2"]
 |===
 
-|latexmath:[T_{\mathit{interval, 0}} \geq 0]
+|latexmath:[T_{\mathit{interval}, 0} \geq 0]
 |The time interval from latexmath:[t_{\mathit{start}}] to the first Clock tick.
 Must be retrieved using <<fmi3GetInterval>> in <<InitializationMode>>.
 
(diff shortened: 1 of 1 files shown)


masoud-najafi (approved):

docs/2_2_common_mechanisms.adoc
```diff
@@ -834,6 +834,10 @@ For restrictions on when to call which of the following functions for <<table-ov
 
 <<fmi3SetClock>> must not be called in Scheduled Execution, instead <<fmi3ActivateModelPartition>> must be called.
 
+The Clocks API function argument names <<intervals>>, <<shifts>>, <<counters>> and <<resolutions>> are chosen to match the link:https://specification.modelica.org/maint/3.6/synchronous-language-elements.html#clock-constructors[Modelica Clock Constructors].
+However, the FMI standard does not assume that the importer tool's clock implementation matches the Modelica Clock implementation.
+In general, the arguments <<counters>> and <<resolutions>> have the meaning of a "numerator" and a "denominator".
```
chrbertsch, line 839:
```suggestion
_[The Clocks API function argument names <<intervals>>, <<shifts>>, <<counters>> and <<resolutions>> are chosen to match the link:https://specification.modelica.org/maint/3.6/synchronous-language-elements.html#clock-constructors[Modelica Clock Constructors]._
_However, the FMI standard does not assume that the importer tool's clock implementation matches the Modelica Clock implementation._
_In general, the arguments <<counters>> and <<resolutions>> have the meaning of a "numerator" and a "denominator".]_
```
The explanation should be non-normative.

docs/2_2_common_mechanisms.adoc
```diff
@@ -860,7 +864,7 @@ include::../headers/fmi3FunctionTypes.h[tag=SetIntervalFraction]
 * `valueReferences` is an array of size `nValueReferences` holding the value references of the Clock variables.
 
 [[counters,`counters`]]
-* <<counters>> _[Note: This variable may increment by values other than 1.]_ and
```
chrbertsch, line 863: Why remove this note?

clagms:
Not sure this PR is to blame, but currently the build is giving errors:
```
asciidoctor: WARNING: docs/2_4_common_schema.adoc: line 1966: id assigned to anchor already in use: reverse-DNS
asciidoctor: WARNING: docs/7___appendix.adoc: line 181: id assigned to anchor already in use: reverse-DNS
asciidoctor: ERROR: docs/7___appendix.adoc: line 259: dropping cells from incomplete row detected end of table
```

Should this be solved in a separate PR @chrbertsch ?

docs/2_2_common_mechanisms.adoc
Tc-Fast: This variable is never incremented. The implementation in the importer tool may increment some counter, but definitely not the "counter" argument of this function.
Furthermore there also may not be a counter in the importer tool implementation altogether. The FMI does not assume the implementation.

Tc-Fast:
@clagms These warnings are not related to this PR. In my opinion they should be fixed.

clagms (approved):

docs/2_2_common_mechanisms.adoc
otronarp: Of course it can be increment it, but it can also be change with any increment (or decrement). Isn't that what the note is there to clarify?

Tc-Fast:
```C
typedef fmi3Status fmi3SetIntervalFractionTYPE(fmi3Instance instance,
                                               const fmi3ValueReference valueReferences[],
                                               size_t nValueReferences,
                                               const fmi3UInt64 counters[],
                                               const fmi3UInt64 resolutions[]);
```
It is const

otronarp: That doesn't prevent the importing tool from changing it between calls to `fmiSetInteralFractionType` (if the `intervalVariability` is `tunable`).

docs/2_2_common_mechanisms.adoc
pmai:
```suggestion
_[The Clocks API function argument names <<intervals>>, <<shifts>>, <<counters>> and <<resolutions>> are chosen to match the naming used in link:https://specification.modelica.org/maint/3.6/synchronous-language-elements.html#clock-constructors[Modelica Clock Constructors]._
_This commonality of naming does not imply any normative reference to the semantics of Modelica._
_Mathematically speaking, the arguments <<counters>> and <<resolutions>> in the fraction variant of these functions serve as numerators and denominators of fractional representations of the corresponding floating-point number arguments of the decimal variant.]_
```

Maybe the final sentence should be split up a bit more, but didn't find the fortune to do so.

docs/2_2_common_mechanisms.adoc
pmai: Given that I think this note was added after people were confused on the rules,  I don't think removing it is a sensible way to go. We might want to adjust it going forward.

docs/2_2_common_mechanisms.adoc
```diff
@@ -860,7 +864,7 @@ include::../headers/fmi3FunctionTypes.h[tag=SetIntervalFraction]
...
+* <<counters>> and
```
pmai, line 867:
```suggestion
* <<counters>> _[Note: This variable may increment by values other than 1.]_ and
```

pmai (changes requested):
My try at avoiding misleading statements about implementations and capturing the current review comments. Feel free to further improve.

chrbertsch (approved):

PR #2005, opened 2024-11-13
Added non-normative explanation for usage of stepSize in ME

### LINKED ISSUE #1991: Introduce the Interval attribute in the DefaultExperiment element of modelDescription.xml
Follow up of the discussion in PR #1980.

FMI 2.0.4 Section 2.2.5 defines the `DefaultExperiment` element of the XML file as:
immagine

The interpretation of `stepSize` is different for FMI-CS, where it means "preferred communication step size", and for FMI-ME, where it means "Default step size for fixed step integrators". After #1939 it was decided to remove the interpretation for FMI-ME entirely.

FMI 3.0.1 [Section 2.4.6](https://fmi-standard.org/docs/3.0.1/#DefaultExperiment) gives the same definition for FMI-CS, while the meaning of `stepSize` is undefined for the other modes.

Although FMI models are not necessarily generated from Modelica models, it is well understood that semantics of Modelica model simulations should be completely captured by FMI. This is stated explicitly in FMI 2.0.4 Section 1.1:

> Expressivity: The FMI provides the necessary features that Modelica®, Simulink® and SIMPACK® models can be transformed to an FMU.

This is not stated explicitly in FMI 3.0.1 (why?) but I understand the rationale is the same, given that many concepts like Stream variables or Clocks are clearly built to mirror their Modelica counterparts. Furthermore, is is well understood that the goal of the Modelica Association is to promote open _coordinated_ standards for system simulation.

Regarding the default experiment settings, Modelica 3.6 defines the experiment annotation in [Section 18.4](https://specification.modelica.org/maint/3.6/annotations.html#annotations-for-simulations), with the following semantics:

> The experiment annotation defines the default start time (StartTime) in [s], the default stop time (StopTime) in [s], the suitable time resolution for the result grid (Interval) in [s], and the default relative integration tolerance (Tolerance) for simulation experiments to be carried out with the model or block at hand. If StartTime is not specified it is assumed to be 0.0.

The relationship between the `StartTime`, `StopTime` and `Tolerance` fields of this annotation and the `startTime`, `stopTime` and `tolerance` attributes of `DefaultExperiment` is obviously that they are identical. Unfortunately, for some reason there is no explicit counterpart in FMI for the `Interval` field.

The definition "suitable time resolution for the result grid" is not as rigorous as the definition of communication time step as defined by FMI, but it is obviously indicating the same concept. So, if a Modelica model is exported as a FMU-CS, the `stepSize` attribute of DefaultExperiment can be given the value of the experiment `Interval` annotation. Unfortunately, it is not possible to pass this information to an FMU-ME, so it is lost for the importing tool, which could instead use it to re-sample a variable step-size simulation on a regular time grid.

I think this should be fixed, either by adding an `interval` attribute to `DefaultExperiment`, or by defining the meaning of `stepSize` in FMI-ME to be the same of FMI-CS, i.e., the default communication interval.

This is of course a minor issue, but yet I see no reason why this piece of information, provided by Modelica since version 3.3 in 2012, should be lost. 

I guess the reason it is missing from the FMI-ME `DefaultExperiment` is that FMI-ME 1.0 was released in 2010, when the Modelica experiment annotation only had  `StartTime`, `StopTime` and `Tolerance`. Then, the definition was carried over to newer versions, while probably nobody in the FMI working group noticed that the Interval field was added to the experiment annotation two years later.

To me, this is obviously a minor oversight that could happily be corrected in the 2.0.5 and 3.0.2 patch releases. If that is not possible, then it should definitely be fixed in 3.1.0.

In #1980, @chrbertsch proposes to add this extra attribute to the layered FMI-LS-REF standard. I think we can definitely do that, but yet that layered standard does not define a _default_ experiment, which what the DefaultExperiment element of the main standard XML does. So I think this issue should (also) be fixed here.

Tc-Fast:
Closes #1991

### DIFF
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -813,7 +813,11 @@ A tool may ignore this information. +
 _[However, it is convenient for a user that <<startTime>>, <<stopTime>>, <<tolerance>> and <<stepSize>> have already a meaningful default value for the model at hand.]_ +
 <<startTime>>, <<stopTime>> and <<stepSize>> refer to the unit of the <<independent>> variable.
 <<stepSize>> defines the preferred <<communicationStepSize>> for Co-Simulation.
-For Model Exchange and Scheduled Execution <<stepSize>> has no defined meaning.
+For Model Exchange and Scheduled Execution <<stepSize>> has no defined meaning. 
+_[This does not prohibit the use of <<stepSize>> in the context of Model Exchange._
+_For example, a fixed step size solver can use it as a default step size._
+_A variable step size solver can use the <<stepSize>> as an initial step size or as result grid for the output._ 
+_Since the required <<stepSize>> depends on the solver itself a defined meaning of <<stepSize>> for Model Exchange is impossible.]_
 
 [[ModelVariables,`<ModelVariables>`]]
 ==== Model Variables


docs/2_4_common_schema.adoc
```diff
@@ -813,7 +813,9 @@ A tool may ignore this information. +
 _[However, it is convenient for a user that <<startTime>>, <<stopTime>>, <<tolerance>> and <<stepSize>> have already a meaningful default value for the model at hand.]_ +
 <<startTime>>, <<stopTime>> and <<stepSize>> refer to the unit of the <<independent>> variable.
 <<stepSize>> defines the preferred <<communicationStepSize>> for Co-Simulation.
-For Model Exchange and Scheduled Execution <<stepSize>> has no defined meaning.
+For Model Exchange and Scheduled Execution <<stepSize>> has no defined meaning. 
+_[This does not prohibit the use of <<stepSize>> in the context of Model Exchange. 
+For example, a fixed step size solver can use it as a default step size, but since the required <<stepSize>> depends on the solver itself a defined meaning of <<stepSize>> for Model Exchange is impossible.]_
```
chrbertsch, line 818:
```suggestion
For Model Exchange and Scheduled Execution <<stepSize>> has no defined meaning. 
_[This does not prohibit the use of <<stepSize>> in the context of Model Exchange._ 
_For example, a fixed step size solver can use it as a default step size, but since the required <<stepSize>> depends on the solver itself a defined meaning of <<stepSize>> for Model Exchange is impossible._
Or it could be used as a result grid to be used for the outputs in case of a variable step solver]_
```

chrbertsch (approved):

t-sommer:
@Tc-Fast, please follow the [contributing guide](https://github.com/modelica/fmi-standard/blob/main/CONTRIBUTING.adoc#commit-messages) and use imperative mood for commit messages. Thanks!

t-sommer (approved):

PR #2038, opened 2025-05-15
Add Power PC archs

### LINKED ISSUE #2037: <arch> names
Are the <arch> names mentioned in https://fmi-standard.org/docs/3.0.2/#platform-tuple-definition  standardized and can we include a reference?
Should be ppc/ppc64 for power PC, right?
Question came up in  https://github.com/casadi/casadi/pull/4113

chrbertsch:
resolves #2037

### DIFF
--- a/docs/2_5_fmu_distribution.adoc
+++ b/docs/2_5_fmu_distribution.adoc
@@ -118,6 +118,12 @@ Architecture `<arch>`::
 
 |x86_64
 |Intel/AMD x86 64-bit
+
+|ppc
+|PowerPC 32-bit
+
+|ppc64
+|PowerPC 64-bit
 |====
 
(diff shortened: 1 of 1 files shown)


docs/2_5_fmu_distribution.adoc
```diff
@@ -118,6 +118,12 @@ Architecture `<arch>`::
 
 |x86_64
 |Intel/AMD x86 64-bit
+
+|ppc
```
pmai, line 122:
```suggestion
|ppc32
```
Let's follow clang and add the 32bit suffix consistently.

pmai (changes requested):
Looks good with the suggestion to use ppc32 for 32bit which matches CLANG...

pmai (approved):

PR #2050, opened 2025-06-06
Improve the specification of fmi3GetInterval and interval qualifiers

### LINKED ISSUE #2018: Clarify rules on fmi3GetInterval*/fmi3GetShift* functions with regards to periodic input clocks
The standard seems to suggest that these functions can be called for clocks with constant/fixed/tunable intervalVariability - at least under some circumstances - in Table 5 by listing as related API:

- constant:
  > fmi3SetClock in EventMode, fmi3ActivateModelPartition in ClockActivationMode, fmi3GetInterval and fmi3GetShift in InitializationMode

However for fixed/tunable it lists:

- fixed:
  > fmi3SetClock in EventMode, fmi3ActivateModelPartition in ClockActivationMode,
  >
  > * if intervalDecimal is defined, fmi3SetInterval in InitializationMode,
  > * if intervalDecimal is not defined, fmi3GetInterval and fmi3GetShift in InitializationMode

So interestingly, for fixed/tunable you can only call fmi3GetInterval/fmi3GetShift if no interval is provided in the modelDescription.xml.

While Table 6 then just generally states on T_interval_i:

> The time interval until the next Clock tick, defined differently for the different Clock types. Tinterval,i
 can be set with fmi3SetIntervalDecimal or retrieved with [fmi3GetIntervalDecimal](https://fmi-standard.org/docs/3.0.2/#fmi3GetIntervalDecimal) as floating point value, or as rational number using [fmi3SetIntervalFraction](https://fmi-standard.org/docs/3.0.2/#fmi3SetIntervalFraction) or [fmi3GetIntervalFraction](https://fmi-standard.org/docs/3.0.2/#fmi3GetIntervalFraction).

which might or might not suggest that fmi3GetInterval* can always be called.

Furthermore the definition of fmi3GetInterval* then provides wording that does not suggest anyone thought of the semantics of calling them for periodic clocks:

> * qualifiers is an array of size `nValueReferences` to retrieve the Clocks qualifiers.
> qualifiers describes how to treat the intervals and counters arguments and is defined as:

> ...

> with the following meanings:

> * `fmi3IntervalChanged` is returned to indicate that the value for the interval has changed for this Clock.
> Any previously returned intervals (if any) are overwritten with the current value.
> The new Clock interval is relative to the time of the current EventMode or ClockUpdateMode in contrast to the interval of a periodic Clock, where the interval is defined as the time between consecutive Clock ticks.
> In Scheduled Execution this means that the corresponding model partition must be scheduled or re-scheduled (if a previous call to fmi3GetInterval returned fmi3IntervalChanged).

>  * `fmi3IntervalUnchanged` is returned if a previous call to fmi3GetInterval already returned a value qualified with fmi3IntervalChanged which has not changed since.
>  In Scheduled Execution this means the corresponding model partition has already been scheduled.

>  * `fmi3IntervalNotYetKnown` is returned for a countdown-aperiodic-clock,`countdown` Clock for which the next interval is not yet known.
> This qualifier value can only be returned directly after the Clock was active and previous calls to fmi3GetInterval never returned fmi3IntervalChanged (nor fmi3IntervalUnchanged).
> In Scheduled Execution this return value means that the corresponding model partition cannot be scheduled yet.

None of the options provided seem to take constant (or even just periodic) clocks into account, and hence it is unclear what should be returned:

-	fmi3IntervalChanged seems to assume we are in Event Mode, but the calls allowed are in Initialization Mode, which is not Event Mode, and it also suggests the intervals are relative to some current time, which again they should not be for periodic clocks.
-	fmi3IntervalUnchanged is only allowed if fmi3IntervalChanged has been reported once, so no go for periodic/constant clocks.
-	fmi3IntervalNotYetKnown is only available for countdown clocks when the next interval is not yet known.

Taken together I think the description of this in the standard should be refined to precisely state whether and when it is allowed to call fmi3GetInterval* and what should be returned for periodic clocks.

PTaeuberDS:
First draft to resolve #2018

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -550,6 +550,8 @@ is a variable of type Clock with <<causality>> = <<output>>.
 The attribute <<intervalVariability>> must be <<triggered>> for <<outputClock,output Clocks>>.
 The importer calls <<fmi3GetClock>> to inquire the Clock's activation state.
 
+*#ToDo#* Update table
+
 .Overview of Clock types.
 [[table-overview-clocks]]
 [cols="9,9,16,9,40,40"]
@@ -656,7 +658,9 @@ latexmath:[T_{\mathit{interval, i}}] can be set with <<fmi3SetIntervalDecimal>> 
 |===
 
 Periodic Clock::
-is a time-based Clock with a constant interval, except when <<intervalVariability>> = <<tunable>>, which indicates that the interval can change when tunable <<parameter, parameters>> change.
+*#ToDo#* Add sentence that interval is defined differently for tunable Clocks (sometimes - when the interval changes)
+
+is a time-based Clock with a constant interval, except when <<intervalVariability>> = <<tunable>>, which indicates that the interval can change.
 The time instant of the first Clock tick is defined by a <<shiftCounter>> or <<shiftDecimal>>.
 +
 The next Clock tick at time instant latexmath:[t_i] is defined as: +
@@ -882,6 +886,14 @@ include::../headers/fmi3FunctionTypes.h[tag=GetIntervalDecimal]
 
 * <<intervals>> is an array of size `nValueReferences` to retrieve the Clock intervals.
 
+*#ToDo#* Explain the meaning of "with specified interval" and link to the respective section
+
+*#ToDo#* State that getInterval can be called for all Clock types (incl. settable)
+
+*#ToDo#* Check also the definition of GetShift
+
+*#ToDo#* When it is possible to have a new interval for tunable Clocks, should it also be possible to have new shifts?
+
 [[qualifiers,`qualifiers`]]
 * <<qualifiers>> is an array of size `nValueReferences` to retrieve the Clocks <<qualifiers>>.
 <<qualifiers>> describes how to treat the <<intervals>> and <<counters>> arguments and is defined as:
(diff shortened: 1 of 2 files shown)


docs/2_2_common_mechanisms.adoc
```diff
@@ -672,14 +682,18 @@ t_i &:= t_{i-1} + T_{\mathit{interval, i}} \qquad i = 1,2,3,{...}
 |===
 
 |latexmath:[T_{\mathit{interval, i}} > 0]
-|The time interval from the previous Clock tick to the current Clock tick, specified differently for the different Clock types.
+|Generally, the time interval from the previous Clock tick to the current Clock tick.
+ An exception applies to a tunable Clock whose interval has just changed in <<EventMode>> or <<ClockActivationMode>>.
+ In this specific case, the new interval is relative to the time of the current <<EventMode>> or <<ClockActivationMode>>, not to the time of the previous Clock tick.
+ The interval is provided differently for the different Clock types.
```
andreas-junghanns, line 688: Maybe you can add a link to where that is described?

docs/2_2_common_mechanisms.adoc
```diff
@@ -672,14 +682,18 @@ t_i &:= t_{i-1} + T_{\mathit{interval, i}} \qquad i = 1,2,3,{...}
...
+
 |===
 
 Constant periodic Clock::
 is a time-based periodic clock that defines its <<intervalDecimal>> and optionally <<intervalCounter>>, <<shiftCounter>> and <<shiftDecimal>> in the <<modelDescription.xml>>.
 
-Fixed periodic Clock::
-is a time-based periodic Clock which ticks with an arbitrary, but fixed interval latexmath:[T_{\mathit{interval}}] starting after an arbitrary latexmath:[T_{\mathit{shift}}], both interval and shift can be changed by the importer.
```
andreas-junghanns, line 682: why drop this part of the sentence?
I think we added this to make sure implementers know that the importer can adjust the actual intervals according to its own needs, like a specific, fixed, step size.

docs/2_2_common_mechanisms.adoc
```diff
@@ -893,19 +912,134 @@ include::../headers/fmi3FunctionTypes.h[tag=IntervalQualifier]
...
+  In Scheduled Execution this return value means that the corresponding model partition cannot be scheduled yet.
+
+<<table-qualifiers-for-time-based-clocks>> shows the allowed qualifiers for different time-based Clock types.
+For some combinations the FMU state is relevant.
+If a cell has two values, the left value refers to <<InitializationMode>>, whereas the right value refers to <<EventMode>>, <<ClockActivationMode>>, or <<ClockUpdateMode>>.
+
+The categorization of fixed and tunable Clocks as "_intervalDecimal specified_" and "_intervalDecimal not specified_" refers to the presence or absence of the <<intervalDecimal>> attribute on the Clock element in the model description.
+For a detailed explanation, refer to the definitions of <<fixed-periodic-clock,fixed Clocks>> and <<tunable-periodic-clock,tunable Clocks>>.
+
+.Allowed qualifiers for time-based Clocks.
```
andreas-junghanns, line 939: To validate this table, I would need a second brain. Can we schedule a meeting?

chrbertsch: > No, you are right. This is a bug/contradiction in the current specification. This needs to be fixed independently from the outcome of this PR.

@masoud-najafi , @PTaeuberDS : can you please create a separate issue / pull request for that? Thanks!

docs/2_2_common_mechanisms.adoc
```diff
@@ -672,14 +682,17 @@ t_i &:= t_{i-1} + T_{\mathit{interval, i}} \qquad i = 1,2,3,{...}
 |===
 
 |latexmath:[T_{\mathit{interval, i}} > 0]
-|The time interval from the previous Clock tick to the current Clock tick, specified differently for the different Clock types.
+|Generally, the time interval from the previous Clock tick to the current Clock tick.
+ An exception applies to a tunable Clock whose interval has just changed in <<EventMode>> or <<ClockActivationMode>>.
+ In this specific case, the new interval is relative to the time of the current <<EventMode>> or <<ClockActivationMode>>, not to the time of the previous Clock tick.
+
```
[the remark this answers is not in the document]
PTaeuberDS: No, it wouldn't. The interval of a tunable clock can change at any (event) time without the clock having ticked. So there is not necessarily a current clock tick.

docs/2_2_common_mechanisms.adoc
PTaeuberDS: Done, s. #2068

docs/2_2_common_mechanisms.adoc
```diff
@@ -707,6 +720,9 @@ If <<intervalDecimal,`intervalDecimal`>> is not specified, then the importer mus
 The shift may only depend on <<fixed>> <<parameter, parameters>>.
 The importer must use <<fmi3GetShift>> to retrieve the Clock shift in <<InitializationMode>>.
 
++
+In both cases, when an interval changes, the new interval is relative to the time of the current <<EventMode>> or <<ClockActivationMode>>, not to the time of the previous Clock tick.
```
masoud-najafi: I'm not sure if I understand this correctly. Suppose the previous clock tick has been at time=T_k and the next tick will be at time=T_{k+1}=T_K+DT, where DT is the current interval of the clock. If an event happens at T_e  where  T_K< T_e <T_{k+1}, and at this time event the interval of the clock is changed by the importer to DT2, what would be the time of the next clock tick?

clagms (commented):
I think overall this is improving the standard. I could not spot any major issues except the possible ambiguous use of "when". See the comments below.

PR #2051, opened 2025-06-10
Convert GitHub related documents to Markdown

t-sommer:
and add logo for dark mode


chrbertsch (commented):
Looks good. 
But why do we need the CCLA as Markdown anyway? On https://fmi-standard.org/about/ we link to the CCLA PDF and  would avoid maintaining different versions of the CCLA. It should be replaced by the MA CLA anyway at some point of time)

chrbertsch (approved):

PR #2080, opened 2025-11-12
Clarify glossary entries for "simulator"

### LINKED ISSUE #1921: Glossary clarificaitons for "simulator - simulation program - simulation tool"
We were contacted by  a user via contact@fmi-standard.org with the following question:

"... The following terms are defined in the glossary: 
**simulation program**: Software to develop and/or solve simulation models. The software includes a solver, may include a user interface and methods for post processing (see also: simulation tool, simulation environment).
**simulator**: A simulator can include one or more simulation programs.
_I am confused about this because the two terms are sometimes used interchangeably in the specifications. This can also be seen in the definition of co-simulation in the FMI-specification, which is defined via simulation tools but referred to synonymously with simulator coupling:_
**Co-simulation:** Coupling of several simulation programs in order to compute the global behaviour of a system that consists of several subsystems. The subsystems are coupled in the sense that the behaviour of each subsystem depends on the behaviour of the remaining subsystems, so that the co-simulation must be computed in a step-by-step fashion. Each simulation program is responsible for computing the behaviour of a subsystem, using the outputs produced by the other simulation programs. Synonyms: dynamic mutual-exchange, simulator coupling, and coupled simulation.
_Without demanding a universal definition, as a newcomer to the field I would be interested in the difference between simulator and simulation tool. If possible, an example would be helpful. Referring to [Palensky 2017]1 (simulator) or [Gomes 2018]2 (simulation unit) consists of the model and the solver. Intuitively, I would equate "simulation tool" with "simulator" according to the above definition."_

While I do not fully understand the problem described above, I think there are some things to clarify

* we should distinguish between "Co-simulation" and "simulator-coupling"/"tool coupling". To my understanding, simulator-/tool-coupling is a special form of co-simulation, where the original simulation tools ("simulators") and are involved (typically also with their GUI), while ca co-simulation could also be done using CS FMUs that had been exported from the original simulations tools.
* "simulation program" is not used outside the glossary, and it is defined  as a synonym to "simulation tool".  Do we need this definition at all?  
* I do not see the difference between "simulation tool" and "simulator" and would use them as synonyms. What would be a simulator, that includes several simulation programs? (A co-simulation platforms such as Model.Connect? Then we need to define this)

Tc-Fast:
Closes #1921

### DIFF
--- a/docs/1___overview.adoc
+++ b/docs/1___overview.adoc
@@ -63,7 +63,7 @@ Clarifications and fixes of all FMI 3.0 patch releases (FMI 3.0.x) are listed in
 === Overview
 
 The Functional Mock-up Interface (FMI) defines a ZIP archive and an application programming interface (API) to exchange dynamic models using a combination of XML files, binaries and C code: the Functional Mock-up Unit (FMU).
-The API is used by a simulation environment, the importer, to create one or more instances of an FMU and to simulate them, typically together with other models.
+The API is used by a simulator, the importer, to create one or more instances of an FMU and to simulate them, typically together with other models.
 The FMI defines three interface types:
 
 * <<fmi-for-co-simulation,Co-Simulation (CS)>> where the FMU typically contains its own solver or scheduler,
@@ -107,7 +107,7 @@ image::images/model-exchange-data-flow.svg[width=40%, align="center"]
 
 ==== FMI for Co-Simulation (CS)
 
-The Co-Simulation interface is designed both for the coupling of simulation tools, and the coupling of subsystem models, exported by a modeling environment together with their solvers as runnable code.
+The Co-Simulation interface is designed both for the coupling of _simulators_, and the coupling of subsystem models, exported by a modeling environment together with their solvers as runnable code.
 (See <<fmi-for-co-simulation>>.)
 
 .Schematic view of data flow between user, the co-simulation algorithm of the importer and the FMU for Co-Simulation
@@ -197,7 +197,7 @@ Expressivity::
 The FMI provides the necessary features to package models of different domains, such as multi-body and virtual ECUs, into an FMU.
 
 Stability::
-The FMI is expected to be supported by many simulation tools worldwide.
+The FMI is expected to be supported by many _simulators_ worldwide.
 Implementing such support is a major investment for tool vendors.
 Stability and backwards compatibility of the FMI has therefore high priority.
 
--- a/docs/2_3_common_states.adoc
+++ b/docs/2_3_common_states.adoc
@@ -92,7 +92,7 @@ For FMUs that have clocks, `eventModeUsed = fmi3True` is required.
 * [[earlyReturnAllowed,`earlyReturnAllowed`]]If `earlyReturnAllowed = fmi3True` the importer can handle <<early-return,early return>>.
 Only in this case, <<fmi3DoStep>> may return with <<earlyReturn,`earlyReturn = fmi3True`>>.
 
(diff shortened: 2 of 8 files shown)


docs/7___appendix.adoc
```diff
@@ -32,10 +32,10 @@ A <<Clock>> tick causes a <<Clock>> activation at that time instant, except for
 |See <<time-instant>>.
 
 |_co-simulation_
-|Coupling of several _simulation programs_ in order to compute the global behavior of a system that consists of several subsystems.
+|Coupling of several _simulators_ in order to compute the global behavior of a system that consists of several subsystems.
```
chrbertsch: Here we we could lose the distinction between "co-simulation" (as a more general concept, could by the co-simulation of multiple stand-alone FMUs) and "simulator coupling" / "tool coupling"  (requiring the installation of different simulation tools and the co-simulation of "tool coupling FMUs)

docs/1___overview.adoc
```diff
@@ -197,7 +197,7 @@ Expressivity::
 The FMI provides the necessary features to package models of different domains, such as multi-body and virtual ECUs, into an FMU.
 
 Stability::
-The FMI is expected to be supported by many simulation tools worldwide.
+The FMI is expected to be supported by many _simulators_ worldwide.
```
chrbertsch:
```suggestion
The FMI is expected to be supported by many _importers_ worldwide.
```

docs/2_3_common_states.adoc
```diff
@@ -92,7 +92,7 @@ For FMUs that have clocks, `eventModeUsed = fmi3True` is required.
 * [[earlyReturnAllowed,`earlyReturnAllowed`]]If `earlyReturnAllowed = fmi3True` the importer can handle <<early-return,early return>>.
 Only in this case, <<fmi3DoStep>> may return with <<earlyReturn,`earlyReturn = fmi3True`>>.
 
-* `instanceEnvironment` is a pointer that must be passed to <<fmi3IntermediateUpdateCallback>>, <<fmi3ClockUpdateCallback>>, and <<logMessage,`fmi3LogMessageCallback`>> to allow the simulation environment an efficient way to identify the calling FMU.
+* `instanceEnvironment` is a pointer that must be passed to <<fmi3IntermediateUpdateCallback>>, <<fmi3ClockUpdateCallback>>, and <<logMessage,`fmi3LogMessageCallback`>> to allow the simulator an efficient way to identify the calling FMU.
```
chrbertsch:
```suggestion
* `instanceEnvironment` is a pointer that must be passed to <<fmi3IntermediateUpdateCallback>>, <<fmi3ClockUpdateCallback>>, and <<logMessage,`fmi3LogMessageCallback`>> to allow the importer an efficient way to identify the calling FMU.
```

docs/3_3_model_exchange_example.adoc
```diff
@@ -1,7 +1,7 @@
 === Code Example
 
-In the following example, the usage of the `fmi3XXX` functions is sketched in order to clarify the typical calling sequence of the functions in a simulation environment.
-Furthermore, it is assumed that one FMU is directly integrated in a simulation environment.
+In the following example, the usage of the `fmi3XXX` functions is sketched in order to clarify the typical calling sequence of the functions in a simulator.
+Furthermore, it is assumed that one FMU is directly integrated in a simulator.
```
chrbertsch:
```suggestion
Furthermore, it is assumed that one FMU is directly integrated in an importer.
```

docs/7___appendix.adoc
chrbertsch:
```suggestion
|Coupling of several _subsystem models_ or _simulators_ in order to compute the global behavior of a system that consists of several subsystems.
```

PR #2088, opened 2025-12-04
Clarify that calls to logMessage must respect log settings

### LINKED ISSUE #2072: Logging of errors when instantiate is called with loggingOn = fmi3False
FMI 3.0 reference FMUs log error messages even if `loggingOn==fmi3False` is given to the instantiate function:
Corresponding code fragment in cosimulation.c:
```
        comp->logEvents            = loggingOn;
        comp->logErrors            = true; // always log errors
```

However, the FMI 3.0 standard says:
> If loggingOn = fmi3False, then any logging is disabled and the [logMessage](https://fmi-standard.org/docs/3.0.2/#logMessage) callback function 
must not be called by the FMU.

I personally prefer that FMUs always report log messages when errors occur and would even prefer to weaken the sentence in the standard but we should deal with this inconsistency.

pmai:
Closes #2072 for the standard side of things.

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -32,7 +32,7 @@ _Defensive implementations should still guard against `NULL` pointers.]_
 
 * The FMI Standard does not provide a runtime platform or portability layer.
 Access to operating system resources and services, such as memory, network or file system, should be implemented with special care because the availability of such resources and services is not guaranteed on every target platform and/or simulator.
-If some resource is required by the FMU but is not available, the FMU must log what resource failed and return with error.
+If some resource is required by the FMU but is not available, the FMU must log what resource failed, respecting the current logging settings, and return with error.
 
 ==== Header Files and Naming of Functions [[header-files-and-naming-of-functions]]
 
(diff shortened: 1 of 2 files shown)


chrbertsch (approved):

chrbertsch (changes requested):
FMI Design Webmeeting: 

Klaus: This does not reflect the notes https://github.com/modelica/fmi-standard/issues/2072#issuecomment-3548403880. Was there another conclusion at the on-site meeting that we did not document.
As stated in the original issue I am not happy with the proposed solution.,
Masoud: I agree with Klaus.
Torsten: Me, too. The way the reference FMUs implement it, is the way to go.
Changing the standard would not break anything. It is a bugfix.
Nikolai: do the log categories "error" and "fatal" not be come useless?
Klaus, Torsten: Almost no FMUs use these categories. This is just a suggestion.
Klaus, Torsten: There is in the standard a mixupt of status and logCategories for errors ...
Torsten. Klaus: We should admit that we made a mistake and fix this "bug" the standard.
Torsten: We would legalize most implementation.

@pmai : could you please give your rationale?

PR #2097, opened 2026-01-05
Make error logging mandatory

### LINKED ISSUE #2072: Logging of errors when instantiate is called with loggingOn = fmi3False

Tc-Fast:
Alternative to #2088.
Closes #2072.

Error logging is now mandatory.
logStatusError and logStatusFatal are obsolete now.

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -215,9 +215,24 @@ The following listing shows the base types used in the FMI C-API:
 include::../headers/fmi3PlatformTypes.h[tags=VariableTypes]
 ----
 
-==== Status Returned by Functions [[status-returned-by-functions]]
+==== Status Returned by Functions and Logging Mechanism [[status-returned-by-functions]]
 
-This section defines the return values the C-API functions indicating success or failure of the function call.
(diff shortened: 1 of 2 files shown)


pmai:
> We will continue in https://github.com/modelica/fmi-standard/pull/2088/files

Actually I thought we agreed to stay here in #2097?, as that is closer to the intent?

chrbertsch:
> > Actually I thought we agreed to stay here in #2097?, as that is closer to the intent?
> 
> That's what I thought as well. I was wondering if I misunderstood it, but since you also think #2097 is what we agreed on, Christian maybe misunderstood it.

OK, if you agree to continue here it was my misunderstanding. I will close  https://github.com/modelica/fmi-standard/pull/2088/files then instead.

PR #2102, opened 2026-01-15
Avoid deprecation warnings on cmake min version

pmai:
Avoid deprecation warnings on newer cmake versions by requiring at least 3.10 which is the defacto base for modern cmake semantics going forward.

### DIFF
--- a/docs/examples/CMakeLists.txt
+++ b/docs/examples/CMakeLists.txt
@@ -1,6 +1,6 @@
 # configuration to build the code examples
 
-cmake_minimum_required (VERSION 3.7)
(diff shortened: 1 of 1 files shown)


docs/examples/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # configuration to build the code examples
 
-cmake_minimum_required (VERSION 3.7)
+cmake_minimum_required (VERSION 3.10)
```
beutlich:
```suggestion
cmake_minimum_required (VERSION 3.7...3.31)
```

Same result (no deprecation warning), without changing the min. required version.

PR #2112, opened 2026-02-03
Add missing abbrivations

Tc-Fast:
Since we changed the glossary, some links are broken. They should be fixed now or be removed.
Also I found some more abbreviations, not listed in the abbreviations table.
I linked all abbreviations in the documentation text (except the very common ones like "FMU" and "FMI").
I linked many references to the glossary (except the very common ones like "importer", etc.).

### DIFF
--- a/docs/7___appendix.adoc
+++ b/docs/7___appendix.adoc
@@ -8,8 +8,17 @@
 |Acronym
 |Meaning
 
+|ABI
+|Application Binary Interface
+
+|ANSI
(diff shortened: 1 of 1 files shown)


chrbertsch (approved):

t-sommer (changes requested):
I think the proposed use of xrefs is too excessive. It makes the text hard to read if words like XML, API, ZIP, etc. that the average reader is familiar with are turned into hyperlinks.

PR #2113, opened 2026-02-03
Remove event mode from common state machine

### LINKED ISSUE #2105: Figure 7 in section 2.3 does not show a proper abstraction of the three specialized state machines for the three standard variants.
Suggestion: Only display the Initialized super-state and let the transitions go into it so the refinements later can specify the concrete states and transitions valid in each variant.

Tc-Fast:
Closes #2105


chrbertsch (approved):
For me this change looks good.
It would be good to get the the insights of the original creators of the state machine image (@andreas-junghanns , @TorstenBlochwitz , @t-sommer ? )

chrbertsch:
Thanks for the update, @Tc-Fast.
However one of the stating points of this discussion was the missing arrow to event mode ...
Should we add dashed arrows as depicted here: 
grafik

PR #2128, opened 2026-02-20
Prune trailing spaces; replace tabs with spaces

Tc-Fast:


chrbertsch (approved):

Tc-Fast:
Added a check for trailing spaces. This check currently excludes all *.svg files, since they are generated/exported by some tool.

chrbertsch:
FMI Design webmeeting: 
instead to --exclude="*.svg" , explicitely state to which files it shall be applied

christian-wolf-eks:
Does it make sense to check for trailing spaces in MD as well but allow for exactly two trailing spaces (for newlines)?

Tc-Fast:
> Does it make sense to check for trailing spaces in MD as well but allow for exactly two trailing spaces (for newlines)?

I change the double quotes (-> line breaks) to empty new lines (-> paragraph break), so i can check markdown for trailing space as well. Vertical spacing is a litle different now, but I don't think you can tell.

Instead of having an exclude list, there is now an include list for the trailing space check.

PR #2129, opened 2026-02-20
Clarify fmiVersion attribute

### LINKED ISSUE #2109: What is the fmiVersion attribute?
We currently have a contradiction regarding the format of the fmiVersion in the modelDescription.xml.

From fmi3ModelDescription.xsd
```XSD
<xs:pattern value="3[.](0|[1-9][0-9]*)([.](0|[1-9][0-9]*))?(-.+)?"/>
```

So it allows e.g.
3.0; 3.0.1; 3.1.424242 etc.

The documentation says in Table 12:

> Version of FMI the XML file complies with. The value for this version is 3.0. Future minor revisions will be denoted as 3.1, 3.2 …​
> 
> [During development prototype FMU implementations can indicate compliance with a certain development version based on the tags available at https://github.com/modelica/fmi-standard/tags. For example the value for the FMI 3.0 Alpha 2 release is 3.0-alpha.2.]

Basically it wants to skip patch versions like 3.0.3 and instead should be 3.0.

This was at some point discussed here with no conclusion:
https://github.com/modelica/fmi-standard/pull/1765

In FMI 2.0 it was:
Version 2.0.5 Page 19
> Returns the version of the “fmi2Functions.h” header file which was used to compile the functions of the FMU. The function returns “fmiVersion” which is defined in this header file. The standard header file as documented in this specification has version “2.0” (so this function usually returns “2.0”).

Version 2.0.5 Page 34
> Version of “FMI for Model Exchange or Co-Simulation” that was used
to generate the XML file. The value for this version is “2.0”.
[Clarification for FMI 2.0.1: The version string for FMUs generated
according to FMI 2.0.1 shall be “2.0” to be compatible with FMI 2.0
importers.]

The fmi2ModelDescription.xsd:
```XSD
<xs:attribute name="fmiVersion" type="xs:normalizedString" use="required" fixed="2.0">
	<xs:annotation>
		<xs:documentation>Version of FMI (Clarification for FMI 2.0.2: for FMI 2.0.x revisions fmiVersion is defined as "2.0").</xs:documentation>
	</xs:annotation>
</xs:attribute>
```

I think the regex in the schema file is wrong. I think the point of using 2.0 instead of 2.0.5 was that all 2.0.x versions are compatible and same holds true for fmi3. For backwards compatibility it should have been the same for fmi3.

Anyway the contradiction needs to be resolved.

Tc-Fast:
Closes #2109

### DIFF
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -223,11 +223,13 @@ The XML attributes of `<fmiModelDescription>` are:
 
 |`fmiVersion`
 |Version of FMI the XML file complies with.
-The value for this version is `3.0`.
-Future minor revisions will be denoted as `3.1`, `3.2` ...
+The format is `3.MINOR` or `3.MINOR.PATCH` where MINOR and PATCH are non-negative integers without leading zeros.
+The current release version is `3.0`.
+Future minor revisions will be denoted as `3.1`, `3.2`, etc.
+Patch releases may be denoted as `3.0.1`, `3.1.2`, etc.
 
-NOTE: During development prototype FMU implementations can indicate compliance with a certain development version based on the tags available at https://github.com/modelica/fmi-standard/tags.
-For example the value for the FMI 3.0 Alpha 2 release is `3.0-alpha.2`.
+NOTE: During development, prototype FMU implementations can indicate compliance with a certain development version by appending a pre-release suffix, such as `3.0-alpha.2`, `3.1-beta.1`, or `3.0.1-rc.1`.
+These development versions are based on tags available at https://github.com/modelica/fmi-standard/tags.
 
 |`modelName`
 |The name of the model as used in the modeling environment that generated the XML file, such as `Modelica.Mechanics.Rotational.Examples.CoupledClutches`.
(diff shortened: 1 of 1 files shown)


chrbertsch (approved):
Looks good to me

docs/2_4_common_schema.adoc
```diff
@@ -223,11 +223,13 @@ The XML attributes of `<fmiModelDescription>` are:
 
 |`fmiVersion`
 |Version of FMI the XML file complies with.
-The value for this version is `3.0`.
-Future minor revisions will be denoted as `3.1`, `3.2` ...
+The format is `3.MINOR` or `3.MINOR.PATCH` where MINOR and PATCH are non-negative integers without leading zeros.
+The current release version is `3.0`.
+Future minor revisions will be denoted as `3.1`, `3.2`, etc.
```
chrbertsch:
```suggestion
The format is `3.MINOR` or `3.MINOR.PATCH` where MINOR and PATCH are non-negative integers without leading zeros.
```

docs/2_4_common_schema.adoc
```diff
@@ -223,11 +223,11 @@ The XML attributes of `<fmiModelDescription>` are:
 
 |`fmiVersion`
 |Version of FMI the XML file complies with.
-The value for this version is `3.0`.
-Future minor revisions will be denoted as `3.1`, `3.2` ...
+The format is `3.MINOR` or `3.MINOR.PATCH` where MINOR and PATCH are non-negative integers without leading zeros.
+Patch releases may be denoted as `3.0.1`, `3.1.2`, etc.
```
chrbertsch:
```suggestion
The format is `3.MINOR` or `3.MINOR.PATCH` where MINOR and PATCH are non-negative integers without leading zeros.
```

PR #2160, opened 2026-04-21
Clarify that different FMU instances can be called in parallel from t…

### LINKED ISSUE #2151: Multi-threading and multiple FMU instances
On the topic of thread-safety, the fmi standard (in both versions 2 and 3) only has the following to say: "FMI functions of one instance are not required to be thread-safe" (in addition to forbidding changes to global settings). No mention is made of the behaviour of functions on different instances.

In another [discussion](https://github.com/modelica/fmi-standard/issues/128) I saw that the following requirement was at least considered at some point: "Instances of the same FMU and instances of different FMUs must not influence each other", however I have not found it in the standard anywhere. My question is whether this requirement somehow follows from another part of the standard or whether there is a good reason that it was left out of the standard. Not being able to assume this severely limits how a simulation environment can use multi-threading to evaluate model equations.

Tc-Fast:
Related to #2151.

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -29,7 +29,7 @@ To prepare the FMU code to run reliably on <<preemption-support, preemptive syst
 [NOTE]
 ====
 This property ensures that functions of different FMU instances can be called safely in any order.
-Additionally, they can be called in parallel provided the functions are called in different processes.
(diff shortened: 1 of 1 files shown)


pmai (approved):
Looks good to me, as it addresses the inadvertent change in 637889058b4a5851e9f6dbae72c630b4ba18d4f9

PR #2164, opened 2026-05-05
Non normative text

### LINKED ISSUE #2163: Convert remaining non-normative (inline) text to admonitions

Tc-Fast:
Closes #2163


chrbertsch:
Discussion: 

Pierre: at some places we could remove these non-normative sentences (e.g. "typicyally (simulation) time")
Klaus: We  cleaned these up before, these are just reminders that should be removed. At some place one could just use normal brackets.

chrbertsch (approved):
As discussed at the FMI Design Meeting

PR #2166, opened 2026-05-26
Fix color theming of common state machine

Tc-Fast:
The common state machine currently is exported with both dark and light color sets. All other images are exported with light color theme only. I guess this is maybe a new feature of draw.io. However, the current light/dark mechanism in the documentation webpage can not handle it.

Easiest fix is to export the SVG files with light theme only.

image


chrbertsch:
What is the matter with this picture in the dark mode? 
For me it looks good - similar to the other state machine images.
grafik
grafik

Tc-Fast:
I think the problem is here, that my system is overall in dark mode and therefor the browser is also in dark mode.
So the browser already handles dark and light for the "general state machine", because it comes with light and dark colors. The "other state machines" don't, so they are always rendered in light.

Here are now two mechanisms for light/dark handling clashing. On the one hand the browser already handles the color switching naturally and then there is also Thorsten's mechanism inverting the colors.

Tc-Fast:
Thats the setting in chrome, I am refering to:
image

chrbertsch (approved):

PR #2167, opened 2026-05-27
Clarify terminal name uniqueness

### LINKED ISSUE #1959: Clarification on what is meant with " unique on each level" for the "name" attribute of a <Terminal>
In section [2.4.9.2.2 Terminals](https://fmi-standard.org/docs/3.0.1/#section-terminals) we wrote:

> The normalized string attribute name of the <Terminal> element is the [unique](https://fmi-standard.org/docs/3.0.1/#uniqueNameAttribute) instance name of the terminal. The terminal name must be unique on each level.

Does that mean the "name" attribute must be unique w.r.t. to the <Terminals> list only, or does the 2nd sentence extent the uniqueness to the names of <ModelVariables>?

I think we need a clarification here,

Tc-Fast:
Closes #1959

### DIFF
--- a/docs/2_4_common_schema.adoc
+++ b/docs/2_4_common_schema.adoc
@@ -2020,7 +2020,8 @@ Element `<fmiTerminalsAndIcons><Terminals>` is defined as:
 image::images/schema/Terminals.png[width=800]
 
 The normalized string attribute `name` of the `<Terminal>` element is the <<uniqueNameAttribute,unique>> instance name of the terminal.
-The terminal name must be unique on each level.
+The terminal name must be unique among siblings; a child terminal may have the same name as its parent.
+Terminal names are independent from the names of <<ModelVariables>>.
 
 [[matchingRule,`matchingRule`]]
 The normalized string attribute `matchingRule` describes the rules for variable matching in a connection of terminals.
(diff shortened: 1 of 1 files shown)


chrbertsch:
regarding: "Terminal names are independent from the names of <<ModelVariables>>.": so they can also be the same? Shall we state this explicitely?

Tc-Fast:
I think "Terminal names are independent from the names of ```<<ModelVariables>>"``` is verbose enough.

chrbertsch (approved):

elmir-nahodovic:
I just want to mention a "discrepancy" to Modelica, because in Modelica you can have a variable that is itself a connector. This is probably a non-issue

In general it makes sense to map a Modelica `connector` to an FMI `<Terminal>`, but this can be a bit strange for some examples, for example in MSL, we have `Modelica.Blocks.Interfaces.RealInput` which is defined as:

`connector RealInput = input Real`

So the variable `Modelica.Blocks.Interfaces.RealInput u;` is either mapped to:

1. 
```
<Terminal
      name="u"
      matchingRule="plug">
      </Terminal>
```
or
2.
```
<Terminal
      name="u"
      matchingRule="plug">
      <TerminalMemberVariable
        variableName="u"
        memberName="u"
        variableKind="signal"/>
```

But if we would import the second variant into a Modelica tool, the only sensible way to define it is as:

```
connector RealInput
input Real u;
end RealInput;
```

Which is different from `connector RealInput = input Real`

PR #2174, opened 2026-06-08
Fix allowed function calls list

### LINKED ISSUE #2169: Missing entries in "Allowed function calls"
1. ```fmi3Get{VariableType}``` missing from Instantiated
The equations table explicitly maps "Get variable start values as defined in ```modelDescription.xml```" → ```fmi3Get{VariableType}```. However the Allowed Function Calls section only lists ```fmi3Set{VariableType}```. The reference FMU include ```Instantiated``` in its Get-variable masks.
 
2. ```fmi3Get{VariableType}``` missing from ConfigurationMode and ReconfigurationMode
Both modes only list ```fmi3Set{VariableType}``` (and the exit function). The reference FMU includes both ```ConfigurationMode``` and ```ReconfigurationMode``` in its Get-variable masks.

3. ```fmi3GetNumberOfVariableDependencies``` / ```fmi3GetVariableDependencies``` not listed anywhere
These two functions are defined in the header and referenced in the schema chapter, but they never appear in any state's Allowed Function Calls section. The reference FMU marks both as ```MASK_AnyState```.

4. ```fmi3EnterConfigurationMode``` missing from ClockActivationMode (SE)
The reference FMU mask is ```MASK_fmi3EnterConfigurationMode = (Instantiated | StepMode | EventMode | ClockActivationMode)```. The CS StepMode and ME/CS EventMode sections both list it, but the SE ClockActivationMode Allowed Function Calls section does not.

5. ```fmi3SetTime``` missing from EventMode (ME)
The reference FMU allows ```fmi3SetTime``` in (EventMode | ContinuousTimeMode). In FMI 2.0 ```fmi2SetTime``` was explicitly listed in Event Mode. The FMI 3.0 EventMode Allowed Function Calls section never mentions ```fmi3SetTime```.

1., 3. and 4. are obviously errors in the standard. 2. and 5. could be both an error in the standard or in the reference FMUs. Not sure about that.

Tc-Fast:
Closes #2169

### DIFF
--- a/docs/2_3_common_states.adoc
+++ b/docs/2_3_common_states.adoc
@@ -276,6 +276,9 @@ This function can be called for variables with <<variability>> latexmath:[\neq] 
 +
 The intention is to set <<start>> and guess values for these variables.
 
+Functions <<get-and-set-variable-values,`fmi3Get{VariableType}`>>::
+Returns the current <<start>> values as defined in <<modelDescription.xml>> or as set by `fmi3Set{VariableType}`.
+
 [[fmi3EnterConfigurationMode,`fmi3EnterConfigurationMode`]]
 Function `fmi3EnterConfigurationMode`::
 +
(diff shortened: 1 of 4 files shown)


pmai (approved):
LGTM

PR #2177, opened 2026-06-09
Improve triggered output clock description

### LINKED ISSUE #2149: Can you assign variables to a triggered output clock (and when are they active)
As discussed in the design meeting on 2026-03-24: While some non-normative text hints that triggered output clocks are mostly intended for signaling an identifiable event to the outside, and less for triggering an internal "time partition".

However nothing in the normative text seems to prevent assigning variables to a triggered output clock. When that happens, however, it is not clear to me how (**when**) to access those variables, as they would only be accessible when the triggered output clock is active.

However it is currently unclear for how long a triggered output clock is active: Given that fmi3GetClock is the only way to detect it being active, and the standard states that calling fmi3GetClock on an active triggered output clock sets that clock immediately to fmi3Inactive, it is unclear whether this just describes API semantics, and the clock is really active until a later time (e.g. the next fmi3UpdateDiscreteStates call), or it directly inactivates the clock.

In the later case there would be no point in time when the importer would both know that the clock is active, and it still is active to allow accessing associated variables. In the former case this could be made to work, however this should be clearly defined in the standard.

If there is no real use case for triggered output clocks being associated with variables, then the standard should likely prohibit this.

Note that this also depends on the outcome of the discussion on when clocked variables are accessible.

chrbertsch:
resolves #2149


docs/2_2_common_mechanisms.adoc
```diff
@@ -904,8 +904,8 @@ This clock type is not allowed in Scheduled Execution.
 
 [NOTE]
 ====
-<<outputClock,Output Clocks>> enable an FMU to provide a trigger to the outside world, e.g. to a <<triggered,triggered input Clock>> of another FMU. +
-If an FMU wants to actively trigger a model partition (Clock) of itself, it should use a <<countdown-aperiodic-clock,countdown Clock>>.
+<<outputClock,Output Clocks>> enable an FMU to provide a trigger to the outside world, e.g. to a <<triggered,triggered input Clock>> of another FMU. It is also allowd to assign variables to a triggered output clock.  +
```
klausschuch:
```suggestion
<<outputClock,Output Clocks>> enable an FMU to provide a trigger to the outside world, e.g. to a <<triggered,triggered input Clock>> of another FMU.
It is also allowed to assign variables to a triggered output clock.  +
```

docs/2_2_common_mechanisms.adoc
```diff
@@ -904,8 +904,8 @@ This clock type is not allowed in Scheduled Execution.
...
+If an FMU only wants to actively trigger a model partition (Clock) of itself, it can also use a <<countdown-aperiodic-clock,countdown Clock>>.
```
klausschuch: this should be a normative note

chrbertsch: Why? It is only a recommendation ...

docs/2_3_common_states.adoc
```diff
@@ -665,6 +661,9 @@ If the semantics of certain Clocks require specific handling—such as simultane
 Function <<fmi3GetClock>>::
 is used to inquire the status of <<Clock,Clocks>>.
 
+NOTE: The importer calls this function for triggered output Clocks. 
+For connected FMUs, these might depend on an algbraic loop and might change their value during event iteration, so it might be necessary to call this function several times.
```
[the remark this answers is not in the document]
chrbertsch: " For a given activation, only the first call reports `fmi3ClockActive`; subsequent calls report `fmi3ClockInactive` until the Clock is activated again." This is only true in SE

docs/2_2_common_mechanisms.adoc
Tc-Fast: Sounds non-normative to me.

docs/2_2_common_mechanisms.adoc
```diff
@@ -904,8 +904,9 @@ This clock type is not allowed in Scheduled Execution.
 
 [NOTE]
 ====
-<<outputClock,Output Clocks>> enable an FMU to provide a trigger to the outside world, e.g. to a <<triggered,triggered input Clock>> of another FMU. +
-If an FMU wants to actively trigger a model partition (Clock) of itself, it should use a <<countdown-aperiodic-clock,countdown Clock>>.
+<<outputClock,Output Clocks>> enable an FMU to provide a trigger to the outside world, e.g. to a <<triggered,triggered input Clock>> of another FMU.
+It is also allowed to assign variables to a triggered output clock. +
+If an FMU only wants to actively trigger a model partition (Clock) of itself, it can also use a <<countdown-aperiodic-clock,countdown Clock>>.
```
chrbertsch:
```suggestion
If an FMU only wants to actively trigger a model partition (Clock) of itself, it might be more efficient to use a <<countdown-aperiodic-clock,countdown Clock>>.
```

Tc-Fast (approved):

TPfitzer (approved):

PR #2181, opened 2026-06-10
Add initial calling sequence diagram

### LINKED ISSUE #2127: Add a more comprehensive Code Example for handling clocks
With PR https://github.com/modelica/fmi-standard/pull/2098 , a first code example for a FMU with Event Mode and clocks and clocked variables was introduced. It would be worthwihile to add an example that is more comprehensive.

@masoud-najafi suggested: 
-- Setting non-clocked inputs ( also during during the event iteration)
-- Getting clocked and non-clocked outputs
-- handling dependent output clocks
-- handling independent output clocks
-- handling changing clocks
-- handling count-down clocks
-- iteration in event mode

pmai:
This adds an initial calling sequence diagram and related toolchain changes to make this work.

### DIFF
--- a/.github/workflows/build-spec.yml
+++ b/.github/workflows/build-spec.yml
@@ -93,7 +93,6 @@ jobs:
         docker run --rm -v $PWD:/documents/ \
           uwebarthel/asciidoctor:1.101 \
           asciidoctor -b html5 -r asciidoctor-diagram/mermaid --failure-level INFO -v docs/index.adoc \
-          --attribute=imagesoutdir=docs/images \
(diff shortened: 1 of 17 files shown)


chrbertsch:
Discussion at FMI Design Meeting: 
the calling sequence is only partial, there might be additional calls to GetClocks necessary ... this is just a starting point
This shall not be a normative part of the standard, this has to be stated very explicitly. It is only one way of implementing it.
The numbers should be switched off in the final version, or we state that the numbers have no meaning.
Perhaps the multiple FMUs could be removed, has to be checked if we need to talk about interaction

PR #2184, opened 2026-06-16
Enhance documentation on discreteStatesNeedUpdate

### LINKED ISSUE #2150: Useful example why fmi3UpdateDiscreteStates might request discreteStatesNeedUpdate
### Discussed in https://github.com/modelica/fmi-standard/discussions/2148

<div type='discussions-op-text'>

<sup>Originally posted by **klausschuch** March 24, 2026</sup>
While discussing https://github.com/modelica/fmi-standard/issues/2133, the following question came up:
Why would an FMU set discreteStatesNeedUpdate=True in a fmi3UpdateDiscreteStates() call?
I can only come up with very artificial example but have no idea for a good model that needs this.
According to the "Equations and Actions" table in https://fmi-standard.org/docs/3.0.2/#EventMode, the values returned by the Get... calls should already consider the new discrete state anyway. Note, that this is unrelated to the fact if [fmi3EvaluateDiscreteStates](https://fmi-standard.org/docs/3.0.2/#fmi3EvaluateDiscreteStates) got called or not.

</div>

pmai:
Added note about handling intermediate states during event iteration.

### DIFF
--- a/docs/2_3_common_states.adoc
+++ b/docs/2_3_common_states.adoc
@@ -716,6 +716,14 @@ include::../headers/fmi3FunctionTypes.h[tag=UpdateDiscreteStates]
 
 [[discreteStatesNeedUpdate,`discreteStatesNeedUpdate`]]
 * When `discreteStatesNeedUpdate = fmi3True`, the importer must stay in <<EventMode>> for another event iteration, starting a new super-dense time instant.
++
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


chrbertsch:
@elmir-nahodovic : we can in this change also cross reference my addition to the guides.

chrbertsch:
@elmir-nahodovic : I have an example in mind with multiple clutches that I want to add

PR #2185, opened 2026-06-16
Clarify modelstructure for clocked variables

### LINKED ISSUE #2180: Ambiguous relationship between clocks attribute and dependencies for clocked variables in ModelStructure
Section 2.2 states:

> Declaring a variable as clocked variable using the clocks attribute specifies a dependency too, but also declares that such a clocked variable can only be accessed when one of the referenced Clocks is active.

Section 2.4 (note in the dependencies attribute description) states:

> This means that all dependencies must be listed, also the dependencies of clocked variables on their clocks.

If a clocked variable explicitly declares dependencies in ModelStructure, must the value reference of its Clock be included in that dependencies list? Or does the clocks attribute already cover this?

Tc-Fast:
Closes #2180

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -968,12 +968,14 @@ Output clocks may depend on input Clocks and other variables.
 Such <<dependencies>> are declared in the <<ModelStructure>>.
 A Clock latexmath:[k] depends on a Clock or variable latexmath:[v] if an activation or the value of latexmath:[v] may cause an activation of latexmath:[k] during the same super-dense time instant.
 
-Declaring a variable as <<clocked-variable,clocked variable>> using the <<clocks>> attribute specifies a <<dependencies,dependency>> too, but also declares that such a clocked variable can only be <<get-and-set-variable-values,accessed>> when one of the referenced Clocks is active.
+Declaring a variable as <<clocked-variable,clocked variable>> using the <<clocks>> attribute declares that such a clocked variable can only be <<get-and-set-variable-values,accessed>> when one of the referenced Clocks is active.
 This also holds for <<clocked-variable,clocked variables>> of type <<Clock>>.
+The association between a clocked variable and its Clocks is expressed solely by the <<clocks>> attribute and must not be repeated as a <<dependencies,dependency>> in the <<ModelStructure>>.
 
 [NOTE]
 ====
-This means that variables referenced by <<clocks>> and <<dependencies>> are not strict subsets of one another because <<clocks>> may list output Clocks that are not part of the latexmath:[{\mathbf{v}_{\mathit{known}}}].
+The <<clocks>> attribute and <<dependencies>> serve different purposes: <<clocks>> declares when a variable can be accessed, while <<dependencies>> declares algebraic dependencies on known variables.
+The Clocks referenced by the <<clocks>> attribute and the variables listed in <<dependencies>> are therefore disjoint.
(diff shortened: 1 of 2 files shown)


chrbertsch (approved):
Looks good to me

chrbertsch (changes requested):
Klaus: the original standard was ambiguous. 
We now decided to allow only one possibility. We should mark this explicitely.
Christian: E.g. add non-normative text: "This was changed in FMI 3.0.3 because .... it was ambiguous" and add it to the release notes.

chrbertsch (approved):
Looks good, thanks!

docs/2_2_common_mechanisms.adoc
```diff
@@ -968,12 +968,20 @@ Output clocks may depend on input Clocks and other variables.
...
 ====
-This means that variables referenced by <<clocks>> and <<dependencies>> are not strict subsets of one another because <<clocks>> may list output Clocks that are not part of the latexmath:[{\mathbf{v}_{\mathit{known}}}].
+The <<clocks>> attribute and <<dependencies>> serve different purposes: <<clocks>> declares when a variable can be accessed, while <<dependencies>> declares algebraic dependencies on known variables.
+The Clocks referenced by the <<clocks>> attribute and the variables listed in <<dependencies>> are therefore disjoint.
+====
+
+[NOTE]
+====
+Prior to FMI 3.0.3, the standard did not unambiguously define whether the association between a <<clocked-variable,clocked variable>> and its Clocks had to be repeated as a <<dependencies,dependency>> in the <<ModelStructure>>.
+This has been clarified in this version of the standard to allow only one possibility: this association is expressed solely by the <<clocks>> attribute and must not be listed as a <<dependencies,dependency>>.
```
klausschuch, line 984: Instead of "in this version", we maybe should specify the exact version "3.0.3" where this will be changed and this will stay for future versions of the document

docs/2_2_common_mechanisms.adoc
```diff
@@ -1392,6 +1400,15 @@ include::examples/snippets.c[tags=AlgebraicLoop3]
 
 When solving algebraic loops in <<EventMode>>, limitations to variable manipulations declared with attribute <<canHandleMultipleSetPerTimeInstant>> must be observed.
 
+A dependency declared in <<dependencies>> between <<clocked-variable,clocked variables>> only contributes to an algebraic loop while the variables involved can be accessed at the same super-dense time instant, that is, while their respective <<clocks>> are simultaneously active.
```
klausschuch:
```suggestion
A dependency declared in <<dependencies>> between <<clocked-variable,clocked variables>> contributes to an algebraic loop only when the variables involved can be accessed at the same super-dense time instant, that is, when their respective <<clocks>> are active at that super-dense time instant.
```

docs/2_2_common_mechanisms.adoc
```diff
@@ -1392,6 +1400,15 @@ include::examples/snippets.c[tags=AlgebraicLoop3]
...
+Therefore, the importer must take the activation of the <<clocks>> declared via the <<clocks>> attribute into account when forming algebraic loop candidates across connected FMUs: a declared dependency between two clocked variables forms a loop candidate only if their Clocks can be active at the same super-dense time instant.
```
klausschuch: Do we need this normative sentence?
Isn't this actually an implementation suggestion?
I'd drop it, otherwise we need to define what is meant with loop candidate.

PR #2188, opened 2026-07-07
Update deserialization notes for FMU states

### LINKED ISSUE #2186: Guidance on Compiler/Platform-Agnostic FMU State Serialization
# Issue: Guidance on Compiler/Platform-Agnostic FMU State Serialization

## Section Reference

Section 2.2.7.4 — Getting and Setting the Complete FMU State (`fmi3SerializeFMUState` / `fmi3DeserializeFMUState`)

## Summary

The current specification defines the serialization and deserialization API for FMU state (`fmi3SerializeFMUState`, `fmi3DeserializeFMUState`) but provides no guidance or requirements regarding the portability of the serialized byte vector across compiler versions, compiler vendors, or system architectures.

This omission means that in practice, serialized FMU state is implicitly coupled to the exact build environment that produced the FMU binary.

## Problem Statement

The spec states:

> "This can be, for example, used to perform an expensive steady-state initialization, copy the received FMU state in a byte vector and store this vector on file. Whenever needed, the byte vector can be loaded from file and deserialized, and the simulation can be restarted from this FMU state."

This use case (persist to file, reload later) implies temporal separation between serialization and deserialization. However, nothing in the normative text addresses what happens when the FMU binary changes between those two points — which is common in practice.

A naive but conforming implementation may serialize internal state via `memcpy` of structs. Such an implementation produces byte vectors that are sensitive to:

1. Struct padding and alignment (compiler-dependent, flag-dependent)
1. Endianness (architecture-dependent)
1. Pointer size (platform-dependent, e.g. 32-bit vs 64-bit)
1. Compiler-specific type sizes (e.g. `long` on Windows vs Linux)
1. Optimization flags that may reorder struct members or eliminate fields

## Consequences of the Current Gap

Without guidance, importers and integrators face the following operational burdens:

1. **Serialized state must be tracked against the exact FMU binary** — any recompilation (even with identical source) may invalidate stored state.
1. **Cross-platform workflows are unsafe** — state serialized on a Linux build cannot be assumed deserializable on a Windows build of the same FMU.
1. **CI/CD pipelines that rebuild FMUs** may silently corrupt checkpointed state.
1. **Long-lived serialized state** (e.g. stored initialization snapshots) becomes fragile across FMU version updates.
1. **No mechanism exists** for an importer to detect incompatibility between a serialized blob and the current FMU binary.

## Requested Guidance

We request the FMI Design Committee consider the following additions:

### Option A: Normative portability requirements

Require that serialized FMU state be portable across:

1. Compiler versions (same vendor)
1. Compiler vendors (for the same target platform)
1. System architectures (endianness, pointer width)

This would effectively mandate a well-defined serialization format (e.g. platform-neutral encoding with explicit sizes, little-endian convention, no raw struct copies).

### Option B: Normative versioning/compatibility mechanism

If full portability is too restrictive, provide a mechanism for importers to detect incompatibility:

1. A function like `fmi3GetSerializedStateVersion` or a version tag embedded in the serialized byte vector.
1. Normative requirement that `fmi3DeserializeFMUState` must return `fmi3Error` (not undefined behavior) if the byte vector is incompatible with the current binary.
1. Guidance that the FMU should document its serialization compatibility guarantees.

### Option C: Informative guidance (minimum)

At minimum, add a normative NOTE or recommendation clarifying:

1. That serialized state portability across compiler versions or platforms is NOT guaranteed unless the FMU explicitly documents otherwise.
1. That importers should not assume serialized state survives FMU recompilation.
1. Recommended practices for FMU developers who wish to provide portable serialization (e.g. avoid raw `memcpy` of structs, use explicit field-by-field encoding, include a format version header).

## Recommendation

We recommend **Option B** as the pragmatic middle ground — it does not impose an unreasonable implementation burden on FMU developers, but gives importers a reliable way to detect and handle incompatibility rather than encountering silent corruption.

Option A would be ideal for interoperability but may be too prescriptive for the diverse FMU ecosystem.

Option C is the minimum acceptable clarification but does not solve the detection problem.

## Impact Assessment

- **Blast radius:** Medium — affects all FMUs that advertise `canSerializeFMUState = true` and any workflow that persists serialized state across time or environments.
- **Backward compatibility:** Options B and C are additive. Option A would require existing FMUs to update their serialization logic.
- **Implementer effort:** Option B is low-to-moderate (error detection on deserialize, optional version tag). Option A is high (full format redesign for some FMUs).

pmai:
Clarified deserialization requirements and limitations for FMU states, closes #2186.

### DIFF
--- a/docs/2_2_common_mechanisms.adoc
+++ b/docs/2_2_common_mechanisms.adoc
@@ -606,6 +606,13 @@ include::../headers/fmi3FunctionTypes.h[tags=DeserializeFMUState]
 This function deserializes the byte vector `serializedState` of length `size`, constructs a copy of the FMU state and returns <<FMUState>>, the pointer to this copy.
 
 These serialization and deserialization functions may be called, only if the capability flags <<canGetAndSetFMUState>> and <<canSerializeFMUState>> are set to `true`.
+Any serialized state must be deserializable into a valid state for the same FMU and the same execution platform.
+
+[NOTE]
+====
+Whether the state can also be deserialized on another platform of the same FMU, or even across different versions of the same FMU is implementation dependent, so that in general an importer or a user cannot expect this to work, or to even be detected as an error.
+FMUs can of course document their serialization format, and make broader guarantees, but that is beyond the scope of this standard.
+====
 
 ==== Clocks [[Clock,Clock]]
 


chrbertsch (approved):
Looks good.

(@pmai: Shall we add a link to FMI-LS-REF with the possibility to reference platform specific serialization files? Or is it a bad idea to reference layered standards from the core standard?)

pmai:
> Looks good.
> 
> (@pmai: Shall we add a link to FMI-LS-REF with the possibility to reference platform specific serialization files? Or is it a bad idea to reference layered standards from the core standard?)

I would not do so just for this, and not as long as FMI-LS-REF is not yet released. I can see the benefit of pointing to layered standards for particular issues, but for FMI-LS-REF I would point to that for more generally interesting things (e.g. to embed multiple parameter sets, etc.), and as part of that, maybe mentioning stored state then also makes sense.

If a new LS for specifying stored state capabilities were to be created, then I think it would make sense to point to that, once released, from the current text here.

KarlWernersson (approved):
Looks good

Tc-Fast (approved):

PR #2192, opened 2026-08-21
Add CLA checkbox to PR template

chrbertsch:


Tc-Fast (commented):
Since you are doing changes to this template, I would suggest to update it in general. 
Three of four links are deprecated. I guess steps like "regenerate schema figures" do not make any sense anymore, since those are automated on CI pipeline.

Tc-Fast (commented):
I think you forgot to add the CLA checkbox now.

Tc-Fast (approved):
Looks good to me.

Issue #1362, opened 2021-03-20
Remove explicit values from fmi3DependencyKind


t-sommer:
as 0 is used for fmi3Independent now.

PR #2199, opened 2026-09-17
Clarify "VariableDependencies" functions

### LINKED ISSUE #2028: Clarify usage of fmi3GetNumberOfVariableDependencies and fmi3GetVariableDependencies
... clarify for which variables one is allowed to call `fmi3GetNumberOfVariableDependencies` as well as which variable types will be included in the return from `fmi3GetVariableDependencies`. Is it all variables?

copied from https://github.com/modelica/fmi-standard/issues/2024#issuecomment-2677696483

Tc-Fast:
## Related Issues

Closes #2028

## Changes

Clarifies when it is allowed to call fmi3GetNumberOfVariableDependencies and fmi3GetVariableDependencies.
Also clarifies for what variables these functions are allowed to be called.

I could not find any hint on what the enum value "fmi3Independent" is used for.
It is never mentioned anywhere in the spec.

The only issue related to it is this one: https://github.com/modelica/fmi-standard/issues/1362
It does not explain anything though.

It is also mentioned here:
https://github.com/modelica/fmi-design/blob/21b79d1b2bcbedafd108d98202ce0daad7879cda/WorkingGroups/04_Hybrid_CoSimulation/2017-Work_and_Meetings/FCPs/Header/fmi3FunctionTypes.h#L81
> /* fmi3Independent = 0, not needed but reserved for future use */

So I have no idea what it is good for, therefor I clarified:
> The meaning of the value `fmi3Independent` is not defined in this version of the standard and it must not be returned by this function.

Maybe someone can remember what "fmi3Independent" should be used for.

## Checklist

* [x] [Built the specification](https://github.com/modelica/fmi-standard/blob/main/CONTRIBUTING.md#building-the-specification-document).
* [x] [Linted the documents](https://github.com/modelica/fmi-standard/blob/main/CONTRIBUTING.md#validating-the-files).
* [x] My employer has signed the [Modelica Association Contrilbutor License Agreement (CLA)](https://github.com/modelica/ModelicaAssociationCLA/releases/download/1.1.1/ModelicaAssociationCLA_1.1.1.pdf)


docs/2_2_common_mechanisms.adoc
```diff
@@ -1280,9 +1280,24 @@ The <<dependencies>> are encoded in the <<modelDescription.xml>> with the elemen
...
+An FMU can indicate via the <<providesPerElementDependencies>> capability flag that it is able to provide detailed dependency information at runtime through the following C-API.
 The dependency information returned by these functions depend on the current operating point and parameter settings.
 
+The functions <<fmi3GetNumberOfVariableDependencies>> and <<fmi3GetVariableDependencies>> must only be called if <<providesPerElementDependencies>> is `true`, and only for a dependent variable that is listed as an unknown in the <<ModelStructure>>:
+
+* before <<fmi3ExitInitializationMode>> has been called: the variables referenced by the <<InitialUnknown>> elements,
+
+* after <<fmi3ExitInitializationMode>> has been called: the variables referenced by the <<Output>>, <<ContinuousStateDerivative>>, <<ClockedState>> and <<EventIndicator>> elements.
+
+For any other variable, the FMU must return <<fmi3Error>>.
```
KarlWernersson: Why must the FMU return an error, it's just an unnecessary restriction on the exporter

KarlWernersson: Example in the ls DAE;  there will be residual variables, these will be declared as local, and will allow get directional Derivatives etc on them, with respect to the general knows and the new "knows" algebraic variables".

Introducing such restrictions would complealty lock out the ls-dea proposal

docs/2_2_common_mechanisms.adoc
```diff
@@ -1280,9 +1280,24 @@ The <<dependencies>> are encoded in the <<modelDescription.xml>> with the elemen
...
+
+The variables returned as independents are the knowns latexmath:[{\mathbf{v}_{\mathit{known}}}] of the respective <<ModelStructure>> element, as defined for the <<dependencies>> attribute.
+The returned dependency information must be consistent with the <<ModelStructure>>: a variable must only be returned as an independent if it is listed in the <<dependencies>> attribute of the respective element, or if that attribute is not present.
+
+NOTE: Calling these functions is an optional optimization: an importer must be able to simulate the FMU correctly based on the <<ModelStructure>> alone.
+The <<dependencies>> declared there are therefore a conservative over-approximation, which these functions only refine by resolving to individual array elements, taking the current size of arrays into account and reflecting the current operating point.
+Consequently, fewer dependencies may be returned than declared in the <<ModelStructure>>, but never additional ones.
```
KarlWernersson: Im not sure how to write it better, because the funcion itself if a bit messy, but and you would get a much longer list of dependencies including their index.

What is limited is fewer or same amount of unique valuereferences in the independents

KarlWernersson:
Regarding independent, i can't see its usefulness either, it's likely a mistake someone looked at the variability and thought it should be there added it and it was missed in the reviews

docs/2_2_common_mechanisms.adoc
Tc-Fast: > Why must the FMU return an error, it's just an unnecessary restriction on the exporter

For ordinary FMUs, it is reasonable to have only meaningful return values for "VariableDependencies" functions on the unknowns.

Anyway, i changed it to:

> For any other variable, the FMU must either
> * return `<<fmi3Error>>` from both `<<fmi3GetNumberOfVariableDependencies>>` and `<<fmi3GetVariableDependencies>>`, or
> * return the complete set of dependencies of that variable, where the value references returned as independents may reference any variable of the FMU that the dependent variable depends on.

docs/2_2_common_mechanisms.adoc
Tc-Fast: That's exactly what I meant. I further clarified:

> The number of returned entries can therefore be larger than the number of value references listed in the `<<dependencies>>` attribute, while the set of value references returned as independents can only become smaller.

